PR #10, opened 2016-10-20
Subscribe to a topic multiple times

keks:
This PR basically does what I explained [here](https://github.com/libp2p/go-floodsub/pull/2#issuecomment-254834842). I rebased everything so that it is easily mergable. Yay!


keks:
paging @whyrusleeping

keks:
CR please :)

floodsub.go
```diff
@@ -126,12 +132,16 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			}
 		case treq := <-p.getTopics:
 			var out []string
-			for t := range p.myTopics {
-				out = append(out, t)
+			for t, subs := range p.myTopics {
+				if len(subs) > 0 {
```
whyrusleeping: should this ever not be > 0? That seems like it could end up being a memory leak

floodsub.go
```diff
@@ -303,59 +351,34 @@ func (p *PubSub) publishMessage(from peer.ID, msg *pb.Message) error {
...
 }
 
+// GetTopics returns the topics this node is subscribed to
 func (p *PubSub) GetTopics() []string {
 	out := make(chan []string, 1)
 	p.getTopics <- &topicReq{resp: out}
 	return <-out
 }
 
-func (p *PubSub) SubscribeComplicated(td *pb.TopicDescriptor) (<-chan *Message, error) {
```
whyrusleeping: We still need the topic descriptors. They arent super used yet, but thats the next step for this impl.

whyrusleeping:
@keks the approach looks good to me, i'd like to see some new tests exercising that behaviour from a users perspective though (as in, how a consumer would use multiple subscriptions)

floodsub.go
keks: I just searched for occurences of myTopics and it seems there is no way that it's empty. I'll just drop the check. Thanks!

keks:
@whyrusleeping there you go!

keks:
I added the test showcasing that you can now subscribe to one topic multiple times. I guess in the test it doesn't look very impressing, but in the ipfs daemon it's important we can do that because the might be pubsub clients that subscribe to the same topic but don't even know about each other.

floodsub.go
```diff
@@ -303,59 +350,47 @@ func (p *PubSub) publishMessage(from peer.ID, msg *pb.Message) error {
...
-	return p.SubscribeComplicated(&pb.TopicDescriptor{
-		Name: proto.String(topic),
-	})
-}
-
-type topicReq struct {
-	resp chan []string
-}
+// Subscribe returns a new Subscription for the given topic
+func (p *PubSub) Subscribe(topic string) *Subscription {
```
whyrusleeping: We should ensure that all subscribe calls go through `SubscribeByTopicDescriptor`, it will make things easier on us moving forward

whyrusleeping:
@keks fantastic, thank you! One more comment and then LGTM

keks:
Done.

Note this introduces an API change since `Subscribe()` will now also return an error (the error returned from `SubscribeByTopicDescriptor()`). At the moment these errors could be ignored because they will not happen anyway, but that didn't seem right to me as a long-term strategy. I you want I can `panic` instead.

keks:
Just saw this fails on Mac. Do you know why stuff like this happens? If not I'll ask around if I can borrow/ssh to one so I can reproduce this...

keks:
There you go...

keks:
I rebased and added the context to `Subsciption.Next()` so it can be cancelled.

whyrusleeping:
I'm ready to merge this, just gotta catch it right after youve rebased.

keks:
*sigh* I'll figure it out eventually...

keks:
actuallu I don't get the problem. The builders complain they can't find the libp2p gx dependency, but that change happened in c0c5a385ba88db56c54fc6b0ccba53f0ab976915. The tests don't even start. Can you kick it so it tries again @whyrusleeping?

Kubuxu:
@keks I think that his package might have been moved to: https://github.com/libp2p/go-libp2p-netutil

Kubuxu:
And it might have been my fail on not fixing this up before updating the dep.

whyrusleeping:
@keks use `QmcDTquYLTYirqj71RRWKUWEEw3nJt11Awzun5ep8kfY7W` for go-netutil, I was in the middle of updating a few things and havent gotten to bubbling them up entirely. (i think)

whyrusleeping:
You can also give me push access to this branch and i can fix it (alternatively, why are you not a contributor yet? let me fix that)

keks:
:tada:

keks:
Tests pass, no conflicts and gx published...I guess this means ready to merge? What do you think @whyrusleeping?

PR #29, opened 2017-07-28
notify: handle new connections asynchronously

Stebalien:
Trying to create a stream from with a connection notifier is *not* allowed
(and will deadlock).

Also,

* use the pubsub context when trying to open the stream.
* cut a new release...


Stebalien:
@Stebalien LGTY?
@Stebalien LGTM!

PR #34, opened 2017-08-30
drop messages to slow peers

Stebalien:
Instead of spawning a go routine for every message sent to every peer, buffer
them (actually, we already buffer 32 so this didn't need to be changed) and drop
messages when the buffer fills up.

fixes https://github.com/ipfs/go-ipfs/issues/4066

(writing to a channel in a go routine can be dangerous...)


whyrusleeping:
Actually, we should probably log when something gets dropped

Stebalien:
@whyrusleeping done.

PR #35, opened 2017-09-02
Drop announce messages to slow peers

film42:
Very similar to https://github.com/libp2p/go-floodsub/pull/34, only this protects the main loop from getting stuck announcing to a slow peer.

I also added a small optimization to `subscribedToMsg` to check to see if our topic list is empty before iterating through the topics in the message.


Stebalien:
Thanks!

PR #39, opened 2017-10-14
mark the peer as dead if the inbound stream closes

### LINKED ISSUE #38: Likely go routine leak of handleSendingMessages
Probably an issue with stream closing versus resetting.

Stebalien:
fixes #38

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -26,6 +26,10 @@ func (p *PubSub) getHelloPacket() *RPC {
 }
 
 func (p *PubSub) handleNewStream(s inet.Stream) {
+	defer func() {
+		p.peerDead <- s.Conn().RemotePeer()
+	}()
+
 	r := ggio.NewDelimitedReader(s, 1<<20)
 	for {
 		rpc := new(RPC)
--- a/comm.go
+++ b/comm.go
@@ -26,10 +26,6 @@ func (p *PubSub) getHelloPacket() *RPC {
 }
 
 func (p *PubSub) handleNewStream(s inet.Stream) {
-	defer func() {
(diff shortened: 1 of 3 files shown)


Stebalien:
Technically, we may still have a *connection* to the peer but they're still "dead" from the perspective of pubsub (they've closed their pubsub stream to us).

Stebalien:
Wait before you merge this. There's an edge case when we receive a second stream from a peer and kill the first.

Stebalien:
Actually, there are quite a few existing race conditions around that. I don't think there are any that can cause leaks but many that can cause two peers to fail to connect (e.g., if they both connect to each other at the same time). Really, we should support multiple connections and multiple streams to each peer.

However, IMO, that's an issue for another time (https://github.com/libp2p/go-floodsub/issues/40) and this PR should be good to go (except for the test hangs but I believe that's https://github.com/libp2p/go-floodsub/issues/22)

comm.go
```diff
@@ -74,21 +77,16 @@ func (p *PubSub) handleSendingMessages(ctx context.Context, s inet.Stream, outgo
 			if !ok {
 				return
 			}
-			if dead {
```
Stebalien: Shouldn't need this anymore (I believe this existed as a hack to prevent some deadlocks but those are no longer possible).

whyrusleeping: I believe you are correct. This was because sends on the other end of this outgoing channel could get filled up and block the loops. All those sends are now coupled with default selects (though please double check this for me).

Stebalien:
I'll do a release once I get https://github.com/libp2p/go-peerstream/pull/13 reviewed once more and merged (so I can bubble both of them up at once).

PR #41, opened 2017-10-16
Rename gx package to go-libp2p-floodsub

film42:
Resolves https://github.com/libp2p/go-floodsub/issues/7

Should be easy to coordinate this change with go-ipfs, etc. I would worry about others using this lib, but it's still beta software, so I think it's fine to switch the package name now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,8 +70,7 @@
   "gxVersion": "0.9.0",
   "language": "go",
   "license": "",
-  "name": "floodsub",
+  "name": "go-libp2p-floodsub",
   "releaseCmd": "git commit -a -m \"gx publish $VERSION\"",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -70,8 +70,7 @@
   "gxVersion": "0.9.0",
   "language": "go",
   "license": "",
-  "name": "floodsub",
+  "name": "go-floodsub",
```
diasdavid: bonus points for renaming it to `go-libp2p-floodsub`. Matches JS (https://github.com/libp2p/js-libp2p-floodsub) and lets the users know that this is for libp2p.

film42: Oooo, thank you!

Issue #42, opened 2017-10-16
Reliable and/or Persistent Pubsub


keks:
## Reliable Pubsub
Sometimes, messages get lost. To allow the receiver to recover lost messages, we could start systematically referencing previous messages when constructing them. That way, when the receiver receives the next message, it will find an unknown message reference and can ask its peers about it (or possibly fetch it like a regular ipfs/ipld object?)

The system I propose is the system is the system used at OrbitDB. Every peer maintains a list of messages that have not been referenced yet, DAG leafs so to speak. Once a new message is received, all message IDs referenced by the new message are removed from the list, and the new message's ID is added.
Once we author a message, we add a field `"after": [...]`, containing the leaf list. Afterwards our leaf list contains only the ID of the message we just authored.

## Persistent Pubsub
If we have something like the above, nodes already keep state about the "newest" messages. If we send these to peers sending us a subscription request for that topic, they can start fetching old messages. How many this would be depends on the application - in orbit it would suffice to fetch one screen worth of backlog, but in other application you may want to fetch the entire history.
This solves the problem described in #19.

Note: I want to make this optional. Whether or not this is used depends on the topic configuration.

keks:
Additional benefits: Some information on relative message timing (i.e. msg x was before msg y) can be derived (not for all messages of course).

keks:
So basically I want to optionally incorporate a orbit-style concurrent log into pubsub, but without orbit's rich query features.

keks:
@Stebalien What do you think about this? If you approve of this I'll go ahead and do it.

keks:
bump @Stebalien :)

Stebalien:
First of all, sorry for the massive delay. I keep putting this off to avoid switching all the pubsub context and switching out all secio/transport context but I should have gotten to this ages ago.

> That way, when the receiver receives the next message, it will find an unknown message reference and can ask its peers about it (or possibly fetch it like a regular ipfs/ipld object?)

I took a stab at this in https://github.com/libp2p/go-floodsub/tree/feat/use-ipld but got stuck in the weeds trying to figure out how to integrate CRDTs. However, you may find some of the changes I made there useful/interesting (take a look at the commit messages, I tried to write informative ones).

> Once a new message is received, all message IDs referenced by the new message are removed from the list, and the new message's ID is added.

[...]

PR #45, opened 2017-11-14
Allow caller to optionally validate messages

keks:
Allows callers of `PubSub.Subscribe` to pass a validator function `func(Message)bool` that returns whether the passed message should be processed or not.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -328,7 +328,7 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) error {
 			if p.validate(subs, &Message{pmsg}) {
 				p.sendMsg <- sendReq{
 					from: rpc.from,
-					msg:  &*Message{pmsg},
+					msg:  &Message{pmsg},
 				}
 			}
 		}()
@@ -426,6 +426,8 @@ type addSubReq struct {
 	resp chan *Subscription
 }
 
+type SubOpt func(*Subscription) error
+
 // WithValidator is an option that can be supplied to Subscribe. The argument is a function that returns whether or not a given message should be propagated further.
 func WithValidator(validate func(*Message) bool) func(*Subscription) error {
 	return func(sub *Subscription) error {
@@ -436,14 +438,14 @@ func WithValidator(validate func(*Message) bool) func(*Subscription) error {
 }
 
 // Subscribe returns a new Subscription for the given topic
-func (p *PubSub) Subscribe(topic string, opts ...func(*Subscription) error) (*Subscription, error) {
+func (p *PubSub) Subscribe(topic string, opts ...SubOpt) (*Subscription, error) {
 	td := pb.TopicDescriptor{Name: &topic}
 
 	return p.SubscribeByTopicDescriptor(&td, opts...)
 }
 
 // SubscribeByTopicDescriptor lets you subscribe a topic using a pb.TopicDescriptor
-func (p *PubSub) SubscribeByTopicDescriptor(td *pb.TopicDescriptor, opts ...func(*Subscription) error) (*Subscription, error) {
+func (p *PubSub) SubscribeByTopicDescriptor(td *pb.TopicDescriptor, opts ...SubOpt) (*Subscription, error) {
 	if td.GetAuth().GetMode() != pb.TopicDescriptor_AuthOpts_NONE {
 		return nil, fmt.Errorf("auth mode not yet supported")
 	}
--- a/floodsub.go
+++ b/floodsub.go
@@ -17,9 +17,9 @@ import (
 )
 
 const (
-	ID              = protocol.ID("/floodsub/1.0.0")
-	maxConcurrency  = 10
-	validateTimeout = 150 * time.Millisecond
+	ID                     = protocol.ID("/floodsub/1.0.0")
+	maxConcurrency         = 10
+	defaultValidateTimeout = 150 * time.Millisecond
 )
 
 var log = logging.Logger("floodsub")
@@ -368,7 +368,7 @@ func msgID(pmsg *pb.Message) string {
 // validate is called in a goroutine and calls the validate functions of all subs with msg as parameter.
 func (p *PubSub) validate(subs []*Subscription, msg *Message) bool {
 	for _, sub := range subs {
-		ctx, cancel := context.WithTimeout(p.ctx, validateTimeout)
+		ctx, cancel := context.WithTimeout(p.ctx, sub.validateTimeout)
 		defer cancel()
 
 		result := make(chan bool)
@@ -474,7 +474,14 @@ func WithValidator(validate Validator) func(*Subscription) error {
 		sub.validate = validate
 		return nil
 	}
+}
 
+// WithValidatorTimeout is an option that can be supplied to Subscribe. The argument is a duration after which long-running validators are canceled.
+func WithValidatorTimeout(timeout time.Duration) func(*Subscription) error {
+	return func(sub *Subscription) error {
+		sub.validateTimeout = timeout
+		return nil
+	}
 }
 
 // Subscribe returns a new Subscription for the given topic
@@ -495,7 +502,8 @@ func (p *PubSub) SubscribeByTopicDescriptor(td *pb.TopicDescriptor, opts ...SubO
 	}
 
 	sub := &Subscription{
-		topic: td.GetName(),
+		topic:           td.GetName(),
+		validateTimeout: defaultValidateTimeout,
 	}
 
 	for _, opt := range opts {
--- a/floodsub.go
+++ b/floodsub.go
@@ -8,12 +8,12 @@ import (
 
 	pb "github.com/libp2p/go-floodsub/pb"
 
-	inet "gx/ipfs/QmNa31VPzC561NWwRsJLE7nGYZYuuD2QfpK2b1q9BK54J1/go-libp2p-net"
-	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
-	peer "gx/ipfs/QmXYjuNuxVzXKJCfWasQk1RqkhVLDM9jtUKhqc2WPQmFSB/go-libp2p-peer"
-	timecache "gx/ipfs/QmYftoT56eEfUBTD3erR6heXuPSUhGRezSmhSU8LeczP8b/timecache"
-	protocol "gx/ipfs/QmZNkThpqfVXs9GNbexPrfBbXSLNYeKrE7jwFM2oqHbyqN/go-libp2p-protocol"
-	host "gx/ipfs/Qmc1XhrFEiSeBNn3mpfg6gEuYCt5im2gYmNVmncsvmpeAk/go-libp2p-host"
+	logging "github.com/ipfs/go-log"
+	host "github.com/libp2p/go-libp2p-host"
+	inet "github.com/libp2p/go-libp2p-net"
+	peer "github.com/libp2p/go-libp2p-peer"
+	protocol "github.com/libp2p/go-libp2p-protocol"
+	timecache "github.com/whyrusleeping/timecache"
 )
 
 const ID = protocol.ID("/floodsub/1.0.0")
--- a/floodsub.go
+++ b/floodsub.go
@@ -16,7 +16,11 @@ import (
 	timecache "github.com/whyrusleeping/timecache"
 )
 
-const ID = protocol.ID("/floodsub/1.0.0")
+const (
+	ID                    = protocol.ID("/floodsub/1.0.0")
+	maxConcurrency        = 10
+	validateTimeoutMillis = 100
+)
 
 var log = logging.Logger("floodsub")
 
@@ -56,6 +60,9 @@ type PubSub struct {
 	// sendMsg handles messages that have been validated
 	sendMsg chan sendReq
 
+	// throttleValidate bounds the number of goroutines concurrently validating messages
+	throttleValidate chan struct{}
+
 	peers        map[peer.ID]chan *RPC
 	seenMessages *timecache.TimeCache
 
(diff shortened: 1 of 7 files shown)


floodsub.go
```diff
@@ -171,7 +175,19 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				continue
 			}
 		case msg := <-p.publish:
-			p.maybePublishMessage(p.host.ID(), msg.Message)
+			subs := p.getSubscriptions(msg) // call before goroutine!
+			go func() {
+				if p.validate(subs, msg) {
```
vyzo: Shouldn't at the very least be some debugging output when we fail validation?

floodsub.go
```diff
@@ -370,20 +403,47 @@ func (p *PubSub) publishMessage(from peer.ID, msg *pb.Message) error {
...
+	return func(sub *Subscription) error {
+		sub.validate = validate
+		return nil
+	}
+
 }
 
 // Subscribe returns a new Subscription for the given topic
-func (p *PubSub) Subscribe(topic string) (*Subscription, error) {
+func (p *PubSub) Subscribe(topic string, opts ...func(*Subscription) error) (*Subscription, error) {
```
vyzo: Can we typedef the func?

floodsub.go
```diff
@@ -309,7 +323,15 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) error {
 			continue
 		}
 
-		p.maybePublishMessage(rpc.from, pmsg)
+		subs := p.getSubscriptions(&Message{pmsg}) // call before goroutine!
+		go func() {
+			if p.validate(subs, &Message{pmsg}) {
```
vyzo: same here re debugging output on validation error.

floodsub.go
```diff
@@ -319,6 +341,17 @@ func msgID(pmsg *pb.Message) string {
 	return string(pmsg.GetFrom()) + string(pmsg.GetSeqno())
 }
 
+// validate is called in a goroutine and calls the validate functions of all subs with msg as parameter.
+func (p *PubSub) validate(subs []*Subscription, msg *Message) bool {
```
vyzo: this validates for all subscriptions -- but there may be a case that the message is valid for some topic and invalid for some other topic. Shouldn't we publish it for the valid topic?

floodsub.go
```diff
@@ -171,7 +175,19 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				continue
 			}
 		case msg := <-p.publish:
-			p.maybePublishMessage(p.host.ID(), msg.Message)
+			subs := p.getSubscriptions(msg) // call before goroutine!
+			go func() {
```
vyzo: Why are we making this contraption with validation in a separate goroutine and then re-entry to the event loop with a separate message?
It seems to me that validators can be nonblocking and that we can simply validate synchronously and proceed to maybePublishMessage.

keks: whyrusleeping said validators may run for a while so they shouldn't block the event loop.

floodsub.go
keks: I don't know - is that log-worthy? To me it seems that a validator returning false is not very exceptional - and if it is, the caller can still log from inside the validator.

floodsub.go
keks: I talked to why about this and we weren't sure what to do. For now we just wanted to drop it. Maybe the right thing is to remove the topics it fails for and one drop the message if all topics have been removed?

floodsub.go
keks: Will do, good point

floodsub.go
keks: The reasoning was "we can always make things more complicated" :)

floodsub.go
vyzo: I think it is, it may indicate a bug or an attack and these things are nice to always nice to know.

keks: I guess it depends on what the validator is used for. Anyway, added logging [here](https://github.com/libp2p/go-floodsub/pull/45/commits/af6cff63be0db01c0672a8041c1d4975c10eebe5#diff-c37f865b99d25cd944a588b8b2e90cbeR347)

whyrusleeping:
@vyzo Doing validation out of the event loop is unfortunately necessary. Validation of messages that take even a few milliseconds will block the entire loop, which could to a significant bottleneck under load. Though It might be nice to find a way to rate limit this in some way to prevent unbounded concurrency

keks:
We can throttle it using a buffered `chan struct{}` but what happens when the maximum number of messages waiting to be validated is reached? Either block in the even loop or drop the message.
If we're going down that way, we should also add timeouts so an eternally blocking validator doesn't make all of floodsub block or drop all the messages.

whyrusleeping:
@keks yeah, i'm thinking maybe thats out of scope for this PR

keks:
Alrighty. I'll open an issue where we can discuss the trade-offs.

vyzo:
@whyrusleeping what kind of validators do we have in mind?
If we have decided we want the asynchronous validation contraption it's fine with me.

vyzo:
Another issue that we should be aware of with the goroutine validation approach is that it can potentially reorder messages. It's consistent with our very weak semantics for pubsub, but still an issue to be aware of.

vyzo:
Another issue that should be noted is the potential for Denail of Service attacks due to unbounded concurrency in the validation. A goroutine is created for each message, which allows a malicious peer to flood the network with potentially deleterious effects.

I think it might be in scope for the PR to address message throttling (#46)

whyrusleeping:
@vyzo i'm playing with the idea of using pubsub for a blockchain as a way to propagate blocks and transactions. Before a node re-broadcasts a block it receives via pubsub, it should make sure that its a valid block. Same with transactions. Validating a transaction is generally pretty cheap, but validating a block is a bit more expensive.

I might be okay with having things received on a given topic block eachother (maybe with some bounded concurrency).

On reordering messages, I'm okay with that. However, the DoS vector is definitely worth thinking about.

keks:
> On reordering messages, I'm okay with that. However, the DoS vector is definitely worth thinking about.

Message reordering can already happen, since messages can be dropped, so the application layer already can't depend on ordered delivery. I want to address this in #42.

re DoS: I'll work on bounding the concurrency tomorrow. Doing it in a separate PR is just more bureaucracy, (albeit not much more).

keks:
I made some further changes, I hope we can merge this now:

- make validators time out after 100ms
  - add context param to validator functions
  - add type Validator func(context.Context, *Message) bool
- drop message if more than 10 messages are already being validated

floodsub.go
```diff
@@ -16,7 +16,11 @@ import (
 	timecache "github.com/whyrusleeping/timecache"
 )
 
-const ID = protocol.ID("/floodsub/1.0.0")
+const (
+	ID                    = protocol.ID("/floodsub/1.0.0")
+	maxConcurrency        = 10
+	validateTimeoutMillis = 100
```
vyzo: Can you make this actual milliseconds and call it validateTimeout?

floodsub.go
```diff
@@ -319,6 +365,21 @@ func msgID(pmsg *pb.Message) string {
 	return string(pmsg.GetFrom()) + string(pmsg.GetSeqno())
 }
 
+// validate is called in a goroutine and calls the validate functions of all subs with msg as parameter.
+func (p *PubSub) validate(subs []*Subscription, msg *Message) bool {
+	for _, sub := range subs {
+		ctx, cancel := context.WithTimeout(p.ctx, validateTimeoutMillis*time.Millisecond)
```
vyzo: and here the millisecond should be part of the timeout constant.

floodsub.go
```diff
@@ -319,6 +365,21 @@ func msgID(pmsg *pb.Message) string {
...
+		defer cancel()
+
+		if sub.validate != nil && !sub.validate(ctx, msg) {
```
vyzo: so the validation timeout is not enforced at all, but simply passed into the validator procedure as the context?

keks:
@vyzo good feedback, thanks!

keks:
Oh, and I increased the timeout to 150ms because otherwise the test that checks whether a message gets dropped when the when all validators are taken would fail because the first message would already have timeouted when it sent the last one.
Maybe we should make the timeout configurable as well?

vyzo:
> Maybe we should make the timeout configurable as well?

Probably a good idea.

keks:
`go test -v` fixes it. with v for vorking I guess.

keks:
@vyzo Further comments? Would like to see this merged :)

floodsub.go
```diff
@@ -16,7 +16,11 @@ import (
 	timecache "github.com/whyrusleeping/timecache"
 )
 
-const ID = protocol.ID("/floodsub/1.0.0")
+const (
+	ID                     = protocol.ID("/floodsub/1.0.0")
+	maxConcurrency         = 10
```
vyzo: Let's make this configurable too?

vyzo: it doesn't have to be configurable per subscription -- I just don't like hardcoding a low value.
But up to you.

keks: Where else would you make it configurable? As a parameter to NewFloodSub? Then we had a global value for all subscriptions, and we would need to hard code a value in go-ipfs. I don't think that is a good option either.

I agree that 10 might not be the best value, so I sat down and did some simulations. They are quite rough, but I think we can still use them to inform our decision. You can find the code and "results" at https://gist.github.com/keks/a3bd10097c9c1d20ef5a2de3b5f794e3. Based on that, I'd suggest a value of 128. That seems quite high, but I believe we might drop too many packets otherwise.

The simulation assumes message validation takes as long as its timeout, so it's a worst-case analysis.

What do you think @whyrusleeping @vyzo?

vyzo: It can be either a parameter in `NewFloodSub` or just make it a variable.
Either way, if we make the validation timeout configurable, we should make the max queue size configurable as well.

whyrusleeping: Yeah, maybe we make this an option on `NewFloodsub`. When adding 'options' to things, i really like this approach: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

floodsub.go
```diff
@@ -309,7 +339,23 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) error {
...
+
+				if p.validate(subs, &Message{pmsg}) {
+					p.sendMsg <- sendReq{
+						from: rpc.from,
+						msg:  &Message{pmsg},
+					}
+				}
+			}(pmsg)
+		default:
+			log.Warning("could not acquire validator; dropping message")
```
whyrusleeping: Maybe in the default here, we can have another select, just like this one, but instead of a default there, we have a timeout. That way we avoid starting a timer for every message, and we don't immediately drop any overflow

floodsub.go
```diff
@@ -319,6 +365,32 @@ func msgID(pmsg *pb.Message) string {
...
+	for _, sub := range subs {
+		ctx, cancel := context.WithTimeout(p.ctx, sub.validateTimeout)
+		defer cancel()
+
+		result := make(chan bool)
+		go func(sub *Subscription) {
+			result <- sub.validate == nil || sub.validate(ctx, msg)
+		}(sub)
+
+		select {
```
whyrusleeping: this select should probably go in a separate loop, after this one. That way we do all the validation in parallel

floodsub.go
```diff
@@ -319,6 +365,32 @@ func msgID(pmsg *pb.Message) string {
...
 
+// validate is called in a goroutine and calls the validate functions of all subs with msg as parameter.
+func (p *PubSub) validate(subs []*Subscription, msg *Message) bool {
+	for _, sub := range subs {
+		ctx, cancel := context.WithTimeout(p.ctx, sub.validateTimeout)
+		defer cancel()
+
+		result := make(chan bool)
+		go func(sub *Subscription) {
+			result <- sub.validate == nil || sub.validate(ctx, msg)
```
whyrusleeping: maybe should return an error from the validator? Could be useful to log it

floodsub.go
keks: I'm not convinced a global setting is helpful here, but I'm okay going with it. And sure, functional options are nice! That's how I implemented other options in this PR as well.

floodsub.go
keks: My thinking was that the validator is user-supplied and they can do all the logging they want themselves. No need to make the API complex.

floodsub.go
keks: Wouldn't that block the main loop? And what would be the difference to just using the timeout instead of the default?

floodsub.go
keks: Yep, good catch!

floodsub.go
```diff
@@ -171,7 +199,27 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+					if p.validate(subs, msg) {
+						p.sendMsg <- sendReq{
+							from: p.host.ID(),
+							msg:  msg,
+						}
+
+					}
+				}(msg)
+			default:
+				log.Warning("could not acquire validator; dropping message")
```
vyzo: Perhaps we should log some more information about the message? The topic and the message id come to mind.

floodsub.go
```diff
@@ -370,20 +469,57 @@ func (p *PubSub) publishMessage(from peer.ID, msg *pb.Message) error {
...
-	resp  chan *Subscription
+	sub  *Subscription
+	resp chan *Subscription
+}
+
+type SubOpt func(*Subscription) error
+type Validator func(context.Context, *Message) bool
+
+// WithValidator is an option that can be supplied to Subscribe. The argument is a function that returns whether or not a given message should be propagated further.
+func WithValidator(validate Validator) func(*Subscription) error {
```
vyzo: can we use `SubOpt` in the return type for this prototype?

floodsub.go
```diff
@@ -370,20 +469,57 @@ func (p *PubSub) publishMessage(from peer.ID, msg *pb.Message) error {
...
+	return func(sub *Subscription) error {
+		sub.validate = validate
+		return nil
+	}
+}
+
+// WithValidatorTimeout is an option that can be supplied to Subscribe. The argument is a duration after which long-running validators are canceled.
+func WithValidatorTimeout(timeout time.Duration) func(*Subscription) error {
```
vyzo: here too

vyzo:
I don't quite like the fact that we are always entering the validation logic, regardless of whether there is a validator appropriate for the message. 
This means that all messages are throttled by the validation throttle, which limits our throughput to 10 concurrent messages overall.

I think we should increase the default throttle to something like a 100 or more.

floodsub.go
```diff
@@ -16,7 +16,11 @@ import (
 	timecache "github.com/whyrusleeping/timecache"
 )
 
-const ID = protocol.ID("/floodsub/1.0.0")
+const (
+	ID                     = protocol.ID("/floodsub/1.0.0")
+	defaultMaxConcurrency  = 10
```
vyzo: Let's make this a 100 at least (or even 1000) -- it's a global throttle on all messages not just per channel.

vyzo:
More on the subject of the global throttle: it is possible to DoS all topics by simply overflowing a topic that has a slow validator. This is an ugly Dos vector that can be addressed by only entering the validation logic only if there is an applicable validator.

vyzo:
Actually, that's not quite sufficient -- i think we need a separate throttle per subscription so that no topic can DoS other validated topics.

vyzo:
Closing in favour of #55,  so that I can work on it.

PR #47, opened 2017-11-18
Extract pb/ to go-libp2p-pubsub

lgierth:
New location: https://github.com/libp2p/go-libp2p-pubsub

This shrinks the dependency tree of go-ipfs-api massively \o/


lgierth:
> Maybe we should extract some interfaces as well?

I only went with what I had an immediate need for -- what do you have in mind?

keks:
Maybe I misunderstood something, so please bear with me. From my understanding go-libp2p-pubsub is about providing generic interfaces that can be implemented by specific protocol that provide pubsub functionality. One of them (or rather the only one) is go-floodsub.
The question is, are the protobuf message definitions specific to the protocol or are compatible with all libp2p pubsub implementations? I know the question is hard to answer because at the moment there only is one implementation so it's hard to draw a line. But I would argue that it is plausible that other pubsub algorithms will require other protobuf messages, so I'd prefer to keep them where they are.

Maybe we can find another way to reduce the size of the dependency tree? I believe the idea is that go-floodsub has many dependencies and you don't want go-ipfs-api to depend on it. I played some things through in my head but none of the options I could come up with off the top of my head were really satisfactory.

lgierth:
Yeah those are valid concerns -- we can:

1. just rename the new repo to something go-libp2p-floodsub-pb
2. copy the relevant part of the protobuf, instead of sharing it

vyzo:
If it's just the pb for floodsub (so that it can be reused by others) then it should be renamed.
But do we really need a separate package for the floodsub pbs?

lgierth:
>  But do we really need a separate package for the floodsub pbs?

I wanna keep the libp2p dependencies to a minimum, which will make it less vulnerable to breaking dependencies when using plain `go get`. The go-floodsub package just pulls in sooo much stuff.

I'm really fine with just downright copying the pb file -- this stuff is explicitly experimental after all.

lgierth:
> which will make it less vulnerable

it = go-ipfs-api

keks:
I believe a good way to deal with this is decoupling pubsub protocol message from the message format used between ipfs daemon and client. While we're at it, let's keep the API types at a central place - go-ipfs/core/types or so. Then both go-ipfs and go-ipfs-api can use that to encode and decode values. There are a lot of types defined in core/commands for marshaling messages. I think it would be clever to reuse these between go-ipfs and go-ipfs-api.

lgierth:
> I believe a good way to deal with this is decoupling pubsub protocol message from the message format used between ipfs daemon and client.

Yeah agreed - the suggestion to just copy the protobuf's over to go-ipfs-api (as a first step) is in that spirit.

> While we're at it, let's keep the API types at a central place - go-ipfs/core/types or so.

It would have to be outside of go-ipfs, so that go-ipfs-api can make use of it. The Core API interfaces are in ipfs/interface-ipfs-core which would be a good place for these types too I guess.

keks:
> ipfs/interface-ipfs-core

I was not aware of that package, great! I think for now copying the pb is ok.

PR #50, opened 2017-11-27
Create Jenkinsfile

VictorBjelkholm:
Here is a first attempt at getting go projects running jenkins.


VictorBjelkholm:
So Travis fails on merge, but not when testing the branch by itself. Wth?

PR #53, opened 2018-01-03
Use atomic counter for message seqnos

### LINKED ISSUE #52: PubSub sometimes failing
[C# test](https://github.com/richardschneider/net-ipfs-api/blob/master/test/CoreApi/PubSubApiTest.cs#L97) and [JS test](https://github.com/ipfs/interface-ipfs-core/blob/master/src/pubsub.js#L496) are intermittently failing.  Both are  publishing multiple messages and then counting the subscribed messages. 

It appears that some published messages are not being relayed to the subscribed channels. In the C# test both publisher and subscriber are on the same peer.

NOTE: these tests are running on Windows.  This was first detected in https://github.com/ipfs/js-ipfs-api/pull/648

vyzo:
Closes #52 

The seqno is now the concatenation of the timestamp and an atomic counter; it's 128-bit wide.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	pb "github.com/libp2p/go-floodsub/pb"
@@ -57,6 +58,9 @@ type PubSub struct {
 	seenMessages *timecache.TimeCache
 
 	ctx context.Context
+
+	// atomic counter for seqnos
+	counter uint64
 }
 
 type Message struct {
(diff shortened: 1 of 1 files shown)


vyzo:
so I don't know what's the deal with the jenkins tests: it says "passed but marked as unstable". 
What is this supposed to mean?

floodsub.go
```diff
@@ -414,8 +418,10 @@ func (p *PubSub) GetTopics() []string {
 
 // Publish publishes data under the given topic
 func (p *PubSub) Publish(topic string, data []byte) error {
-	seqno := make([]byte, 8)
-	binary.BigEndian.PutUint64(seqno, uint64(time.Now().UnixNano()))
+	seqno := make([]byte, 16)
+	counter := atomic.AddUint64(&p.counter, 1)
+	binary.BigEndian.PutUint64(seqno[:8], uint64(time.Now().UnixNano()))
```
whyrusleeping: If we initially set `counter` to `time.Now().UnixNano()`, then we don't need to get the current time every time we send a message. 

Not a big deal, but might save us some syscalls

whyrusleeping: It also saves us from having to send a whole extra 8 bytes per message

vyzo: good points, nice small efficiencies. i'll open a new PR.

vyzo: actually, there is no syscall involved in getting the time; but still.

vyzo: opened follow-up PR.

PR #54, opened 2018-01-03
Narrow the message seqno to 64bit

vyzo:
Follow-up to #53 

Narrows the message seqno back down to 64bits, by using just the counter initialized at the instantiation time.
This saves us 8 bytes and potentially a syscall in some systems.


vyzo:
per conversation in irc -- it seems absurdly unlikely that we can run ahead of the current time with extant computing machinery and operating systems. We would need to send > 1B messages/s _sustained_ to run ahead for even 1s.

PR #55, opened 2018-01-13
Message Validators

### LINKED ISSUE #46: Throttle validation of incoming messages
When #45 is in, we'll need to find a way to throttle parallel validator executions, so if they take a long time we don't run out of memory.

I plan on doing this by passed them all a buffered `chan struct{}` that they write to before they are started and read from when they finish. Question is: Should we just block the event loop or should we drop a message when the maximum number of validators is running? I think we already drop messages in a similar situation, so I'd propose going the same way here.

This leads us to the next problem, which is that validators that block *very* long need to be cancelled because otherwise they'll start piling up until either the event loop blocks or we drop all packets. So, how large should the timeout be? 200ms?

ping @vyzo @whyrusleeping

### LINKED ISSUE #56: Per topic validation
In #55, validation is per-subscription. Unfortunately, this means:

1. If we have multiple subscribers subscribing with the same validator, we'll run it multiple times (e.g., check a signature multiple times). This is really inefficient.
2. If we have multiple subscribers subscribing with *different* validators, these subscriptions will affect each other. That's *not* what a user will suspect.

Really, validation is a property of the topic, not the subscription. Eventually, I'd like to be able to tell pubsub to relay a topic without explicitly subscribing (#28).

---

In terms of implementation, we have a few choices:

1. Require that all topics be namespaced and use the namespace to pick the validator.
2. Allow validators to decide whether or not they care about a topic.

2 is probably the easiest but not the cleanest. Possible interface:

```go

// Register a validator with the pubsub.
func (p *PubSub) RegisterValidator(v Validator)

type Validator interface {
    // Apply this validator to this topic.
    Validates(topic string) bool // or `ValidatorFor(topic string) TopicValidator`?
    // Apply this validator to this topic by descriptor. Do we even need topic descriptors anymore? We don't use them... Having both of these is annoying...
    ValidatesTopicDescriptor(topic *pb.TopicDescriptor) bool
    Validate(msg) bool
}
```

vyzo:
Implements asynchronous message validators.
Continuation of #45  which was merged and rebased in a feature branch so that I can continue work on it.
Closes #46.
Closes #56.

cc @whyrusleeping @keks


vyzo:
Removed the global validation throttle and cleaned up the subscription validation logic.
Let's reopen review?

floodsub.go
```diff
@@ -54,6 +58,9 @@ type PubSub struct {
 	// topics tracks which topics each of our peers are subscribed to
 	topics map[string]map[peer.ID]struct{}
 
+	// sendMsg handles messages that have been validated
+	sendMsg chan sendReq
```
vyzo: I am thinking that this should be a `chan *sendReq` for consistency with the other channels.

floodsub.go
```diff
@@ -324,6 +347,85 @@ func msgID(pmsg *pb.Message) string {
...
+	}
+
+	if needval {
+		// validation is asynchronous
+		// XXX vyzo: do we want a global validation throttle here?
+		go p.validate(subs, src, msg)
+		return
+	}
+
+	go func() {
```
whyrusleeping: hrm... spawning a goroutine for each message being sent is a bit wonky. Maybe:
```go
select {
case p.sendMsg <- .....:
default:
  go func() {
    p.sendMsg <- ....
  }()
}
```

floodsub.go
```diff
@@ -78,8 +85,10 @@ type RPC struct {
 	from peer.ID
 }
 
+type Option func(*PubSub) error
+
 // NewFloodSub returns a new FloodSub management object
-func NewFloodSub(ctx context.Context, h host.Host) *PubSub {
+func NewFloodSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
```
vyzo: I think we could avoid the interface change if we don't use any options -- do we want this for future proofing?

floodsub.go
vyzo: Good point, we only have to spawn a new goroutine if it's a publish, in which case the push happens within the event loop.
I'll update.

floodsub.go
vyzo: Let's keep it, it's a nice interface; also we may want to pass an option for global validation throttle.

subscription.go
```diff
@@ -31,3 +36,24 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
 func (sub *Subscription) Cancel() {
 	sub.cancelCh <- sub
 }
+
+func (sub *Subscription) validateMsg(ctx context.Context, msg *Message) bool {
+	result := make(chan bool, 1)
+	vctx, cancel := context.WithTimeout(ctx, sub.validateTimeout)
+	defer cancel()
+
+	go func() {
```
whyrusleeping: annoying that we have to use an extra goroutine for this, but probably okay.

vyzo:
Added a global validation throttle to have a hard upper limit in active validation goroutines.
I also set some reasonable defaults for the throttles: per validator throttle gets 100, global gets a round 8192.

subscription.go
vyzo: yeah, that's how it _go_es :)

vyzo:
summoning @Stebalien

subscription.go
Stebalien: Can we not just trust that validate will respect its context?

floodsub.go
```diff
@@ -324,6 +359,94 @@ func msgID(pmsg *pb.Message) string {
...
+			log.Warningf("message validation throttled; dropping message from %s", src)
+		}
+		return
+	}
+
+	sreq := &sendReq{from: src, msg: msg}
+	select {
+	case p.sendMsg <- sreq:
+	default:
+		go func() {
```
Stebalien: I'm not a fan of this pattern (too easy to open a bunch of go routines). Could we just call `maybePublishMessage` directly and rename the `sendMsg` channel name to `validatedMsgs` or something?

floodsub.go
```diff
@@ -324,6 +359,94 @@ func msgID(pmsg *pb.Message) string {
...
+	if needval {
+		// validation is asynchronous and globally throttled with the throttleValidate semaphore
+		select {
+		case p.validateThrottle <- struct{}{}:
+			go func() {
+				p.validate(subs, src, msg)
+				<-p.validateThrottle
+			}()
+		default:
+			log.Warningf("message validation throttled; dropping message from %s", src)
```
Stebalien: We'll probably want per-peer or per-topic throttles in the future but, for now, this is probably fine.

floodsub.go
```diff
@@ -324,6 +359,94 @@ func msgID(pmsg *pb.Message) string {
...
+	results := make([]chan bool, 0, len(subs))
+	throttle := false
+
+loop:
+	for _, sub := range subs {
+		if sub.validate == nil {
+			continue
+		}
+
+		rch := make(chan bool, 1)
```
Stebalien: I'd just use a single channel and count results from it. Allocating channels isn't *that* expensive but it's certainly not free.

floodsub.go
vyzo: we are always in the event loop actually. I added a buffer so that we almost always avoid the goroutine.

floodsub.go
vyzo: we already have those! That was part of the improvements in this PR (although most of the stuff got rewritten).

floodsub.go
vyzo: I added a buffer to sendMsg so that we should almost never overflow.
But you are right, we can just call `maybePublishMessage`!
I'll keep the `sendMsg` channel named as it is, because it indicates a message that is ready to send.

floodsub.go
vyzo: fair enough; it was _really_ convenient to write this way though :)

vyzo: alright, implemented!

floodsub.go
vyzo: I skipped the dance, and called `maybePublishMessage` directly .

floodsub.go
Stebalien: I see... I'm worried that it's still possible to starve a topic by maxing out the global validator first. Ideally, we'd take the per-topic ticket first, then the global one.

vyzo: I think it would be hard to do, because a throttled subscription would release the semaphore quickly.

vyzo:
so https://github.com/libp2p/go-floodsub/pull/55/commits/f1be0f12966b357474b260166482a3892a192601, which removed the external goroutine enforcing the timeout bombs two of keks' tests -- they explicitly test the external goroutine timing out.

vyzo:
Removed the two offending tests, as they test faulty behaviour.

More to the point, per discussion with @Stebalien on irc: if we timeout externally, the faulty validator is still running, while we unthrottle and leak a goroutine. It is preferable to have the faulty validator to cause the topic to throttle, at which point it can be identified as a bug and be debuged, rather than having a silent goroutine leak.

floodsub.go
vyzo: I improved the comments around the throttle to more clearly describe the intention and existence of per-topic throttles.

vyzo:
@Stebalien ping -- let's move this forward?

Stebalien:
I don't see how this PR is a step in the right direction.

Validators should either be a property of the topic (registered at initialization, preferably) *or* validators should be a property of the subscription in which case a message's validity for subscription A shouldn't affect the message's validity for subscription B. Given that making validators a property of the subscription makes it difficult to determine if we should forward the message, validators should be a property of the topic.

vyzo:
Ok, I will rewrite to implement per topic validators then.

vyzo:
summoning @Stebalien @whyrusleeping

floodsub.go
```diff
@@ -324,6 +370,75 @@ func msgID(pmsg *pb.Message) string {
 	return string(pmsg.GetFrom()) + string(pmsg.GetSeqno())
 }
 
+// pushMsg pushes a message performing validation as necessary
+func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
+	if len(vals) > 0 {
+		// validation is asynchronous and globally throttled with the throttleValidate semaphore.
```
whyrusleeping: Is this comment still true?

vyzo: Yes it is still true. 
I can remove the global throttle if we deem so, I just like to have a limit on how much goncurrency we can spawn as a result of incoming network traffic.

vyzo: although, subscription reference below! that's not true, I will fix the comment.

whyrusleeping:
Oh wait, I was thinking backwards, no global validators, but still global throttling.

vyzo:
> Why are you so opposed to having validators decide which topics they care about? 

Hrm, I am not terminally opposed to them -- I just find the interface a little clunky.

> The way this currently works, subscribing to the same topic twice will be painful.

Why? The validator registration is decoupled from the subscription and it's something we do at initialization.

jvsteiner:
Hey, I'm a brand-newbie in looking at this project, but for me, in the use cases I have in mind, I'd like to be able to validate by topic (although I can see why "by subscription would also be useful) - but perhaps more importantly, I'd like not to propagate messages to other peers, when they do not validate.  I might even want to peers who send invalid messages.  I guess peer attribution might be a bit outside the scope of what's currently possible - but that's my $0.02.

vyzo:
@jvsteiner with this pr, we do exactly that: we don't broadcast messages that don't validate on a per-topic basis. We have rejected subscription-based validators because of incoherent semantics. 

The global validators will simply add an additional layer below per topic validators: they will allow us to define validator classes for pubsub namespaces and will be implemented in follow up work.

Stebalien:
@whyrusleeping I believe we're waiting on you for the final signoff.

PR #57, opened 2018-01-14
TestSubscribeMultipleTimes fails go test -race.

rargulati:
TestSubscribeMultipleTimes consistently fails `go test -race` locally.
This may explain why it's a flappy test in CI (see https://github.com/libp2p/go-floodsub/issues/22).

Specifically to this test, not enough time is given for Subscribe to finish before Publish is executed.
On making this change, `libp2p/go-floodsub` passes `go test -count=2 -race` locally.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -253,7 +253,6 @@ func (p *PubSub) announce(topic string, sub bool) {
 }
 
 // notifySubs sends a given message to all corresponding subscribers.
-// A topic must have a subscription(s), otherwise notifySubs fails.
 // Only called from maybePublishMessage.
 func (p *PubSub) notifySubs(msg *pb.Message) {
 	for _, topic := range msg.GetTopicIDs() {
--- a/floodsub.go
+++ b/floodsub.go
@@ -252,8 +252,9 @@ func (p *PubSub) announce(topic string, sub bool) {
 	}
 }
 
-// notifySubs sends a given message to all corresponding subscribbers.
-// Only called from processLoop.
+// notifySubs sends a given message to all corresponding subscribers.
+// A topic must have a subscription(s), otherwise notifySubs fails.
+// Only called from maybePublishMessage.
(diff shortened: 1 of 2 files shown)


rargulati:
Hey all 👋 , I'm new to this codebase, libp2p, and ipfs.

Reading https://github.com/libp2p/community/blob/master/CONTRIBUTE.md didn't give me any info on how to tackle the `codecov/project` issue.

I wanted to make the smallest change possible to get the test suite working, but I also recommend turning on the `-race` flag for CI if it hasn't been enabled already (increases CI run time, unfortunately). I believe that's modifying `$GOTFLAGS` in https://raw.githubusercontent.com/ipfs/ci-helpers/master/travis-ci/run-standard-tests.sh (though I'm not familiar with jenkins/travis).

floodsub.go
```diff
@@ -252,8 +252,9 @@ func (p *PubSub) announce(topic string, sub bool) {
 	}
 }
 
-// notifySubs sends a given message to all corresponding subscribbers.
-// Only called from processLoop.
+// notifySubs sends a given message to all corresponding subscribers.
+// A topic must have a subscription(s), otherwise notifySubs fails.
```
vyzo: please don't add this comment, it's incorrect.

rargulati: Done in 27e81ba

floodsub.go
```diff
@@ -252,8 +252,8 @@ func (p *PubSub) announce(topic string, sub bool) {
 	}
 }
 
-// notifySubs sends a given message to all corresponding subscribbers.
-// Only called from processLoop.
+// notifySubs sends a given message to all corresponding subscribers.
+// Only called from maybePublishMessage.
```
vyzo: Keep this comment as is please too, the intention is to convey that it's in the event loop.
Btw, thanks for fixing the typo!

rargulati: Ah, ok, makes sense - done in 4643b97

PR #61, opened 2018-01-26
use the external protobuf definitions

Stebalien:
Otherwise, we can't import packages that import *both* this library and that
one (go-libp2p-pubsub).


Stebalien:
Required to fix ipget: https://github.com/ipfs/ipget/pull/50

Stebalien:
I'm not sure if we intend on actually moving this package. IMO, we'd put an *interface* in `go-libp2p-pubsub` and leave this as an implementation of that interface.

vyzo:
this is fine too -- I just don't want to have a proliferation of implementation packages with the new protocols.

Stebalien:
> I just don't want to have a proliferation of implementation packages with the new protocols.

Good point. Actually, in terms of "proliferation" we'd probably want that at a lower-level (different routing algorithms all managed by a single pubsub instance).

However, having a separate interface allows us to, e.g., swap in an alternative implementation that defers to a running ipfs daemon.

vyzo:
https://github.com/libp2p/go-libp2p-pubsub/issues/1

lgierth:
> I think that splitting out in go-libp2p-pubsub was a mistake -- we want to move this package (go-floodsub) to its place without losing history as we evolve the implementation.

Feel free to scratch the go-libp2p-package, we can live without the protobuf definitions in go-ipfs-api

Stebalien:
@lgierth how? go-ipfs-api depends on it. Should we revert that and make it depend on go-floodsub again?

lgierth:
In go-ipfs-api we use it pretty much only to parse the API's pubsub messages responses -- imo it'd be fine to just do that without protobufs. We parse every other response without protobufs as well.

jvsteiner:
My $0.02 - as a user who currently needs to import both `go-floodsub` and `go-ipfs-api` (which imports `go-libp2p-pubsub`) my main concern is that the current setup drives protobuf import conflicts.  Where the protobufs live doesn't matter much, and this seems to resolve that ambiguity.  When my team discussed this, we concluded this change was the most likely way to resolve the issue.

IMO, I'm not sure I see a great deal of value in separating the interface from the implementation - at least not to the extent of separating them into separate repositories.  In `go-libp2p/p2p/host/*` there are an implementations of `go-libp2p-host.Host` and this is kinda unhelpful. dunno if it's possible, but it might be useful to hold the interface and the reference implementation of that interface in the same repository - perhaps something like `go-libp2p-pubsub` for the interface and `go-libp2p-pubsub/floodsub` for the reference implementation.  This way, gx releases occur simultaneously, CI can check both are in sync, and users can rely on the fact that whatever version they import contains a working implementation - maybe it reduces some of the degrees-of-freedom that have to be thought through.

Stebalien:
See: https://github.com/libp2p/go-libp2p-pubsub/issues/4

PR #67, opened 2018-02-20
gossipsub

vyzo:
Implements the gossipsub protocol; see https://github.com/vyzo/gerbil-simsub for a high-level literate specification.

TODO:
- [X] tests tests tests!
- [X] subscription announce messages, sent by base `PubSub`, need to be reliable so that we accurately track peers in a topic

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -19,6 +19,10 @@ const (
 	GossipSubD   = 6
 	GossipSubDlo = 4
 	GossipSubDhi = 12
+
+	// gossip parameters
+	GossipSubHistoryLength = 5
+	GossipSubHistoryGossip = 3
 )
 
 func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
@@ -28,7 +32,7 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
 		fanout:  make(map[string]map[peer.ID]struct{}),
 		gossip:  make(map[peer.ID][]*pb.ControlIHave),
 		control: make(map[peer.ID]*pb.ControlMessage),
-		mcache:  NewMessageCache(3, 5),
+		mcache:  NewMessageCache(GossipSubHistoryGossip, GossipSubHistoryLength),
 	}
 	return NewPubSub(ctx, h, rt, opts...)
 }
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -199,27 +199,16 @@ func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
 
 		// gossipsub peers
 		gmap, ok := gs.mesh[topic]
-		if ok {
-			// direct peers in the mesh for topic
-			for p := range gmap {
-				tosend[p] = struct{}{}
-			}
-		} else {
-			// fanout peers, we are not in the mesh for topic
+		if !ok {
+			// we are not in the mesh for topic, use fanout peers
 			gmap, ok = gs.fanout[topic]
 			if !ok {
-				// we don't have any yet, pick some
-				var peers []peer.ID
-				for p := range tmap {
-					if gs.peers[p] == GossipSubID {
-						peers = append(peers, p)
-					}
-				}
+				// we don't have any, pick some
+				peers := gs.getPeers(topic, func(peer.ID) bool { return true })
 
 				if len(peers) > 0 {
 					gmap = make(map[peer.ID]struct{})
 
-					shufflePeers(peers)
 					for _, p := range peers[:GossipSubD] {
 						gmap[p] = struct{}{}
 					}
@@ -252,6 +241,24 @@ func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			// Drop it. The peer is too slow.
 		}
 	}
+}
+
+func (gs *GossipSubRouter) getPeers(topic string, filter func(peer.ID) bool) []peer.ID {
+	tmap, ok := gs.p.topics[topic]
+	if !ok {
+		return nil
+	}
+
+	peers := make([]peer.ID, 0, len(tmap))
+	for p := range tmap {
+		if gs.peers[p] == GossipSubID && filter(p) {
+			peers = append(peers, p)
+		}
+	}
+
+	shufflePeers(peers)
+
+	return peers
 }
 
 func (gs *GossipSubRouter) Join(topic string) {
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -68,27 +68,19 @@ func (gs *GossipSubRouter) HandleRPC(rpc *RPC) {
 	}
 
 	iwant := gs.handleIHave(ctl)
-	msgs := gs.handleIWant(ctl)
+	ihave := gs.handleIWant(ctl)
 	prune := gs.handleGraft(rpc.from, ctl)
 	gs.handlePrune(rpc.from, ctl)
 
-	if len(iwant) == 0 && len(msgs) == 0 && len(prune) == 0 {
+	if len(iwant) == 0 && len(ihave) == 0 && len(prune) == 0 {
 		return
 	}
 
-	// TODO piggyback gossip IHAVE
-	out := rpcWithControl(msgs, nil, iwant, nil, prune)
-
-	mch, ok := gs.p.peers[rpc.from]
-	if !ok {
-		return
-	}
-
-	select {
-	case mch <- out:
-	default:
-		// TODO PRUNE messages should be reliable; schedule for piggybacking or retry
-		log.Infof("dropping message to peer %s: queue full", rpc.from)
+	out := rpcWithControl(ihave, nil, iwant, nil, prune)
+	if len(prune) == 0 {
+		gs.sendMessage(rpc.from, out)
+	} else {
+		gs.sendControl(rpc.from, out)
 	}
 }
 
@@ -207,12 +199,7 @@ func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
 				peers := gs.getPeers(topic, func(peer.ID) bool { return true })
 
 				if len(peers) > 0 {
-					gmap = make(map[peer.ID]struct{})
-
-					for _, p := range peers[:GossipSubD] {
-						gmap[p] = struct{}{}
-					}
-
+					gmap = peerListToMap(peers[:GossipSubD])
 					gs.fanout[topic] = gmap
 				}
 			}
@@ -229,17 +216,78 @@ func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			continue
 		}
 
-		mch, ok := gs.p.peers[pid]
-		if !ok {
-			continue
-		}
+		gs.sendMessage(pid, out)
+	}
+}
 
-		select {
-		case mch <- out:
-		default:
-			log.Infof("dropping message to peer %s: queue full", pid)
-			// Drop it. The peer is too slow.
-		}
+func (gs *GossipSubRouter) Join(topic string) {
+	gmap, ok := gs.mesh[topic]
+	if ok {
+		return
+	}
+
+	gmap, ok = gs.fanout[topic]
+	if ok {
+		gs.mesh[topic] = gmap
+		delete(gs.fanout, topic)
+	} else {
+		peers := gs.getPeers(topic, func(peer.ID) bool { return true })
+		gmap = peerListToMap(peers[:GossipSubD])
+		gs.mesh[topic] = gmap
+	}
+
+	for p := range gmap {
+		gs.sendGraft(p, topic)
+	}
+}
+
+func (gs *GossipSubRouter) Leave(topic string) {
+	gmap, ok := gs.mesh[topic]
+	if !ok {
+		return
+	}
+
+	for p := range gmap {
+		gs.sendPrune(p, topic)
+	}
+	delete(gs.mesh, topic)
+}
+
+func (gs *GossipSubRouter) sendGraft(p peer.ID, topic string) {
+	graft := []*pb.ControlGraft{&pb.ControlGraft{TopicID: &topic}}
+	out := rpcWithControl(nil, nil, nil, graft, nil)
+	gs.sendControl(p, out)
+}
+
+func (gs *GossipSubRouter) sendPrune(p peer.ID, topic string) {
+	prune := []*pb.ControlPrune{&pb.ControlPrune{TopicID: &topic}}
+	out := rpcWithControl(nil, nil, nil, nil, prune)
+	gs.sendControl(p, out)
+}
+
+func (gs *GossipSubRouter) sendControl(p peer.ID, out *RPC) {
+	gs.sendRPC(p, out, true)
+}
+
+func (gs *GossipSubRouter) sendMessage(p peer.ID, out *RPC) {
+	gs.sendRPC(p, out, false)
+}
+
+func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC, ctl bool) {
+	// TODO control messages and gossip piggyback
+	// - control messages (GRAFT/PRUNE) must be reliable and should
+	//   be scheduled for piggyback or retry if the queue is full
+	// - gossip should be piggybacked on messages oppurtinistcally
+
+	mch, ok := gs.p.peers[p]
+	if !ok {
+		return
+	}
+
+	select {
+	case mch <- out:
+	default:
+		log.Infof("dropping message to peer %s: queue full", p)
 	}
 }
 
@@ -259,14 +307,6 @@ func (gs *GossipSubRouter) getPeers(topic string, filter func(peer.ID) bool) []p
 	shufflePeers(peers)
 
 	return peers
-}
-
-func (gs *GossipSubRouter) Join(topic string) {
-	// TODO
-}
-
-func (gs *GossipSubRouter) Leave(topic string) {
-	// TODO
 }
 
 func (gs *GossipSubRouter) heartbeatTimer() {
@@ -289,6 +329,14 @@ func (gs *GossipSubRouter) heartbeatTimer() {
 
 func (gs *GossipSubRouter) heartbeat() {
 	// TODO
+}
+
+func peerListToMap(peers []peer.ID) map[peer.ID]struct{} {
+	pmap := make(map[peer.ID]struct{})
+	for _, p := range peers {
+		pmap[p] = struct{}{}
+	}
+	return pmap
 }
 
 func shufflePeers(peers []peer.ID) {
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -248,10 +248,11 @@ func (gs *GossipSubRouter) Leave(topic string) {
 		return
 	}
 
+	delete(gs.mesh, topic)
+
 	for p := range gmap {
 		gs.sendPrune(p, topic)
 	}
-	delete(gs.mesh, topic)
 }
 
 func (gs *GossipSubRouter) sendGraft(p peer.ID, topic string) {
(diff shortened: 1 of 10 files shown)


vyzo:
Added a small zoo of basic gossipsub tests, including a mixed mode test with floodsub peers.

vyzo:
made a small tweak -- sources that have not joined the mesh also emit gossip.

vyzo:
Removed a potentially harmful topic membership check for mesh peers; potential inconsistency if the ANNOUNCE was lost (or reordered after GRAFT on retry).

gossipsub.go
```diff
@@ -0,0 +1,569 @@
...
+	toprune := make(map[peer.ID][]string)
+
+	// maintain the mesh for topics we have joined
+	for topic, peers := range gs.mesh {
+
+		// do we have enough peers?
+		if len(peers) < GossipSubDlo {
+			ineed := GossipSubD - len(peers)
+			plst := gs.getPeers(topic, ineed, func(p peer.ID) bool {
+				_, ok := peers[p]
```
whyrusleeping: this looks like we're filtering something, mind adding a comment on what we're filtering on?

gossipsub.go
```diff
@@ -0,0 +1,569 @@
...
+			case <-gs.p.ctx.Done():
+				return
+			}
+		case <-gs.p.ctx.Done():
+			return
+		}
+	}
+}
+
+func (gs *GossipSubRouter) heartbeat() {
```
whyrusleeping: This function feels too long. Mind trying to break it up a little bit?

gossipsub.go
```diff
@@ -0,0 +1,569 @@
...
+		// push control messages that need to be retried
+		ctl := out.GetControl()
+		if ctl != nil {
+			gs.pushControl(p, ctl)
+		}
+	}
+}
+
+func (gs *GossipSubRouter) heartbeatTimer() {
+	ticker := time.NewTicker(1 * time.Second)
```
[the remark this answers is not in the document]
vyzo: Not necessarily.
It will send GRAFT/PRUNE if it needs to adjust the overlay, and schedule gossip for piggybacking.
If the gossip is not sent by the next heartbeat, then it will be flushed in its own messages.

gossipsub.go
vyzo: Sure i'll refactor a bit, although I would like to keep the main logic together.

gossipsub.go
vyzo: we are filtering the peers that are not already in our peer list; i can add a comment to that extent.

gossipsub.go
vyzo: I can easily refactor out the part that sends the GRAFT/PRUNE with coalescing, which is also incidental for the logic in the heartbeat.

vyzo: done

vyzo:
Implemented retry of ANNOUNCE messages in pubsub, now that we have a test that exercises the relevant code paths.

gossipsub_test.go
```diff
@@ -0,0 +1,857 @@
...
+		ch, err := ps.Subscribe("fizzbuzz")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		chs = append(chs, ch)
+	}
+
+	// wait for heartbeats to build mesh
+	time.Sleep(time.Second * 2)
```
whyrusleeping: Why do we always have to wait for heartbeats? Is it because we don't send out subscription notices immediately now?

whyrusleeping:
Could we add some tests that check the number of messages sent, and maybe a way of tracking the overall efficiency of the implementation (like how many nodes received the same message from multiple peers) maybe in terms of bandwidth overhead? like, received 500 bytes for every 200 bytes of useful data at a message size of 100 bytes.

gossipsub_test.go
vyzo: Well, subscription notices are still sent immediately, just retried if they fail.

Now that I think of it, we can probably greatly reduce this delay in almost all the tests -- maybe down to say 100ms, just enough for announcements to go out. 

My rationale for this delay was to wait a couple of heartbeats to avoid interference from nodes who have joined but haven't seen any peer announcements yet. Also, I wanted to avoid interference from the overlay construction, but it should still be connected after the announcements get sent and nodes pick their peers.

vyzo:
Hrm, these are tests i would like to have too -- but not sure they are really unit tests.
What would be the conditions for test failure?

whyrusleeping:
Yeah, they are definitely integrations tests. No need to write them as unit tests. We should run these tests for floodsub and for gossipsub and compare the results, and choose some failure threshold, i.e. gossipsub should not use more bandwidth than floodsub

gossipsub_test.go
vyzo: Actually, there is a genuine concern that is an artifact of the concurrent Join from subscriptions.
If we subscribe all the nodes together, then they won't have any peer announcements when the do the Join, and they'll have to wait a heartbeat before they start adding peers to the mesh.
We can avoid this if we add a small (say 10ms) delay after each subscription.

whyrusleeping: Hrm... I'm very skeptical of 'fixing' things by adding delays.

vyzo: Also, in many of the tests the subsscriptions are created before connecting the network, which means that all nodes start empty and build the mesh purely in the heartbeat

paralin: @vyzo It's really really bad practice to use delays like this, especially without a select statement to escape out of it if the context is canceled.

vyzo: @paralin , this is just a test that _needs_ a delay -- and there is nothing to cancel the context so a select would be totally useless.

vyzo:
We have developed a conflict, so I will resolve and rebase.

vyzo:
Rebased; also added a context done check that was missing in the announce retry goroutine.

gossipsub_test.go
paralin: @vyzo Gotcha, I notice now it's a test.

vyzo:
The `TestGossipsubControlPiggyback` would occasionally hang in line 704 because of #69, so I added a fix for the issue.

ORBAT:
Hi folks! Any particular reason you went with a custom protocol instead of something built on Chord/Pastry/PolderCast etc?

vyzo:
hrm, seems like the fix lost the coverage for control piggybacking -- probably because of the slowdown with the `Errorf` logging.
I will downgrade that to `Infof`.

vyzo:
@ORBAT several reasons: simplicity of implementation, robustness, and perhaps most important of all backwards compatibility with floodsub so that we can easily deploy.

vyzo:
and control piggyback coverage is back, at least for `GRAFT`.

whyrusleeping:
@ORBAT It looks like PolderCast didnt make it into our pubsub research reading list: https://ipfs.io/ipfs/QmNinWNHd287finciBwbgovkAqEBQKvnys1W26sY8uupc5/pubsub%20reading%20list.pdf
Likely because the paper costs $30 to read.

In any case, we've done a pretty thorough review of the problem space before we arrived at our version zero protocol, floodsub. With the idea that it is the base layer protocol, and provides very few guarantees. This code, which we're calling gossipsub is an iterative improvement over floodsub that essentially only adds a fairly simple tree pruning via gossip. Simplicity and ease of implementation are very important for us, gossipsub can be implemented in [150 lines of scheme](https://github.com/vyzo/gerbil-simsub/blob/master/simsub/gossip.ss) and not too many lines of go.

That said, this is still under review. Review of the protocol and/or implementation is very much welcome.

vyzo:
Rebased on master to absorb the cherry-pick.

whyrusleeping:
Review from while i was on a plane today:
1. docs please, its not clear that `Join` and `Leave` are just callbacks from the pubsub to the router
	- in general, we should have docs on the expectations and contracts that a router is expected to fulfil
2. why a random delay on announceRetry?
3. random thought, but what happens if some peers set their gossipsub parameters really low? like 1 or 2.
4. publishing to a topic you arent subscribed to seems like it might get expensive. also maybe a bit leaky? If i publish once to a topic, it looks like i set fanout peers for that topic. Does that ever get cleaned up?
5. any thought around getting DoSed through the 'topics' field filling up? There are probably a thousand other ways to DoS pubsub, but this one seems pretty easy
6. heartbeat interval should probably be configurable, with the default being a constant. Maybe link to your experiments showing why 1 second is a good default
7. i would turn the 'getNetSequenceNumber' thing in pubsub.Publish into a separate function
8. more docs on the different implications between mesh and fanout
9. in heartbeat, it seems like we could combine the looping over plst with the initial getPeers call. Why not just put the logic from that look in the closure youre passing?
10. is there no 'too many peers' check for fanout becoming too large?
11. in heartbeat, we call flush to push out piggybacked messages that didnt get sent, and then at the end of heartbeat we send some more messages. Why not call flush after that to potentially coalesce the grafts and prunes from this heartbeat with the gossip info in flush?

vyzo:
>   docs please, its not clear that Join and Leave are just callbacks from the pubsub to the router
>       in general, we should have docs on the expectations and contracts that a router is expected to fulfil

Sure, will add!

>    why a random delay on announceRetry?

I wanted to give the writer a chance to make progress, without an immediate retry. 
The delay could be fixed I guess, but it seemed better to randomize it as it avoids pathologies from periodic publish processes.

>    random thought, but what happens if some peers set their gossipsub parameters really low? like 1 or 2.

The overlay construction algorithm will continue to build connected overlays as long as `D-low >= 2` and `D-high >= D+1`. 
If you set all parameters to 1 or 2 globally then you will probably fail to connect non-trivially sized overlays.


>    publishing to a topic you arent subscribed to seems like it might get expensive. also maybe a bit leaky? If i publish once to a topic, it looks like i set fanout peers for that topic. Does that ever get cleaned up?

Well, it's not expensive as it is a passive operation (you select peers that have announced the topic).

But you are right, it _is_ leaky! -- if you only publish one message and then forget the topic, the fanout peers will linger.
I can add some logic in the heartbeat that cleans up `fanout` if you haven't published again for some time.

>    any thought around getting DoSed through the 'topics' field filling up? There are probably a thousand other ways to DoS pubsub, but this one seems pretty easy

What's the vector? Are you concerned about the `topics` field in the envelope causing excessive cpu usage from processing?

>    heartbeat interval should probably be configurable, with the default being a constant. Maybe link to your experiments showing why 1 second is a good default

We should probably make all parameters configurable, not just the heartbeat interval. 
The 1s choice was made with responsiveness in mind, and also to minimize the gossip linger interval. We can easily increase the heartbeat, but then
- it will take longer to adjust the overlay
- the gossip will linger longer, as messages are gossiped for 3 heartbeat cycles.

>    i would turn the 'getNetSequenceNumber' thing in pubsub.Publish into a separate function

sure, can do

>    more docs on the different implications between mesh and fanout

sure, i will add some comments

>    in heartbeat, it seems like we could combine the looping over plst with the initial getPeers call. Why not just put the logic from that look in the closure youre passing?

Not sure I follow. The closure is just a filter, it shouldn't have side-effects. 
Specifically, it is called for every peer to filter it and there is no guarantee that the peer will be selected,
so we can't fuse the subsequent iteration in the closure.

>    is there no 'too many peers' check for fanout becoming too large?

fanout cannot become too large, we only add peers if we have less than `D`.

>    in heartbeat, we call flush to push out piggybacked messages that didnt get sent, and then at the end of heartbeat we send some more messages. Why not call flush after that to potentially coalesce the grafts and prunes from this heartbeat with the gossip info in flush?

This may be a little cumbersome to write, as the flush should happen before we emit fresh gossip (otherwise it will flush the fresh gossip too, which is not what we want). 
I can break `emitGossip` out of the `graft/prune` loop, move the `sendGraftPrune` right after the iteration and then do flush, followed by another iteration over the topic lists to emit gossip.
Not sure if it's worth the trouble and code uglification though.

vyzo:
Updates based on the feedback:
- added docs for the router interface, the contract should be cleaner now
- made all gossipsub parameters, including the heartbeat interval, variables so that users can set their own configuration
- added expiration for fanout lists with a 1 minute TTL (thanks for spotting that leak!)
- refactored the next seqno contraption out of `pubsub.Publish`

vyzo:
@mgoelzer fixed in https://github.com/libp2p/go-floodsub/pull/67/commits/88a9bb054176be98655e0ea13294c20074b5c95b

Also a note about our review norms: we don't normally block merging with an `X` for minor things, we just inline comment and fix them.

pubsub.go
```diff
@@ -401,6 +442,12 @@ func msgID(pmsg *pb.Message) string {
...
 func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
+	id := msgID(msg.Message)
+	if p.seenMessage(id) {
+		return
+	}
+	p.markSeen(id)
+
 	if len(vals) > 0 {
 		// validation is asynchronous and globally throttled with the throttleValidate semaphore.
 		// the purpose of the global throttle is to bound the goncurrency possible from incoming
```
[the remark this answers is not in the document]
vyzo: not a typo.

pb/rpc.proto
```diff
@@ -17,6 +19,30 @@ message Message {
 	repeated string topicIDs = 4;
 }
 
+message ControlMessage {
```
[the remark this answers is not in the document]
vyzo: I am following the protocol style we use in other protobufs in ipfs; keep it simple, stupid.

pubsub.go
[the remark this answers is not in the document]
whyrusleeping: not a typo?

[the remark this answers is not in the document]
vyzo: more of a pun :)
goroutines provide goncurrency, just like coroutines provide concurrency.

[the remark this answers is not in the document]
vyzo: obviously, we can fix it if you guys find my humor offensive :)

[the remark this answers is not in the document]
whyrusleeping: lol, i don't mind

pb/rpc.proto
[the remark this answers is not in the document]
lanzafame: Totally understand, just unfortunate that we went with a style that isn't in line with what the tooling suggests.

prestonvanloon:
Is this still in review? Is there anything we can do to help out?

whyrusleeping:
@prestonvanloon I think we will merge it shortly, The only thing I want to hold off for is for @Stebalien to get back (he's out this week) and give a +1.

jamesray1:
https://github.com/libp2p/go-floodsub/pull/79 should help for documentation.

gossipsub.go
```diff
@@ -0,0 +1,613 @@
...
+// For each topic we publish to without joining, we maintain a list of peers
+// to use for injecting our messages in the overlay with stable routes; this
+// is the fanout map. Fanout peer lists are expired if we don't publish any
+// messages to their topic for GossipSubFanoutTTL.
+type GossipSubRouter struct {
+	p       *PubSub
+	peers   map[peer.ID]protocol.ID         // peer protocols
+	mesh    map[string]map[peer.ID]struct{} // topic meshes
+	fanout  map[string]map[peer.ID]struct{} // topic fanout
+	lastpub map[string]int64                // last pubish time for fanout topics
```
jamesray1: This comment should say publish. ;)

jamesray1:
Could we have an ABI with JSON tests for other programming languages? This is needed for ensuring compatibility over the wire.

whyrusleeping:
> Could we have an ABI with JSON tests

What do you mean? Like test vectors and such for the wire protocol?

jamesray1:
Yep, exactly! The goal is to make sure that if we have different implementations of gossipsub that they will be compatible.

whyrusleeping:
@jamesray1 I can add that to my todo list :)

In general though, it's all just that protobuf being sent back and forth.

jamesray1:
Right, so if we use the same proto file we can make it compatible.

mhchia:
Will this PR be rebased to the current master?

It seems some of the APIs of the dependencies like `"github.com/libp2p/go-libp2p-netutil".GenSwarmNetwork` is changed.
So, I rebase this PR to the master branch locally. Some tests fail but I'm not sure the reason.
```
$ go test -run TestDenseGossipsub
--- FAIL: TestDenseGossipsub (1.03s)
	floodsub_test.go:53: dial attempt failed: <peer.ID X8bjX2> --> <peer.ID NtTuXy> dial attempt failed: EOF
FAIL
exit status 1
FAIL	github.com/libp2p/go-floodsub	1.049s
```
I changed the argument `d` passed to `denseConnect` from `10` to `8`, or other lower numbers to workaround this problem. But I didn't figure out the reason.
```
func denseConnect(t *testing.T, hosts []host.Host) {
	connectSome(t, hosts, 8)
}
```
Does anyone know the possible reason for this issue?  Or is this just the problem in my local environment?

Big thanks for reading my question.

whyrusleeping:
Hey @mhchia, sorry about that, We will get this rebased asap. I'll handle it in the morning if @Stebalien (who has more context on these changes) doesnt beat me to it.

Stebalien:
I've rebased but I'm having some trouble reproducing the issue. I'm currently running the test in a loop to see if that gets me anywhere.

Stebalien:
No dice.

jamesray1:
@whyrusleeping I just read https://github.com/libp2p/interop/issues/1. Having a daemon will of course be useful, although not having to depend on code in Go is preferable, and JSON tests are needed.

whyrusleeping:
I had no reason not to merge this, so I did. Next steps, putting it into a flag in ipfs.

diasdavid:
@whyrusleeping shouldn't this be a separate pubsub implementation so that folks can pick the pubsub implementation to use? 

Will this PR make PubSub in go-ipfs not interop with js-ipfs?

diasdavid:
@whyrusleeping just confirmed that this package would be better named `go-pubsub` and then we would have two others for `go-floodsub` and `go-gossipsub` to plug in here but refactoring things in go is hard so that will happen later.

jamesray1:
Also you really should use constant variables instead of literals.

mhchia:
@jamesray1 
Maybe it is because we can use different routing mechanism in the underlying overlay, not necessarily using DHT? In this case, "The initial contact nodes can be obtained via rendezvous with DHT provider records" might only be an example.

jamesray1:
@mhchia, sure, that's fine.

whyrusleeping:
@diasdavid 

> but refactoring things in go is hard so that will happen later.

Its more 'extracting things into multiple packages in go is annoying to do when you might be changing things in both really soon'.

@jamesray1 @mhchia Yeah, the DHT is only an example. You can use any means to rendezvous. Take a look at our rendezvous spec proposal for ideas towards a more specialized way of doing rendezvous

jamesray1:
What duration should we use for timeout requests? 

Context: implementing a system config for Kademlia to use to get nodes.

https://github.com/libp2p/rust-libp2p/blob/7507e0bfd9f11520f2d6291120f1b68d0afce80a/kad/src/high_level.rs#L36

As for the timeout duration, according to [RabbitMQ](https://www.rabbitmq.com/heartbeats.html), that is twice the heartbeat interval, which is 1 s in this Go implementation, therefore I think it should be 2 s. I am guessing to use [40000 s](https://github.com/amule-project/amule/blob/166ee417573e36c5162ad46b8da353d6b001ad03/src/include/protocol/ed2k/Constants.h#L34) for `kbuckets_timeout` (the Duration after which a node in the k-buckets needs to be pinged again.) but I'm not really sure, perhaps the spec and this implementation should also define initialization?

gossipsub.go
```diff
@@ -0,0 +1,613 @@
...
+	iwantlst := make([]string, 0, len(iwant))
+	for mid := range iwant {
+		iwantlst = append(iwantlst, mid)
+	}
+
+	return []*pb.ControlIWant{&pb.ControlIWant{MessageIDs: iwantlst}}
+}
+
+func (gs *GossipSubRouter) handleIWant(ctl *pb.ControlMessage) []*pb.Message {
+	ihave := make(map[string]*pb.Message)
```
raulk, line 143: Does this need to be a map, or could it be a slice? Is it a map to deduplicate message IDs?

vyzo: yes, it needs to deduplicate.

PR #70, opened 2018-02-28
UnregisterTopicValidator for removing topic validators

### LINKED ISSUE #68: Can't remove a topic validator
Hi !
I might be missing something, but I cannot find a way to unregister a topic validator.
Indeed, in my case I have multiple services that use the floodsub layer. A service subscribes to a topic and registers a validator for that topic. When stopping the service, I can call `Cancel()` on the `Subscribtion`, but there is no way to remove the topic validator from `Pubsub.topicVals`.

Maybe I am doing this wrong, but it feels necessary to have one of these:
- unregister the topic validator when canceling the subscription (not good IMO as you might want to set the validator once only at launch and then register/unregister as much as you want)
- expose a method to clean the validator for a given topic.

vyzo:
Closes #68

### DIFF
--- a/floodsub_test.go
+++ b/floodsub_test.go
@@ -341,6 +341,31 @@ func TestOneToOne(t *testing.T) {
 	checkMessageRouting(t, "foobar", psubs, []*Subscription{sub})
 }
 
+func TestRegisterUnregisterValidator(t *testing.T) {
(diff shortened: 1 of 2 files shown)


vyzo:
i'll add a test to keep codecov happy.

PR #72, opened 2018-03-07
Don't block the event loop because of slow subscriptions

### LINKED ISSUE #69: Slow subscriptions can block the event loop
The event loop dispatches messages to subscriptions in `notifySubs`, which puts the new message to the subscription's channel.
This channel is buffered for 32 messages, and then the put will block -- which blocks the entire event loop.

Still not sure if this is a bug or a feature, but i don't like slow subscribers blocking the router. Sadly, there is no good solution to the problem, other than dropping messages on overflow just like we do with our output channels.

@Stebalien @whyrusleeping thoughts on the issue?

vyzo:
Cherry picked from #67.
Closes #69

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -329,7 +329,7 @@ func (p *PubSub) notifySubs(msg *pb.Message) {
 			select {
 			case f.ch <- &Message{msg}:
 			default:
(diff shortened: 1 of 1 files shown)


vyzo:
Code climate complains, but I am not sure it makes sense to add a test here to pacify it.

PR #76, opened 2018-05-07
package.json: Import libp2p and not separate deps

hsanjuan:
This project builds on top of libp2p but imports separate
libp2p libraries instead of the main package. This makes it
difficult to use while using libp2p. i.e. when a host has been
created with libp2p.New() and the libp2p-peer version does not
match the one here.

### DIFF
--- a/.gx/lastpubver
+++ b/.gx/lastpubver
@@ -1,1 +1,1 @@
-0.9.13: QmduQtUFqdq2RhM84yM2mYsdgJRAH8Aukb15viDxWkZtvP
+0.9.14: QmPynuWZTkJA2aYNwL1SBMdHGeHAndxK5Ap1attWnxdfua
--- a/.gx/lastpubver
+++ b/.gx/lastpubver
@@ -1,1 +1,1 @@
-0.9.12: QmVKrsEgixRtMWcMd6WQzuwqCUC3jfLf7Q7xcjnKoMMikS
+0.9.13: QmduQtUFqdq2RhM84yM2mYsdgJRAH8Aukb15viDxWkZtvP
--- a/package.json
+++ b/package.json
@@ -31,5 +31,6 @@
   "license": "",
   "name": "go-libp2p-floodsub",
   "releaseCmd": "git commit -a -m \"gx publish $VERSION\"",
-  "version": "0.9.12"
+  "version": "0.9.13"
 }
+
--- a/package.json
+++ b/package.json
@@ -21,9 +21,9 @@
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmNh1kGFFdsPu79KNSaL4NUKUPb4Eiz4KHdMtFY6664RDp",
+      "hash": "QmWsV6kzPaYGBDVyuUfWBvyQygEc9Qrv9vzo8vZ7X4mdLN",
       "name": "go-libp2p",
-      "version": "5.0.14"
+      "version": "5.0.17"
     }
   ],
   "gxVersion": "0.9.0",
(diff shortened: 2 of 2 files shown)


hsanjuan:
Exactly doing this is the best way to avoid potential cycles, duplicate depedencies, version mismatches. Right now you have to maintain a bunch of libp2p subdependencies here manually which hopefully match those in the latest version of libp2p, those in go-ipfs, and those in every project that wants to use this. It's really bad UX to the external user... it should be clear for what libp2p version is used for something that builds on top...

vyzo:
Let's consider the big picture in the usage of the package and our dependency graph.

We don't want to get in a situation that we have to update libp2p to advance development in pubsub (and other packages with similar standing in the dependency graph, eg dht). 
This gets us back to a monolith, and only now it's distributed.

My opinion is that we shouldn't import libp2p directly in components, unless we have to construct a host or have other integrations.

Kubuxu:
For master:
```
 …/libp2p/go-floodsub  - master > gx deps stats     
Total Import Count: 3496
Unique Import Count: 67
Average Import Depth: 6.14
```

For gx/libp2p-dep:
```
 …/libp2p/go-floodsub - gx/libp2p-dep > gx deps stats
Total Import Count: 12298
Unique Import Count: 88
Average Import Depth: 7.20
```

hsanjuan:
Let's try to focus this from a user perspective please. Anyone using go-floodsub needs to pass in a host.Host. A libp2p basic host is created with go-libp2p. So please @Kubuxu explain to me how a reduced dep tree is important to me as a user if I necessarily need to import the whole of libp2p to make use of this library.

We want to commit to libp2p being usable and standing on its own, but this component requires fixing 8 hashes in order to make it work with any other project that is not precisely in the libp2p version for which the current dependency set is adjusted (and not having dups). This is not good. It is easy to miss the update of some hashes here too when the time comes and this only creates dups in the gx dependency tree.

I understand it eases your development if you can have dancing dependencies here. I understand it saves you some build time probably, and it's handy to adjust as you need. But it's does so by creating confusion and work for any other external user.

Kubuxu:
> It is easy to miss the update of some hashes here too when the time comes and this only creates dups in the gx dependency tree.

This means that our tooling is lacking in this regard and we should improve it (gx repos).
It doesn't imply a change of how libp2p ecosystem packages depend (which is a design decision) on each other.

Another reason against this change is with dependency on `go-libp2p` this package would have to be updated with every released change in `go-libp2p`. No matter how small bugfixes, transport changes and so on they are. Currently, this package depends only on interfaces which change much more rarely than dependencies of `go-libp2p` itself.

whyrusleeping:
@hsanjuan i'm not sure how this makes things easier. If i want to use floodsub, i import floodsub. Since floodsub is a libp2p service, you'll import libp2p anyways to construct your host, and then pass it into the floodsub constructor.

Is the problem that youre running into dependencies being out of sync often?

hsanjuan:
The problem is:

* I am using libp2p 5.0.14
* I import latest floodsub hash
* Version mismatches for things under libp2p (`libp2p-peer`) and doesn't build
* Not easy to see which tag of floodsub was compatible with 5.0.14
* If I fix problematic floodsub deps, still going to run into duplicate deps (harmless but ugly), or at the very least I'm going to have to check them one by one.
* Solution: this PR. just import go-libp2p, with the right version, and all my problems are gone because it will use the right versions for all things.
* Need to use a new version of libp2p? No problem, bubbling will catch this project too.

I'm trying to make the case that things built on top of libp2p (i.e. need a host to run), depend on go-libp2p.

I wouldn't have problems with the arguments exposed if this was a top-level library, with no dependencies on top. But this is meant to be integrated in other places that are going to depend on libp2p.

I can fix this for me easily enough. I just find the way it is now makes more complicated to use this lib.

hsanjuan:
Guys, I think all arguments are exposed and I haven't managed to convince much. Let's simply move on and keep hacking. <3 Sorry for the noise.

PR #79, opened 2018-05-29
Add and update info on gossipsub and implementations

jamesray1:
Note that you could merge this before or after #67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ TODO
 
 ## Implementations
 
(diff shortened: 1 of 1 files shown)


mhchia:
@jamesray1  hi James, thank you for mentioning.
However, I think I'm not working on gossipsub, I just test, use, and look into it.
In this case, I think it is better not to list me in the document.

jamesray1:
@mhchia, removed as requested!

mhchia:
thanks :)

PR #82, opened 2018-06-04
Make FloodSub Protocols Configurable

laser:
This changeset makes it possible for the floodsub router to use consumer-specified protocols.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -14,18 +14,26 @@ const (
 	FloodSubID = protocol.ID("/floodsub/1.0.0")
 )
 
-// NewFloodSub returns a new PubSub object using the FloodSubRouter
-func NewFloodSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
-	rt := &FloodSubRouter{}
+// New returns a new floodsub-enabled PubSub objecting using the protocols specified in ps
+func New(ctx context.Context, h host.Host, ps []protocol.ID, opts ...Option) (*PubSub, error) {
+	rt := &FloodSubRouter{
+		protocols: ps,
+	}
 	return NewPubSub(ctx, h, rt, opts...)
 }
 
+// NewFloodSub returns a new PubSub object using the FloodSubRouter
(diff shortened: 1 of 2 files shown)


floodsub.go
```diff
@@ -14,18 +14,26 @@ const (
 	FloodSubID = protocol.ID("/floodsub/1.0.0")
 )
 
+// New returns a new floodsub-enabled PubSub objecting using the protocols specified in ps
+func New(ctx context.Context, h host.Host, ps []protocol.ID, opts ...Option) (*PubSub, error) {
```
laser: This API is motivated by two things:

1) I didn't want to introduce a breaking change to your existing API
2) I didn't think that it was quite the time to significantly refactor the mechanism for configuring the FloodSub (to match what was recently incorporated into the DHT, for example)

laser:
That recent test failure looks suspicious to me (can't reproduce it locally) - but I don't have the authority to retry one of these builds. Will push an empty commit to see if it goes green, which we can then rebase out of existence if the build passes.

Stebalien:
That is an odd error. Maybe another libp2p node running on the same machine? No idea.

floodsub.go
Stebalien: So, given that we have both a FloodSub and a PubSub, I'd rather not call this `New` (a bit ambiguous`). It's a mouthful but how about `NewFloodsubWithProtocols`. Eventually, we can deprecate this but at least it won't clash with any names we may need in the future. For example, we may want to define a `New` constructor in the future that sets up the default pubsub routers.

laser: Sounds totally reasonable to me. Just made the change and pushed.

laser:
@Stebalien - Is there anything further you'd like to see in this PR before merging?

Stebalien:
(going to let @bigs review this first)

Stebalien:
Meh, this is tiny and we have users who really need it. Merging.

bigs:
Sounds good!

laser:
thanks, gents!

PR #87, opened 2018-07-13
Use proto3 syntax

### LINKED ISSUE #85: Floodsub warning: no syntax specified for the proto file
> ```
> protoc --rust_out . rpc.proto
> [libprotobuf WARNING google/protobuf/compiler/parser.cc:546] No syntax specified for the proto file: rpc.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)
> ```

https://github.com/libp2p/rust-libp2p/issues/287

prestonvanloon:
1. Updated proto syntax to `proto3`
1. Regenerated the pb.go
1. Fixed all spots where the proto was expecting the address of a string or boolean, but now expects the value itself. 

Fixes #85


prestonvanloon:
Looks like the proto package needs to be upgraded in the Travis setup

prestonvanloon:
I have no idea how to update gogo-protobuf to 1.0.0 with gx so this will continue to fail CI until it is updated from 0.0.0.

prestonvanloon:
This might cause some backwards incompatibility to update the proto to proto3, especially since other projects are using a copy of the proto definition. This would require updates across all projects. 

protoc is defaulting to proto2 anyway, so adding the explicit syntax declaration would silence the warning.

PR #95, opened 2018-08-09
some logging for gossipsub

vyzo:
The pubsub routers are very quiet, which can make it hard to diagnose the network state.
That's probably ok for floodsub, as the router doesn't do much, but gossipsub could be more verbose.

This pr adds some basic event logging to gossipsub router, so that we can have a clue about network state when debugging.


vyzo:
i will also gx publish so that it can be used immediately.

whyrusleeping:
sounds good!

PR #97, opened 2018-08-26
Implement message signing

vyzo:
Implements message signing, optionally enabled with the `WithMessageSigning` option:
- Adds two fields to the Message protobuf: signature and key.
- When the option has been specified, all published messages are signed with the host key.
- When a signature is present, it is always validated. If strict mode has been specified, then all unsigned messages are rejected without validation. Otherwise, it is up to the topic validator to reject unsigned messages.
- RSA keys are attached on signed messages, otherwise the key is extracted from the peer ID.

TBD:
- [X] tests


vyzo:
added some tests.

pubsub.go
```diff
@@ -500,22 +547,34 @@ loop:
 	}
 
 	if throttle {
-		log.Warningf("message validation throttled; dropping message from %s", src)
```
whyrusleeping, line 503: why drop the warning?

pubsub.go
```diff
@@ -500,22 +547,34 @@ loop:
...
-		return
+		return false
 	}
 
 	for i := 0; i < rcount; i++ {
 		valid := <-rch
 		if !valid {
-			log.Warningf("message validation failed; dropping message from %s", src)
```
whyrusleeping, line 510: again, dropping the warning feels weird

pubsub.go
vyzo: it gets logged on the return path, so this would be logging it twice.

pubsub.go
vyzo: ditto.

pubsub.go
vyzo: Basically the function doesn't do the dropping any more, just validation; so the logging has been moved to the caller.

sign.go
```diff
@@ -0,0 +1,82 @@
...
+	pid, err := peer.IDFromBytes(m.From)
+	if err != nil {
+		return err
+	}
+
+	if m.Key == nil {
+		// no attached key, it must be extractable from the source ID
+		pubk, err = pid.ExtractPublicKey()
+		if err != nil {
+			return err
```
whyrusleeping: maybe log a more specific error here. "could not validate because public key was not available or extractable: $ERR"

sign.go
```diff
@@ -0,0 +1,82 @@
...
+
+func verifyMessageSignature(m *pb.Message) error {
+	var pubk crypto.PubKey
+
+	pid, err := peer.IDFromBytes(m.From)
+	if err != nil {
+		return err
+	}
+
+	if m.Key == nil {
```
whyrusleeping, line 50: i'd pull this whole logic out into a 'get key' method either on the message, or that takes a message

sign.go
```diff
@@ -0,0 +1,82 @@
...
+			return err
+		}
+
+		// verify that the source ID matches the attached key
+		if !pid.MatchesPublicKey(pubk) {
+			return fmt.Errorf("bad signing key; source ID %s doesn't match key", pid)
+		}
+	}
+
+	xm := pb.Message{
```
whyrusleeping: I wonder if we can do something where we enforce that the protobuf puts the signature at the end, and simply slice the raw bytes to the right point instead of remarshaling

pb/rpc.proto
```diff
@@ -19,6 +19,8 @@ message Message {
 	optional bytes data = 2;
 	optional bytes seqno = 3;
 	repeated string topicIDs = 4;
+	optional bytes signature = 5;
```
whyrusleeping, line 22: in light of my other comment, how about we make key come before signature, and space them out a bit (instead of 5 and 6, use 50 and 51) so future additions to the protocol don't have to come after it.

cc @Stebalien @warpfork to tell me "no, you cant just slice protobufs like that"

whyrusleeping:
Approach looks good to me, nice and simple. My main concern now is making sure we don't shoot ourselves in the foot on perf. This code as written looks pretty allocation heavy.

pubsub.go
```diff
@@ -186,6 +190,13 @@ func WithValidateThrottle(n int) Option {
 	}
 }
 
+func WithMessageSigning() Option {
```
whyrusleeping: Maybe we should consider using different keys than our nodes peer key?

sign.go
vyzo: sure, that's probably a good idea.

sign.go
vyzo: ok

sign.go
vyzo: that would be very nice indeed -- i don't like the remarshal either.

pb/rpc.proto
vyzo: the key is not included in the signature, so it can stay where it is.

pubsub.go
vyzo: Sure, we can do that too -- the `from` in published messages will have to change to match the signing key though.

vyzo: I think we can have a `WithMessageSigningKey` option that explicitly specifies the signing key.

pb/rpc.proto
warpfork: I'm not sure I can say "you can't" but I probably "wouldn't" if I could help it.  I'd be a lot more comfortable if an envelope type `{content, sig}` was added.

I feel like holistically, even across wildly different formats, people tend to rue the day they made in-band signatures.  For (distant, but again, holistically comparable) example, android putting sigs on apks *inside* the filesystem has been no end of comedic problems over the years.

whyrusleeping: Yeah, I guess one way to do this would be to have in the top level protobuf both the existing regular unsigned messages, and a new field for signed messages.

sign.go
vyzo: we could also have a pre-allocated buffer where we `WriteMsg` into instead of marshaling.

Stebalien: 1. Let's not rely on slicing protobufs, if at all possible.
2. Please read: https://github.com/ipfs/notes/issues/249. We need to include *something* that says that the signature is for pubsub.

Regardless, we need to make sure we can add new fields and have them covered by the signature by default.

pubsub.go
Stebalien: Yeah, `from` can be any valid key. However, I'd prefer it if it were a peer ID (by default at least).

Stebalien: Also, this can't be optional.

sign.go
whyrusleeping: mmm, good point. We could prefix the data we're signing with something like "libp2p-pubsub"

vyzo: Hrm, not convinced it's absolutely necessary as this is transient data, but sure we can add a prefix.

pubsub.go
vyzo: Sure, ID works -- we can pull the key from the peerstore anyway.

> Also, this can't be optional.

Do you want to specify the ID in the option instead of relying on `host.ID()` for the default and using a separate option?

vyzo: We can always vararg an optional peer ID argument in the option.
If it is not included, use the host ID; otherwise use the specified ID to retrieve the key from the peerstore.

vyzo: > Also, this can't be optional.

@Stebalien did you mean you want message signing to be the default?

sign.go
vyzo: done

sign.go
vyzo: done

sign.go
vyzo: Added a `libp2p-pubsub:` prefix.

pubsub.go
```diff
@@ -452,7 +463,7 @@ func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
 	}
 	p.markSeen(id)
 
-	if len(vals) > 0 {
+	if len(vals) > 0 || msg.Signature != nil {
```
mhchia, line 478: A dumb question: Is it correct that `PubSub.processLoop` only focuses on dispatching events to workers, and leave computations(like this signature validation) done by goroutines, to avoid increasing response time?

pubsub.go
```diff
@@ -89,6 +90,9 @@ type PubSub struct {
 	peers        map[peer.ID]chan *RPC
 	seenMessages *timecache.TimeCache
 
+	// key for signing messages; nil when signing is disabled (default for now)
```
mhchia, line 93: Does it make sense to have a flag as a field of the struct `PubSub` indicating whether to verify the signatures or not and is set when the `PubSub` is initialized? This way the existing code can just check this flag, instead of checking `msg.Signature != nil`. Because in my imagination `msg.Signature == nil` should be an invalid behavior when authentication is required, while it is valid when authentication is not enabled.

vyzo: The signatures are always verified when present. 
The question is what happens when the signature is not present, in which case a validator can reject the message as unsigned when authentication is required.

We could have a flag that only accepts signed messages however.

pubsub.go
vyzo: Yes, this is correct. Also, concurrency -- not doing this validation in parallel wastes idle cores.

pubsub.go
mhchia: Ahh you're right. If the authentication is required, it is sufficient to just add the check for whether the signature is available in that case. Thank you!

pubsub.go
Stebalien: Given that we deduplicate messages based on `(from, seqno)`, message signing must be mandatory.

vyzo: How about we have a `strict` flag as an argument to `WithMessageSigning`?
When true, then reject unsigned messages. No default, the user gets to specify behaviour.

Stebalien: We shouldn't *enforce* message signing by default, for now, unless we're willing to break things. However, we should *always* sign messages (or switch to deduplicating based on the hash).

vyzo: Signing without enforcing the signatures doesn't add anything over not siging wrt to message ID collision attacks. 
I think we should have the strict option when enabling it, but keep it disabled by default (at least in the near term).

vyzo:
rebased on master.

pubsub.go
Stebalien: > Signing without enforcing the signatures doesn't add anything over not siging wrt to message ID collision attacks.

Sorry, I was being lazy instead of explicit.

IMO:

1. We should *always* sign.
2. For *now*, we should probably disable verification by default for backwards compatibility (unless we're willing to break things).
3. We should definitely provide some form of strict option, defaulting to false. We'll eventually switch it to true by default and then, make it a no-op (someday).

vyzo: 1. I don't know about this, it has perfomance implications.
2. that's the approach here, so good.
3. we can just add the strict option to `WithMessageSigning` which lets the user in the loop.  we can make it have a default value by making it a variadic interface.

vyzo: @whyrusleeping thoughts on the "always sign" matter?

vyzo: Also, we don't have to make it "awlays sign" in the first step; let's add the signatures, have some usage to see how it fares in practice perf wise, and then we can decide whether to always sign in a later iteration.
I think it's a bit too drastic to go "always sign" in one step.

vyzo: Implemented strict mode with a flag in the `WithMessageSigning` option.

Stebalien: > I don't know about this, it has perfomance implications.

I'd *rather* use CIDs or hashes to identify messages but we need to do something.

vyzo: Well, ideally we would use message signing with strict mode.
I would argue that it's sane to keep the default behaviour unchanged (for a while at least), and provide the easy upgrade path of `WithMessageSigning(true)` for applications that need to secure their pubsub.
Eventually we can make it the default.

I don't think we need to take reflexive action and dump our simple message id construction just yet -- note that using hashes also has _some_ perf implications.

Stebalien: > I don't think we need to take reflexive action

Either is fine by me. My preference is to layer things (handle authorship at a higher layer) but it *is* useful to have that baked into the bottom layer.

---

> I would argue that it's sane to keep the default behaviour unchanged (for a while at least), and provide the easy upgrade path of `WithMessageSigning(true)` for applications that need to secure their pubsub.

I'm not worried about "security", I'm worried about griefing by bored children (e.g., what happens on IRC). However, as long as we turn it on by default in go-ipfs (with strict off), I can live with it.

vyzo: Makes sense to enable it in go-ipfs, I'll add the relevant options when we integrate it.

pb/rpc.proto
bigs: that also kind of disrupts the chronology of the development of the protobuf. i think we'd probably be best off keeping it as is

bigs (approved):
i really like this change and think it's cleanly implemented. could we maybe make this configurable behavior for now so we can do some testing in the IPFS cluster? i can't imagine it'd be a huge drain on perf relative to the usefulness it adds to the end user

sign.go
```diff
@@ -0,0 +1,101 @@
...
+
+const SignPrefix = "libp2p-pubsub:"
+
+func verifyMessageSignature(m *pb.Message) error {
+	pubk, err := messagePubKey(m)
+	if err != nil {
+		return err
+	}
+
+	xm := pb.Message{
```
Stebalien: This is fine if we're never planning on adding additional fields. Otherwise, we should probably copy the *entire* message, remove just the signature (and key), and remarshal that. That should keep the unknown fields.

sign.go
```diff
@@ -0,0 +1,101 @@
...
+	bytes = withSignPrefix(bytes)
+
+	sig, err := key.Sign(bytes)
+	if err != nil {
+		return err
+	}
+
+	m.Signature = sig
+	switch key.Type() {
+	case crypto.RSA:
```
Stebalien: Ah. Actually, the "correct" way to do this is to:

1. Generate a peer ID.
2. Try to extract the key from the peer ID (`ExtractPublicKey`).

We may (will) have other key types that don't "embed".

sign.go
vyzo: It's probably more futureproof to do it that way, although ... message copying again.

sign.go
vyzo: It's also a lot more slower. Are we planning on adding other non-embeddable keys in the future?

vyzo: maybe that overhead is negligible in the grand scale of things though.

sign.go
Stebalien: It should be as efficient or more efficient than the current code. I'm suggesting:

```
xm := *m
xm.Signature = nil
xm.Key = nil
```

sign.go
Stebalien: We could add some "Embeds(key) bool` helper somewhere.

>  Are we planning on adding other non-embeddable keys in the future?

I don't think sekp256 keys embed and there's a PR in progress to add more general ECDSA keys that definitely won't embed.

sign.go
vyzo: Sure, we can do that.

sign.go
vyzo: Ok, but let's define a helper function in go-libp2p-peer.

edit: or go-libp2p-crypto, but somewhere more general than here.

sign.go
vyzo: done.

vyzo:
@whyrusleeping 

> My main concern now is making sure we don't shoot ourselves in the foot on perf. 

Agreed! I am concerned that we start entering the validation pipeline with it.

> This code as written looks pretty allocation heavy.

I think that spawning an (admittedly very short-lived) goroutine for each message to validate the signatures should be a bigger concern.
Maybe we should pre-spawn `NUMCPU` goroutines to handle the validation instead and feed them through a buffered channel; if the buffer fills, we should drop the message.

pubsub.go
```diff
@@ -598,14 +666,19 @@ func (p *PubSub) GetTopics() []string {
...
-		&pb.Message{
-			Data:     data,
-			TopicIDs: []string{topic},
-			From:     []byte(p.host.ID()),
-			Seqno:    seqno,
-		},
+	m := &pb.Message{
+		Data:     data,
+		TopicIDs: []string{topic},
+		From:     []byte(p.host.ID()),
```
Stebalien, line 675: This needs to be the ID of the key we're using to sign.

vyzo: indeed, good catch.

vyzo: it's also not a bug in the current implementation, as we don't allow arbitrary keys yet.

vyzo: I will fix it anyway, so that it's futureproof (in case we want to allow arbitrary keys in the future)

vyzo: done.

vyzo:
Per discussion with @whyrusleeping on zoom, we decided that the goroutine issue should not block merging this, but it's definitely something we want to fix.
I'll make a follow up issue and pr.

vyzo:
Follow up in #103.

sign.go
vyzo: for now I added an extraction test.

vyzo:
rebased on master

Stebalien (approved):

PR #100, opened 2018-08-28
fix data race in gossipsub piggybacking

### LINKED ISSUE #98: Data race in gossipsub piggybacking
`GossipSubRouter.sendRPC` calls `gs.piggybackGossip` on a shared RPC object. It should probably copy it first.

vyzo:
Fixes data race in ownership of RPC when piggybacking control messages.
Closes #98

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -124,17 +124,20 @@ func rpcWithControl(msgs []*pb.Message,
 }
 
 func copyRPC(rpc *RPC) *RPC {
-	return &RPC{
+	res := &RPC{
 		RPC: pb.RPC{
 			Subscriptions: rpc.Subscriptions,
 			Publish:       rpc.Publish,
-			Control: &pb.ControlMessage{
-				Ihave: rpc.Control.Ihave,
-				Iwant: rpc.Control.Iwant,
-				Graft: rpc.Control.Graft,
-				Prune: rpc.Control.Prune,
-			},
 		},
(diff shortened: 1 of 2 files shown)


vyzo:
We should probably gx publish and bubble up this one to ipfs, it's a nasty bug.

comm.go
```diff
@@ -122,3 +122,22 @@ func rpcWithControl(msgs []*pb.Message,
 		},
 	}
 }
+
+func copyRPC(rpc *RPC) *RPC {
+	res := &RPC{
```
Stebalien:
```go
res := new(RPC)
*res = *rpc
if rpc.Control != nil {
    *res.Control = *rpc.Control
}
```

Stebalien (commented):
Does it really cost *that* much to create a new RPC per peer?

comm.go
vyzo: Hrm, we still need to create the control object -- it's a pointer.

vyzo: But still, more idiomatic and terse.

vyzo:
No, it doesn't cost all that much if it is to fix a bug :)

Stebalien (approved):

Stebalien:
> Does it really cost that much to create a new RPC per peer?

My point was that we could create a new one per peer in `Publish`, if that's the only real issue. But we can change that later.

PR #102, opened 2018-08-29
Add control parameter for gossipsub initial heartbeat delay

### LINKED ISSUE #101: delay for heartbeat in tests is inconvenient
In tests that use gossipsub, we have to sleep for a bit to wait for the heartbeats to set everything up properly. This is pretty inconvenient, as adding two seconds of wait time to every single test gets pretty expensive.

Is there a way to force this to happen immediately?

cc @vyzo

vyzo:
So that tests can bootstrap quickly; also makes the default behaviour be 100ms initial delay (instead of 1s)
Closes #101

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -27,7 +27,8 @@ var (
 	GossipSubHistoryGossip = 3
 
 	// heartbeat interval
-	GossipSubHeartbeatInterval = 1 * time.Second
+	GossipSubHeartbeatInitialDelay = 100 * time.Millisecond
+	GossipSubHeartbeatInterval     = 1 * time.Second
 
 	// fanout ttl
 	GossipSubFanoutTTL = 60 * time.Second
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -27,7 +27,8 @@ var (
 	GossipSubHistoryGossip = 3
 
 	// heartbeat interval
-	GossipSubHeartbeatInterval = 1 * time.Second
+	GossipSubHeartbeatInitialDelay = 100 * time.Millisecond
```
mhchia, line 30: Just a quick question, in the tests I saw `100 * time.Millisecond` many times. Is it a measured safe delay for initial connections between peers and all the other needed functionalities in host and Swarm to set up? Thank you

vyzo: it's probably over-generous and not really measured, but it's a ballparked initialization time for stuff.
You can probably get away with smaller delays most of the time, but the gossipsub tests make a lot of peers.

mhchia: It makes a lot of sense. Thank you for the explanation.

Stebalien (approved):

whyrusleeping (approved):

PR #105, opened 2018-09-14
Fix bug where peers connecting to one another simultaneously never establish a pubsub stream

### LINKED ISSUE #93: Peers not always joining a topic and/or connecting
We've been running into some hickups in OrbitDB recently as we've been upgrading from go-ipfs 0.4.13 to 0.4.16 (and now 0.4.17).

We've observed a behaviour where pubsub peers don't always connect to each other. Detailed information on the behaviour and what we've discovered so far in https://github.com/orbitdb/orbit-db/pull/417. To summarize: the tests are run with two local go-ipfs nodes, both subscribe to the same pubsub topic, but sometimes (about 50% of test runs) they don't seem to connect to each other on pubsub level even though they're connected on swarm level (see this comment https://github.com/orbitdb/orbit-db/pull/417#issuecomment-407002661).

Looking at the commit log in this repo, I noticed that the pubsub implementation was changed (yay, great to see the gossip pubsub coming together!) some time ago, so most likely this is regression. I believe 0.4.13 was working fine, but we haven't verified which version after that doesn't work anymore.

If anyone has an idea why this is happening and work on a fix, would highly appreciate it. We're (finally) adding back the support for go-ipfs in OrbitDB, but the tests are not running reliably at all times due to this bug, so we're hesitant to make a new release until this is fixed (alternatively we could instruct user to use =< 0.4.13 but that's no fun). 

If there's any information I can provide to solve this, please let me know!

bigs:
This patch establishes a failing test and then provides patch for a subtle bug, where two peers connecting to one another at the same time fail to establish an active pubsub stream. This bug is rooted in the fact that both peers tend to get a notification that their OUTBOUND dial succeeds before their peer's INBOUND dial. As a result of the old logic, both peers would then happily throw away the stream they established on their outbound connection to favor the peer's inbound connection, resulting in both peers holding on to bad streams.

This fix is a bit heavy handed, but the logic was the most sound I could muster. My patch changes behavior such that peers favor a connection initiated by the peer with the greatest peer ID (evaluated in integral space).

This branch closes #93.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -87,7 +87,8 @@ type PubSub struct {
 	// eval thunk in event loop
 	eval chan func()
 
-	peers        map[peer.ID]*rpcpair
+	peers        map[peer.ID]chan *RPC
+	peerStreams  map[peer.ID]inet.Stream
 	seenMessages *timecache.TimeCache
 
 	ctx context.Context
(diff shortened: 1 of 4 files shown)


bigs:
in particular, i could see storing an `inet.Stream` in the `rpcpair` instead of the `peer.ID`, though i strayed away from that because stream closing/reset is handled elsewhere in the code.

pubsub.go
```diff
@@ -86,7 +87,7 @@ type PubSub struct {
 	// eval thunk in event loop
 	eval chan func()
 
-	peers        map[peer.ID]chan *RPC
+	peers        map[peer.ID]*rpcpair
```
vyzo: we shouldn't need to change this type (and the floodsub/gossipsub code) at all.

vyzo (commented):
this very ugly...

vyzo:
Let's just keep the connection with the highest ID in key space. 
I don't think we need this rpcpair ugliness - you shouldn't have to change the peers map (and any code in floodsub/gossipsub) at all. The higher ID can be deduced by the key in the map, compared to our own ID.

bigs:
tagged as blocked until we can decide on a better fix at swarm level

bigs:
@Stebalien this is a kind of complicated issue. connection de-duplication only seems to happen at the `Host` level (i.e. `BasicHost#Connect`) and it's not race safe at all. it simply queries `Connectedness` before starting a dial. i'm actually not so sure what would be the best course of action. my gut is shifting towards keeping the fix in floodsub (& generally protocol implementation) territory. users should have some degree of freedom over what connection to a peer they want to use.

vyzo:
Seems it's an instance of this: https://github.com/libp2p/go-libp2p-swarm/issues/79

Stebalien:
@vyzo same deal but not quite the same, as far as I know. However, in *this* case, I think we can just use two streams, right? We could *also* disambiguate but IMO, two streams is the correct approach.

bigs:
yeah that’s my plan. should we assume users will deal with reconnection?

Stebalien:
Reconnection? In this case, I'd just have both sides use their own stream for sending, that should "just work".

bigs:
closing this as it's plain wrong!

PR #111, opened 2018-10-10
randomsub implementation

vyzo:
So that we have the whole zoo; people actually want to use it, even if it's just for testing.

### DIFF
--- a/randomsub.go
+++ b/randomsub.go
@@ -34,7 +34,7 @@ type RandomSubRouter struct {
 }
 
 func (rs *RandomSubRouter) Protocols() []protocol.ID {
-	return []protocol.ID{RandomSubID, RandomSubID}
+	return []protocol.ID{RandomSubID, FloodSubID}
 }
 
 func (rs *RandomSubRouter) Attach(p *PubSub) {
--- a/randomsub.go
+++ b/randomsub.go
@@ -1,0 +1,108 @@
+package floodsub
+
+import (
+	"context"
+
+	pb "github.com/libp2p/go-floodsub/pb"
+
+	host "github.com/libp2p/go-libp2p-host"
+	peer "github.com/libp2p/go-libp2p-peer"
+	protocol "github.com/libp2p/go-libp2p-protocol"
+)
+
+const (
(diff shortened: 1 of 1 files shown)


bigs:
u crazy for this one

randomsub.go
```diff
@@ -0,0 +1,108 @@
...
+				tosend[p] = struct{}{}
+			} else {
+				rspeers[p] = struct{}{}
+			}
+		}
+	}
+
+	if len(rspeers) > RandomSubD {
+		xpeers := peerMapToList(rspeers)
+		shufflePeers(xpeers)
```
bigs, line 80: unnecessary, as in golang map access is randomized!

bigs: we can just count how many we've iterated through and break when we've hit our limit

bigs (changes requested):

randomsub.go
vyzo: afaik it is simply non-deterministic, not randomized between iterations.

bigs (approved):

randomsub.go
vyzo: more precisely, it simply starts from a different (random) point in the map on each iteration -- the relative order of the keys will not change if there haven't been any insertions in the map.

raulk:
Would be good to get a test for this at some point.

raulk:
(not intending to be the party pooper 😛)

vyzo:
feel free to write one :)

raulk:
@vyzo here it is: https://github.com/raulk/go-floodsub/commit/8ee2d829d720d322360960a46be665ac01b31bc6

Could you please check out why it fails with randomsub but succeeds with floodsub and gossipsub?

Perhaps randomsub doesn't have the expected emergent properties, but maybe there's something wrong with the test!

EDIT: Let me know when you want it in the codebase, and I'll send in a PR.

vyzo:
@raulk It's not unexpected that it would lose some messages for some peers -- there is a non-zero probability that the random propagation will not deliver a message to a node. That's why we don't use it in production (and the main reason for why I didn't bother with the test in the first place).

PR #115, opened 2018-10-17
Keep line width to 79 chars

jamesray1:
To be able to see two files on one screen. I've changed tab widths to 4 chars.

### DIFF
--- a/pb/rpc.proto
+++ b/pb/rpc.proto
@@ -3,80 +3,80 @@ syntax = "proto2";
 package floodsub.pb;
 
 message RPC {
-	repeated SubOpts subscriptions = 1;
-	repeated Message publish = 2;
+    repeated SubOpts subscriptions = 1;
+    repeated Message publish = 2;
 
-	message SubOpts {
-		optional bool subscribe = 1; // subscribe or unsubcribe
-		optional string topicid = 2;
-	}
+    message SubOpts {
+        optional bool subscribe = 1; // subscribe or unsubcribe
+        optional string topicid = 2;
+    }
 
     optional ControlMessage control = 3;
 }
 
 message Message {
-	optional bytes from = 1;
-	optional bytes data = 2;
-	optional bytes seqno = 3;
-	repeated string topicIDs = 4;
-	optional bytes signature = 5;
-	optional bytes key = 6;
+    optional bytes from = 1;
+    optional bytes data = 2;
+    optional bytes seqno = 3;
+    repeated string topicIDs = 4;
+    optional bytes signature = 5;
+    optional bytes key = 6;
 }
 
 message ControlMessage {
-	repeated ControlIHave ihave = 1;
-	repeated ControlIWant iwant = 2;
-	repeated ControlGraft graft = 3;
-	repeated ControlPrune prune = 4;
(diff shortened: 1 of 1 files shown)


pb/rpc.proto
```diff
@@ -11,45 +11,44 @@ message RPC {
...
-	optional ControlMessage control = 3;
+    optional ControlMessage control = 3;
 }
 
 message Message {
 	optional bytes from = 1;
 	optional bytes data = 2;
 	optional bytes seqno = 3;
 	repeated string topicIDs = 4;
-	optional bytes signature = 5;
```
lanzafame, line 22: Were these two lines meant to be deleted?

lanzafame (commented):
There is now mixed indentation, and the comments are harder to grok because of the line break, plus if anyone were to fix the indenting, i.e. `gg=G` in vim, then the comments are no longer obviously associated.

pb/rpc.proto
jamesray1: Oh, no, looks like those lines have been added some time recently, I was editing from an older file. :S

jamesray1:
@lanzafame that's weird, the rendering on GitHub is different on VS Code and in the GitHub editor, where the indentation is lined of comments are lined up.

lanzafame:
Yeah that's because Github and VS Code will render tabs at a different size, see image for how tabs are now used for TopicDescriptor to the end of the file:

Also because of the tabs between the field and the comments more comment than previously necessary needs to be on the next line down. It is almost not worth it in my opinion.

pb/rpc.proto
```diff
@@ -11,45 +11,44 @@ message RPC {
...
-	optional bytes key = 6;
```
vyzo, line 23: Please don't remove those!

vyzo (changes requested):
Do we really care to keep things in 79 chars?
We are not using ttys anymore and most people edit in much wider screen areas.

Also, the indentation for the comments now looks very strange.

vyzo:
I would like to add that I am very much against this change; we don't have any char limitation anywhere else and it really doesn't make much sense in 2018.

lanzafame:
At this point, the readability is being sacrificed for 'ideal' of 79 char limit. Especially, when this can be solved in editor-space by a soft wrap toggle hotkey or shrinking the text size for the time it takes to read the 3 words that go over the 79 char limit.

jamesray1:
> Especially, when this can be solved in editor-space by a soft wrap toggle hotkey or shrinking the text size for the time it takes to read the 3 words that go over the 79 char limit.

I wasn't aware of the former in VS code, but looked into it and I see that you can use Alt+Z. I tried to shrink and expand selected text but that didn't work. I didn't think zoom in and out was a feature in VS Code, but I checked again and you can use Ctrl+-/+.

jamesray1:
Having to zoom out (and squint) or keep scrolling across isn't ideal, however.

lanzafame:
Agreed, hence why I suggested a soft wrap toggle first as it doesn't
require scrolling across or squinting.

jamesray1:
Right, whoops, momentarily forgot about that! Thanks!

PR #121, opened 2018-10-25
rework message signing options

Stebalien:
Previously, there was no non-breaking upgrade path to enable message signing by default while still providing a way to disable it. This patch breaks the options into:

1. WithMessageSigning - actually enable message signing.
2. WithStrictSignatureVerification - enforce strict signature verification.
3. WithMessageAuthor - set the message author (independent of signing).

Now, users can explicitly set all three options and any changes we make to the defaults will never override these explicit settings.

This patch *also* enables *signing* by default but not strict signature verification. We obviously can't enable verification by default yet but I see no reason not to enable signing. Basically, we're either hurting performance or security by default. Security issues are generally only noticed once it's too late. Performance issues can be caught ahead of time.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -236,8 +236,8 @@ func WithMessageAuthor(author peer.ID) Option {
 // WithStrictSignatureVerification enforces message signing. If set, unsigned
 // messages will be discarded.
 //
-// This currently defaults to false but as we transition to signing by default
-// but will eventually default to true.
+// This currently defaults to false but, as we transition to signing by default,
+// will eventually default to true.
 func WithStrictSignatureVerification(required bool) Option {
 	return func(p *PubSub) error {
 		p.signStrict = required
--- a/pubsub.go
+++ b/pubsub.go
@@ -147,6 +147,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		host:             h,
 		ctx:              ctx,
 		rt:               rt,
+		signID:           h.ID(),
+		signKey:          h.Peerstore().PrivKey(h.ID()),
 		incoming:         make(chan *RPC, 32),
 		publish:          make(chan *Message),
 		newPeers:         make(chan inet.Stream),
@@ -187,6 +189,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	return ps, nil
 }
 
+// WithValidateThrottle sets the upper bound on the number of active validation
+// goroutines.
 func WithValidateThrottle(n int) Option {
 	return func(ps *PubSub) error {
 		ps.validateThrottle = make(chan struct{}, n)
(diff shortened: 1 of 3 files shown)


pubsub.go
```diff
@@ -187,18 +189,58 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
+func WithMessageSigning(enabled bool) Option {
 	return func(p *PubSub) error {
-		p.signID = p.host.ID()
-		p.signKey = p.host.Peerstore().PrivKey(p.signID)
-		p.signStrict = strict
+		if enabled {
+			p.signKey = p.host.Peerstore().PrivKey(p.signID)
+			if p.signKey == nil {
+				return fmt.Errorf("can't sign for peer %s: no private key", p.signID)
+			}
```
vyzo, line 208: Shouldn't we set the signID as well for consistency?
Also, how well does it combine with the `WithMessageAuthor` option? Seems that this needs to come first.

pubsub.go
```diff
@@ -187,18 +189,58 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
+
+// WithMessageAuthor sets the author for outbound messages to the given peer ID
+// (defaults to the host's ID). If message signing is enabled, the private key
+// must be available in the host's peerstore.
+func WithMessageAuthor(author peer.ID) Option {
+	return func(p *PubSub) error {
+		if author == "" {
+			author = p.host.ID()
+		}
+		if p.signKey != nil {
```
vyzo, line 224: not sure that's the correct way to discern whether signing is enabled -- maybe we should have a flag instead.

pubsub.go
Stebalien: Either way should work.

* WithMessageSigning just uses the existing signID (defaults to the host ID).
* WithMessageAuthor *sets* the signID.

---

* If you pass WithMessgaeSigning first, WithMessageAuthor will update the signKey.
* If you pass WithMessageAuthor first, WIthMessageSigning will use that author.

pubsub.go
Stebalien: Either way. This is how we *currently* do it, as far as I can tell. Basically, if we've set a signing key, we use it. Otherwise, we don't sign anything.

vyzo: Right, but we didn't have to contend with an arbitrary author key.

pubsub.go
vyzo: ok, that's reasonable.

pubsub.go
```diff
@@ -187,18 +189,58 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
+		p.signID = author
+		return nil
+	}
+}
+
+// WithStrictSignatureVerification enforces message signing. If set, unsigned
+// messages will be discarded.
+//
+// This currently defaults to false but as we transition to signing by default
+// but will eventually default to true.
```
vyzo: typo - double but in the sentence.

vyzo (approved):

pubsub.go
Stebalien: I'd be open to doing this some other way. My only real requirement is that we catch not having the appropriate key early.

vyzo: It's fine this way, it's just configuration and as long as the options work as advertised it doesn't matter how the sausages are made :)

PR #122, opened 2018-10-25
forbid sign strict without signing enabled

Stebalien:
see https://github.com/ipfs/go-ipfs/pull/5647/files#r228366143

### DIFF
--- a/floodsub_test.go
+++ b/floodsub_test.go
@@ -900,6 +900,16 @@ func assertPeerList(t *testing.T, peers []peer.ID, expected ...peer.ID) {
 	}
 }
 
+func TestNonsensicalSigningOptions(t *testing.T) {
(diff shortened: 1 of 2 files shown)


floodsub_test.go
```diff
@@ -900,6 +900,16 @@ func assertPeerList(t *testing.T, peers []peer.ID, expected ...peer.ID) {
 	}
 }
 
+func TestNonsensicalSigningOptions(t *testing.T) {
```
schomatis, line 903: :grin:

schomatis (approved):

vyzo:
Not sure how useful this is, as if we disable message signing then the strict option will be ignored.

vyzo (approved):

Issue #130, opened 2018-12-12
Faulty peer disconnection notification logic


vyzo:
If there are multiple connections to the same peer, then the disconnect notification is emitted even though the peer is still connected, breaking gossipsub's peer tracking.
See the logs in https://github.com/status-im/nim-libp2p/issues/8

vyzo:
Also of note that the peer connection event is emitted twice as well.

vyzo:
related: #117

vyzo:
Upon reflection, the issue here is that our peer management is completely incapable of handling multiple connections to the same peer.
The connection notification is emitted twice, which results in two sets of readers and writers, each of which can emit a peerDead event on failure.
So once the host closes one of the two connections, we are guaranteed to declare the peer dead even if the peer is still connected with the other connection.

We will need to completely rework on peer management logic it seems.

vyzo:
cc @cheatfate

vyzo:
Source issue confirmed as fixed by @cheatfate; closing.

PR #132, opened 2018-12-13
rework peer tracking logic to handle multiple connections

vyzo:
Multiple connections are simply not handled at all by the current code; see https://github.com/libp2p/go-libp2p-pubsub/issues/130.

This is an attempt to rework the peer tracking logic so that multiple connections are handled gracefully.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -316,15 +316,18 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				continue
 			}
 
+			close(ch)
+
 			if p.host.Network().Connectedness(pid) == inet.Connected {
 				// still connected, must be a duplicate connection being closed.
 				// we respawn the writer as we need to ensure there is a stream active
 				log.Warning("peer declared dead but still connected; respawning writer: ", pid)
-				go p.handleNewPeer(ctx, pid, ch)
+				messages := make(chan *RPC, 32)
+				go p.handleNewPeer(ctx, pid, messages)
+				p.peers[pid] = messages
 				continue
 			}
 
-			close(ch)
 			delete(p.peers, pid)
 			for _, t := range p.topics {
 				delete(t, pid)
--- a/pubsub.go
+++ b/pubsub.go
@@ -57,8 +57,14 @@ type PubSub struct {
 	// send subscription here to cancel it
 	cancelCh chan *Subscription
 
-	// a notification channel for incoming streams from other peers
-	newPeers chan inet.Stream
+	// a notification channel for new peer connections
+	newPeers chan peer.ID
+
+	// a notification channel for new outoging peer streams
+	newPeerStream chan inet.Stream
+
+	// a notification channel for errors opening new peer streams
+	newPeerError chan peer.ID
 
 	// a notification channel for when our peers die
 	peerDead chan peer.ID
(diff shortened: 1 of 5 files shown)


vyzo:
cc @cheatfate

Stebalien:
I believe this is still going to be pretty racy:

**Hello Packet**

An old `handleSendingMessage` can stick around after we replace an old connection. This old handler can read the hello packet and throw it into oblivion.

Fixes:

1. `handleSendingMessage` should probably request a hello packet directly instead of having it sent on the general `outgoing` channel (that could be shared).
2. Don't share the `outgoing` channel. That is, when we start a new peer handler, we should close the old outgoing channel and open a new one. This will kill off the old `handleSendingMessage` handler.

Just 2 may be sufficient but I'm not sure.

---

This will also get rid of that sleep...

vyzo:
2 should be sufficient I think, will update.

pubsub.go
```diff
@@ -259,28 +267,53 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+			if !ok {
+				continue
+			}
+
+			close(ch)
+
+			if p.host.Network().Connectedness(pid) == inet.Connected {
+				// still connected, must be a duplicate connection being closed.
+				// we respawn the writer as we need to ensure there is a stream active
+				log.Warning("peer declared dead but still connected; respawning writer: ", pid)
```
Stebalien, line 311: There's a chance we'll hit this frequently the connection may close before we realize we're disconnected. It may not be ab issue but we should watch for this.

pubsub.go
```diff
@@ -259,28 +267,53 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+			if !ok {
+				log.Warning("new stream for unknown peer: ", pid)
+				s.Reset()
+				continue
+			}
+
 			p.rt.AddPeer(pid, s.Protocol())
 
+		case pid := <-p.newPeerError:
+			delete(p.peers, pid)
```
Stebalien, line 298: There's a race here where:

1. We have an old zombie connection that doesn't know it's dead yet.
2. We get a new connection.

With the current logic, we may still try to open a stream using the old connection. We'll then never try to create one with the new connection.

Unfortunately, this is a pretty common situation when a connection dies and we replace it immediately.

Maybe we should drop this channel and send a notification along `peerDead` instead? If we still think we're connected, we'll continue to retry until we realize we're actually disconnected.

Stebalien (changes requested):

Stebalien:
(why does connection state logic have to be this damn hard...)

pubsub.go
vyzo: This is only used when the stream failed to open; we can't pass it through the `peerDead` channel because the peer may simply not support the pubsub protocols.

vyzo: The problem here is if the stream fails to open because the connection was closed (if it was a duplicate connection). 
Is there a way to discriminate whether the failure was an unsupported protocol?
In this case we can notify on `peerError` for unsupported protocol and `peerDead` for connection closed errors.

vyzo: named error in `go-multistream.ErrNotSupported` can be used for this per discussion on irc.

vyzo: done.

Stebalien (approved):

iulianpascalau:
Hello, fetched the new version directly from the master branch, ran the same test 10 times, error log did not appear in any of the tests. 
We can consider this issue also closed. 
Big thanks!

PR #137, opened 2018-12-27
Add a message ID and hash to Message

jamesray1:


vyzo:
no, it's already in the RPC envelope.

vyzo:
Correcting myself, the `seqno` is in the RPC envelope, from which it gets derived.

Issue #142, opened 2019-01-09
basic pubsub test - fail to make it pass


nikkolasg:
Hi, I wanted to play with pubsub a bit, with a very basic scenario:
- 3 peers all subscribe to the same topic "random"
- peer 1 is connected to 2 and peer 2 is connected to 3
- peer 1 publishes some data for the "random" topic

I would expect that peer 2 would receive the message and also transmit it to peer 3 but in my code only peer 2 does.

I fail to find the relevant missing piece of info in the library... 
Is my scenario a valid one or is this expected to fail according to the specs ? And if it is a valid one, would you know why it does not work with my code below ?

I'd be happy to add lots of comments and make a PR to show such a basic example.

PS: I added the `DisableRelay` option since otherwise that does not work, as shown in the issue https://github.com/libp2p/go-libp2p-examples/issues/21

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"time"

	libp2p "github.com/libp2p/go-libp2p"
	host "github.com/libp2p/go-libp2p-host"
	pstore "github.com/libp2p/go-libp2p-peerstore"
	pubsub "github.com/libp2p/go-libp2p-pubsub"
)

const gossipSubID = "/meshsub/1.0.0"

func main() {

	//golog.SetAllLoggers(gologging.DEBUG) // Change to DEBUG for extra info
	h1 := newHost(2001)
	h2 := newHost(2002)
	h3 := newHost(2003)
	fmt.Printf("host 1: \n\t-Addr:%s\n\t-ID: %s\n", h1.Addrs()[0], h1.ID().Pretty())
	fmt.Printf("host 2: \n\t-Addr:%s\n\t-ID: %s\n", h2.Addrs()[0], h2.ID().Pretty())
	fmt.Printf("host 3: \n\t-Addr:%s\n\t-ID: %s\n", h3.Addrs()[0], h3.ID().Pretty())

	time.Sleep(100 * time.Millisecond)

	// add h1 to h2's store
	h2.Peerstore().AddAddr(h1.ID(), h1.Addrs()[0], pstore.PermanentAddrTTL)
	// add h2 to h1's store
	h1.Peerstore().AddAddr(h2.ID(), h2.Addrs()[0], pstore.PermanentAddrTTL)
	// add h3 to h2's store
	h2.Peerstore().AddAddr(h3.ID(), h3.Addrs()[0], pstore.PermanentAddrTTL)
	// add h2 to h3's store
[...]

PR #143, opened 2019-01-10
Basic example with README

nikkolasg:
This PR is a proposition to add a basic example on how to use the pubsub library. Since I had to go through the tests to find out, it may be helpful for others to have a clear and simple example first.
The code is from the issue https://github.com/libp2p/go-libp2p-pubsub/issues/142 
It's just a proposition, feel free to postpone if you had other plans (other repo, etc).

**NOTE** It does not work when the connections (line 55 - 57) happen *before* creating the `PubSub` and registering the topic.. I don't know why. Is this expected ? (@vyzo I tried increasing the timeout but it still hangs).

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -1,0 +1,66 @@
+# Basic PubSub Example 
+
+The `main.go` exposes the basic methods to play with the PubSub api. 
+It creates a simple scenario with three peers (1, 2 & 3) with the following
+network topology:
+```
+1 <---> 2 <---> 3
+```
+
+**Goal**: Node 1 sends a message to the *"random"* topic and node 2 and 3 must
+receive the message.
+
+# Code explanation
+
+1. **Host Creation**: We first need to create the three hosts using [go-libp2p]("https://github.com/libp2p/go-libp2p"). A `Host` is the central interface in libp2p that is able to connect to other `Host` (w/o encryption, signing, etc), relay messages, etc.
+```go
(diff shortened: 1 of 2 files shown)


vyzo:
Do you mind adding this to `go-libp2p-examples` instead?
That is the central place for libp2p example code, and we don't have anything for pubsub yet.

Note the reason it doesn't work if you connect before creating the pubsub is that there are no connection notifications, so the peers are not discovered.

nikkolasg:
> Do you mind adding this to go-libp2p-examples instead?

Sure, will do :)    - although I'm not sure to understand why travis complains about missing dependencies ... not familiar with this setup -

> Note the reason it doesn't work if you connect before creating the pubsub is that there are no connection notifications, so the peers are not discovered.

But even, after the creation of the pubsub, there will be no "connections notifications" as well, no ? Why having the pubsub before changes anything ?  The pubsub will likely send some notifications, but I don't understand that it works only when the notifications are sent in a non-opened connection yet ... ?

In any case, it's a bit a surprising behavior and can raise weird questions - do I need to subscribe to all topics before connecting ?, etc - : if the opportunity arises, it'd a good thing to document or change so it always work regardless of the order ;)

vyzo:
It's only the construction of the pubsub object that must precede the connections.
You can then connect before subscribing.

JackBekket:
It's acctually would be nice to see pubsub in examples

PR #146, opened 2019-01-11
godoc documentation is not markdown

vyzo:
This updates the doc.go stub to issue html instead of markdown; doh.


Kubuxu:
In future you can test it with:
```bash
go get -u golang.org/x/tools/cmd/godoc
godoc -http :6060
```

PR #149, opened 2019-01-15
Implement peer blacklist

vyzo:
Adds a peer blacklist, which is checked before doing any processing on an incoming message.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -98,6 +98,10 @@ type PubSub struct {
 	// eval thunk in event loop
 	eval chan func()
 
+	// peer blacklist
+	blacklist     map[peer.ID]struct{}
+	blacklistPeer chan peer.ID
+
 	peers        map[peer.ID]chan *RPC
 	seenMessages *timecache.TimeCache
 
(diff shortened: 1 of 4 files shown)


vyzo:
sorry for the force pushing mess, i was trying to get the test coverage right... it should be ready for review now.

vyzo:
there is a hole currently in that the incoming stream is not tracked; this allows a blacklisted peer to replay old valid messages from other peers.

vyzo:
actually it's ok, we do the blacklist check on the node that forwarded the message. but we should also check the message origin as well, as it might be blacklisted.

vyzo:
the issue is that the validator can't see the peer that forwarded the message, so it can't make blacklist decisions for that peer.

vyzo:
follow up in #150.

pubsub.go
```diff
@@ -179,6 +183,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		topics:           make(map[string]map[peer.ID]struct{}),
 		peers:            make(map[peer.ID]chan *RPC),
 		topicVals:        make(map[string]*topicVal),
+		blacklist:        make(map[peer.ID]struct{}),
```
Stebalien: May want to use an LRU. This could turn into a DoS vector.

Stebalien (commented):
We should probably use an LRU to prevent bad peers from forcing us to leak memory.

pubsub.go
vyzo: Good point. As we discussed with why, let's make it an interface the user can pass with a default implementation using a map.

vyzo:
Added a blacklist type, with map and lru cache backed implementations.
The blacklist can be configured with the `WithBlacklist` option.

Stebalien (approved):

PR #154, opened 2019-01-20
fix nonsensical check

### LINKED ISSUE #153: Duplicate blacklist check in pubsub
https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L294

```
			if p.blacklist.Contains(pid) {
				log.Warning("already have connection to peer: ", pid)
				continue
			}

			if p.blacklist.Contains(pid) {
				log.Warning("ignoring connection from blacklisted peer: ", pid)
				continue
			}
```

The second if should always be false, correct?

blame cc @vyzo

vyzo:
Fixes #153

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -291,7 +291,7 @@ func (p *PubSub) processLoop(ctx context.Context) {
 	for {
 		select {
 		case pid := <-p.newPeers:
-			if p.blacklist.Contains(pid) {
+			if _, ok := p.peers[pid]; ok {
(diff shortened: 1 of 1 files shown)


paralin:
I don't want to sound pedantic, but I'd also recommend adopting a more formal commit message style (see Angular 2's commit style or the Linux kernel commit style) just because it makes peer reviewing much easier.

vyzo:
well, that _was_ a brainfart, so i called it as such :p
but ok, i'll call it something else if it offends your sensibilities.

PR #156, opened 2019-01-21
move timecache check/update after validation

### LINKED ISSUE #155: The timecache should be checked/updated after validation
Currently, the timecache is checked/updated for seen messages prior to message validation.
This was done with the intention of avoiding duplicate message validation.
Unfortunately it introduces a censorship attack vector:
If an adversarial actor can predict a peer's sequence number (by observing a prior message), then it can attempt to send an invalid message with the predicted id. 
The message will fail validation, due to signature firstly; but it will also poison the timecache, resulting in dropping a sebsquent message from the peer with the predicted id.

cc @whyrusleeping @Stebalien

vyzo:
Fixes #155.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -625,12 +625,6 @@ func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
 		return
 	}
 
-	id := msgID(msg.Message)
-	if p.seenMessage(id) {
-		return
-	}
-	p.markSeen(id)
-
 	if len(vals) > 0 || msg.Signature != nil {
 		// validation is asynchronous and globally throttled with the throttleValidate semaphore.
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Looks correct as a direct fix. However, we should consider https://github.com/libp2p/go-libp2p-pubsub/issues/155#issuecomment-456216431 as validation is already a performance issue for us.

vyzo:
I will also add a check of the cache before validation, but without update.
This will eliminate the duplicate valid message validation performance issue.

Stebalien:
Well, not quite *eliminate*. If we get a burst of messages (as expected), we'll still see duplicate validations.

vyzo:
Indeed, but that's almost impossible to avoid. But at least its only for the window of active validation.

PR #162, opened 2019-02-12
gx publish 0.11.12

vyzo:


vyzo:
No update needed, there is no go-libp2p dependency

Issue #161, opened 2019-02-11
Message Propagation/Gossip Modularity


aschmahmann:
Currently there is a base PubSub with different PubSubRouters for floodsub and gossipsub. However, it would be useful in certain situations to have a gossip/routing layer that didn't require assumptions as strong as those made by PubSub.

For example, #42 could be easier to solve if a new protocol with different requirements could be layered on top of the basic routing layer without requiring PubSub itself to change. Similarly, it could help with not needing IPNS over pubsub to need to hit the DHT on the initial request (not sure the issue number, but I've seen this issue referenced a few times). Additionally, the peer-base team made created this protocol https://github.com/peer-base/peer-base/blob/master/docs/PROTOCOL.md for synchronizing CRDTs across a set of peers without using much of anything from pubsub (yes, part of this is that Javascript gossipsub isn't done yet, but even from Go their use case is still not covered without writing a whole new system)

Another use case is p2p synchronization of append-only DAGs ipld/replication#3 and https://github.com/aschmahmann/ipshare/blob/master/sync/MultiWriterIPNS.md. I'd like to take whatever synchronization scheme we come up with and follow a pattern similar to gossipsub so that, instead of gossipsub's A sends message to B who sends message to C, have A synchronizes with B who synchronizes with C.

Visually this might look like wanting to upgrade from `(A -> B, B -> C) to (A <-> B, B <-> C)`

I'm thinking my first stab at this might involve some copy + paste + edit work from pubsub + gossipsub, but before I dive deeper in than I already have I'd like some of your thoughts @vyzo @bigs

PR #171, opened 2019-04-08
Added subscription specific configurations to PubSub

aschmahmann:
There's still work to be done here, both in documentation and some design work. It'd be great to get some feedback on the direction this is going though.

-----

Overall the additions are:
 - Added the ability to have subscription based options
 - Added some configurable parameters to gossipsub
 - Added a gossipsub last-writer-wins-style persistence configuration


raulk:
Hey @aschmahmann, thanks for the contribution! 🎉  Could you link to an issue where the need and rationale behind this was discussed? Thanks!

gossipsub.go
```diff
@@ -514,13 +568,13 @@ func (gs *GossipSubRouter) emitGossip(topic string, peers map[peer.ID]struct{})
 	for _, p := range gpeers {
 		// skip mesh peers
 		_, ok := peers[p]
-		if !ok {
+		if !ok || ok {
```
aschmahmann: The original gossipsub implementation seems to look for GossipSubD (6) peers to re-emit gossip to and ignores members of the mesh (so sends 0-6 gossips). Is this intentional, why not have the GossipSubD peers all be not members of the mesh (so it's always 6 if 6 non-mesh peers are available)?

Additionally, in order to have a persistence option we need to have the ability to rebroadcast messages during a heartbeat. We can do this by:
1. Adding heartbeat functions into the configuration parameters
2. Adding a boolean persistence flag into the configuration parameters
3. Adding a particular rebroadcast function into the configuration parameters

I could be convinced any of these work, but given that currently the mesh parameters (GossipSubD, Dlo, Dhigh, GossipSubHeartbeatInterval, ...) are not configurable having a boolean seems not too bad. However, it's possible we may want the gossip parameters to be configurable as well.

@pgte @vyzo Any thoughts here?

gossipsub_test.go
```diff
@@ -297,14 +297,14 @@ func TestGossipsubFanoutExpiry(t *testing.T) {
 		}
 	}
 
-	if len(psubs[0].rt.(*GossipSubRouter).fanout) == 0 {
+	if len(psubs[0].rt.(*GossipConfigurableRouter).fanout) == 0 {
```
aschmahmann: Since NewGossipSub return a PubSub instance these type assertions were the only places where I ran into issues with renaming GossipSubRouter into GossipConfigurableRouter. However, if we think there are more places where this could be an issue we could utilize a modified GossipSubRouter that is based on a GossipConfigurableRouter.

gossipsubimpl.go
```diff
@@ -0,0 +1,23 @@
...
+	host "github.com/libp2p/go-libp2p-host"
+	protocol "github.com/libp2p/go-libp2p-protocol"
+)
+
+const (
+	GossipSubID = protocol.ID("/meshsub/1.0.0")
+)
+
+// NewGossipSub returns a new PubSub object using a GossipConfigurableRouter as the router and configured with classic GossipSub parameters.
+func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
```
aschmahmann: It feels like this should be in a separate file now that NewGossipSub is just one of multiple possible gossip based configurations. Up for any suggestions on where this code lives or what to name the file.

pubsub.go
```diff
@@ -139,6 +140,10 @@ type PubSubRouter interface {
 	Leave(topic string)
 }
 
+type PubSubRouterConfig interface {
+	JoinWithProtocol(topic string, proto protocol.ID) error
```
aschmahmann: This seemed like the easiest way to add in topic/subscription based options. However, it's a little bizarre and I'd be fine with either replacing the interface with a simple function (non-breaking change) or investigating an interface breaking change/addition to PubSubRouter.

pubsub.go
```diff
@@ -454,6 +475,12 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
 	// announce we want this topic
 	if len(subs) == 0 {
 		p.announce(sub.topic, true)
+		if p.rtConfig != nil && sub.topicProtocol != "" {
+			if err := p.rtConfig.JoinWithProtocol(sub.topic, sub.topicProtocol); err != nil {
```
aschmahmann: Wasn't really sure what to do here since I didn't want to throw a panic, but there's no way to return an error. The checks in Subscribe help make sure this doesn't happen though.

gossipsub.go
```diff
@@ -208,21 +230,29 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
 		tmap, ok := gs.p.topics[topic]
 		if !ok {
 			continue
 		}
 
-		// floodsub peers
-		for p := range tmap {
-			if gs.peers[p] == FloodSubID {
-				tosend[p] = struct{}{}
+		if withFloodSubPeers{
```
aschmahmann: The various configurations that I've implemented so far could likely all be done in floodsub, episub, etc.  Two caveats are 1) we'd need to allow the layer above the router implementations to have more access to the internals (e.g. which peers I'm syncing/sending messages to for a given topic, some control over limiting or manually triggering a broadcast) 2) The existing less performant implementations (e.g. floodsub) would do extremely poorly for some of these configuration options (e.g. persistence) anyway.

Additionally, in future implementations of say episub the configs used by gossipsub should be reusable.

gossipconfig.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+//HandleRPC, handleIHave, handleIWant, Publish, pushGossip
+
+type MessageCacher interface {
+	MessageCacheReader
+	Put(msg *pb.Message)
+}
+
+type ClassicGossipSubConfiguration struct {
```
aschmahmann: Up for a better name if you've got one

gossipsub.go
```diff
@@ -1,22 +1,16 @@
...
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 
-	host "github.com/libp2p/go-libp2p-host"
 	peer "github.com/libp2p/go-libp2p-peer"
 	protocol "github.com/libp2p/go-libp2p-protocol"
 )
 
-const (
-	GossipSubID = protocol.ID("/meshsub/1.0.0")
```
vyzo: don't remove this from here.

gossipsub.go
```diff
@@ -35,18 +29,31 @@ var (
...
-func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
-	rt := &GossipSubRouter{
-		peers:   make(map[peer.ID]protocol.ID),
-		mesh:    make(map[string]map[peer.ID]struct{}),
-		fanout:  make(map[string]map[peer.ID]struct{}),
-		lastpub: make(map[string]int64),
-		gossip:  make(map[peer.ID][]*pb.ControlIHave),
-		control: make(map[peer.ID]*pb.ControlMessage),
-		mcache:  NewMessageCache(GossipSubHistoryGossip, GossipSubHistoryLength),
+func NewGossipConfigurableRouter(configs ...GossipConfiguration) *GossipConfigurableRouter {
```
vyzo: don't remove the old constructor.

gossipsub.go
```diff
@@ -56,32 +63,44 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
 // to use for injecting our messages in the overlay with stable routes; this
 // is the fanout map. Fanout peer lists are expired if we don't publish any
 // messages to their topic for GossipSubFanoutTTL.
-type GossipSubRouter struct {
+type GossipConfigurableRouter struct {
```
vyzo: Don't rename the struct/make a new one, just keep and extend the existing one.

gossipsub.go
```diff
@@ -208,21 +230,29 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
 	}
 }
 
-func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
-	gs.mcache.Put(msg)
-
-	tosend := make(map[peer.ID]struct{})
+func (gs *GossipConfigurableRouter) Publish(from peer.ID, msg *pb.Message) {
 	for _, topic := range msg.GetTopicIDs() {
+		config, ok := gs.topicConfigs[topic]
```
vyzo: that's just backwards; can't have a `config` object handling hte most important function of the router!

gossipsubimpl.go
vyzo: this file is redundant.

subscription.go
```diff
@@ -2,13 +2,15 @@ package pubsub
...
+	protocol "github.com/libp2p/go-libp2p-protocol"
 )
 
 type Subscription struct {
-	topic    string
-	ch       chan *Message
-	cancelCh chan<- *Subscription
-	err      error
+	topic         string
+	topicProtocol protocol.ID
```
vyzo, line 10: I don't think this belongs here.

vyzo (commented):
Some very passing comments from first reading.
What's the rationale for these changes?

aschmahmann (commented):
The main rationale for having a persistent pubsub is at #42. I also briefly give some reasons for needing this at #161.

gossipsub_test.go
vyzo: Please don't change the name of the router. It's the gossipsub router, and it has become configurable.

pubsub.go
vyzo: We are fine we interface-breaking changes here, routers are only used internally for now.

pubsub.go
vyzo: this looks very redundant.

vyzo:
Just a note: you are changing very carefully thought-out code, don't go doing stuff willy-nilly!

gossipsub.go
aschmahmann: Isn't that the whole premise of every inversion of control system? We could create new routers and pubsub instances for every configuration option instead of using one pubsub instance.

However, that means
1. having multiple heartbeat tickers running at the same time
2. not reusing any of the gossipsub structures across implementations 
3. ending up with weird multiple tenancy issues unless we change PubSub more. For example, if right now you create a FloodSub and GossipSub instance on the same node you end up with overly excessive message passing *because* GossipSub is backwards compatible with FloodSub and will handle both messages. Either we need one router to handle multiple configurations or we need to make PubSub able to handle backwards compatible protocols, this way involves less protocol work.

gossipsub.go
aschmahmann: This was moved to a new file because I wanted to separate a particular gossipsub implementation from configurable gossipsub. However, I don't feel strongly as to where it lives. If you really think they should all be in the same file that's fine.

gossipsub.go
aschmahmann: see above, it was moved not *re*moved

gossipsub.go
aschmahmann: The new struct breaks the contract of the old struct, but reuses almost all of its code. I don't really care what the struct is named, but anyone who was using `GossipSubRouter` instead of `NewGossipSub` is going to be quite surprised.

gossipsub.go
vyzo: yes, it should live here.

gossipsub.go
vyzo: keep it here and make a new constructor `NewGossipSubRouterWithConfig`

gossipsub.go
vyzo: what? These are all internal in the library, I don't care if we break the old struct.

subscription.go
aschmahmann: I disagree, there is in PubSub a declaration `type SubOpt func(sub *Subscription) error` that is currently almost unusable because `Subscription` has no fields to work with.

I assume that pattern taken with both `Option` (for PubSub) and `SubOpt` (for Subscriptions) is designed to allow adding fields to the struct, and functionality as option wrappers, without breaking the struct contract/interface. Is that incorrect, if so how is `SubOpt` supposed to be used and where would you put the relevant topic/Subscription information?

pubsub.go
aschmahmann: The check looks redundant, or the whole `JoinWithProtocol` thing looks redundant?

Both appear redundant, however `JoinWithProtocol` is necessary if we don't want to break the interface for PubSubRouter and still support topic/Subscription options.

The router configuration check is necessary since other routers (e.g. FloodSub) may not have a router configuration. The subtopic check is necessary since I think it's reasonable that if your configurable router only has one option (e.g. standard gossipsub created by `NewGossipSub`) that you shouldn't need to subscribe with a topic protocol since it is implied. This has the added benefit of not breaking anything for people who have been using PubSub via NewGossipSub, NewFloodSub,.....

gossipsub.go
vyzo: it's intentional, you can't have them be members of the mesh. It's the one thing that makes the whole protocol work even if the mesh is not fully connected.

pubsub.go
vyzo: We don't care if we break the router interface.
There are no users outside the library, and it's considered unstable anyway.

aschmahmann:
> Just a note: you are changing very carefully thought-out code, don't go doing stuff willy-nilly!

Nothing here has been done "willy-nilly". Everything I've done so far has been targeted at making sure "user space" around PubSub and gossipsub has been minimally affected (which is true for anyone using the GossipSub, FloodSub, ... constructors), while at the same time allowing for changes that are needed by consumers of PubSub (e.g. IPNS).

I see from some of your comments that we're actually ok with some breaking changes within the PubSub system which should certainly help clean up the code.

I respect that a lot of thought went into the PubSub system and is part of why I'd like to enable users to tinker with it from user space as much as possible so they don't end up trying to copy-paste or fork pubsub. I think it'd be great to have some document discussing how we'd like to enable people to extend PubSub and also consider whether the configuration options I'm putting in are sufficient or just sufficient for the cases that I'm currently considering.

pubsub.go
aschmahmann: great, so it should be easy enough to remove this and combine it into the existing interface (or just replace/update the existing Join function). This does mean making small changes to FloodSub and RandomSub though.

gossipsub_test.go
aschmahmann: no problem

gossipsub.go
aschmahmann: as mentioned below I'm fine changing it back since we're fine breaking the struct contract.

gossipsub.go
aschmahmann: my question is why isn't the function passed into getPeers something like "peer is not in mesh" instead of "return true"?

"return true" means even if we have 1012 peers subscribed to `topic` (of which 12 are in our mesh) we will emit requests to <=6 peers.

"peer is not in mesh" would return 6 peers randomly selected from the 1000 subscribed peers not in the mesh.

gossipsub.go
aschmahmann: sure

gossipsub.go
aschmahmann: ok

gossipsubimpl.go
aschmahmann: yep, since we're moving everything back to the original file.

gossipsub.go
aschmahmann: Given that we're ok breaking the router implementations/interfaces a bit do you think I should try and setup some of the configurations (e.g. persistence) to work across all router implementations or is leaving the configurations as something to be plugged into future routers (e.g. episub) sufficient?

pubsub.go
aschmahmann: sounds good. Take a look at my comment below (on withFloodSubPeers) which should help me figure out how to rewrite the interfaces for the next iteration.

gossipsub.go
vyzo: it's a filter, we select all peers and then prune the ones already in the mesh.

gossipsub.go
vyzo: that's fine, although I don't know how useful it will be for floodsub.

gossipsub.go
vyzo: so wtf is up with the `ok || !ok`? That's vacuously true and it breaks the invariant of the if.

aschmahmann: But getPeers has a maximum length (which is currently 6) so first you filter "all" into 6 and then you filter those down based on the mesh. Why not filter down to 6 based on the mesh?

yes, the `ok || !ok` is obviously superfluous and going to get removed, but I want to know how to remove it before I do so. I gave three options above for how to remove it and put this comment here to flag it. See the details above.

raulk:
@aschmahmann could you open an RFC discussion in libp2p/specs? Please outline context, motivation, rationale, goals, etc. Then link to this PR as a sample implementation. Inferring all that from various issues is less than ideal and is subject to bias and interpretation. Also, not all libp2p implementers watch all go repos, so we elevate protocol discussions to the specs repo. libp2p is a specs-driven project so any changes introduced here would have to be preambled by changes in specs.

aschmahmann:
@raulk There are 3 potential specs I could add, and would like some advice on which ones to go with. There's: 1) PubSub patch 2) GossipSub patch 3) Something about Last-Writer-Wins persistence

1) There doesn't appear to be anywhere in the PubSub spec to add anything interesting from my work, although I'm up for any suggestions. I haven't touched any wire protocols and there's no mention in the spec on the `Join` function or on any of the options in the go implementation (e.g. Blacklist).
2) I can patch the GossipSub spec to allow for Joining a topic with a particular protocol ID and forwarding certain functions to the particular configurations. However, I need confirmation from @vyzo that the inversion of control mechanism I'd like to use is ok. While IoC is an implementation detail it affects whether the change occurs at the PubSub or GossipSub layers. Given that there's nowhere in the PubSub spec for this to go it seems to imply that either the PubSub spec is missing info or the GossipSub spec has too much implementation specific info.
3) Happy to write this up, although I need to do a little thinking first because since @vyzo suggested it was possible for me to break the PubSubRouter interfaces it might be useful to implement LWW at the PubSub level instead of the GossipSub level. Note that if it's added at the PubSub level it's unclear where to put the spec changes, as described above.

raulk:
@aschmahmann I would start with an issue in libp2p/specs theorising context (what is the current situation), description of change (what you want to change), motivation/rationale (why you want to change it), proposed design (how you want to change it), proposed next steps.

Then the community can gauge if this is something of interest across implementations, and can advice how to best integrate it.

aschmahmann:
@raulk @vyzo, I published a high level overview of why and what I want to change at the issue above (#175). As you'll notice there's limited spec change proposals (although if my LWW implementation is going to be part of libp2p it may want to get its own spec proposal at some point).

aschmahmann:
@vyzo while PubSub's `Subscribe` function takes in Options (and as a result can be setup for configuration) the `Publish` function does not. I was thinking of adding a `PublishWithOptions` function to accommodate this without breaking anyone who is using a default `Publish` currently. What do you think?

vyzo:
what options do you want to pass to `Publish`?

aschmahmann:
For now likely just WithProtocol(proto protocol.ID) as I am with Subscribe

vyzo:
that won't work, the protocol ID is a property of the stream.

aschmahmann:
Yes, it is a property of the stream. In Master currently the way that you would setup a routing system with a different configuration (e.g. adding persistence) is by creating a new PubSubRouter with a new protocol ID.

The implementation here is similar except that instead of having one PubSub instance per router I am effectively putting multiple routers into a single PubSub instance and associating each topic with a particular router. If you look at what I'm doing with `Subscribe` and `Join` it's effectively this. However, because you can `Publish` to a topic without first subscribing I should be able to set the router that a topic is using from the `Publish` (or `PublishWithOptions`) function as well.

gossipsub.go
```diff
@@ -564,17 +579,20 @@ func (gs *GossipConfigurableRouter) emitGossip(topic string, peers map[peer.ID]s
...
 	}
 
-	gpeers := gs.getPeers(topic, GossipSubD, func(peer.ID) bool { return true })
-	for _, p := range gpeers {
-		// skip mesh peers
-		_, ok := peers[p]
-		if !ok || ok {
-			gs.pushGossip(p, &pb.ControlIHave{TopicID: &topic, MessageIDs: mids})
-		}
+	emitPeers := config.GetEmitPeers(gs.wrapGetPeers(topic), peers)
```
aschmahmann: @vyzo adding an `EmittingStrategy` to the configuration was one way to allow developers to choose which peers messages will get rebroadcasted to.

vyzo:
> I am effectively putting multiple routers into a single PubSub instance and associating each topic with a particular router. 

You are killing gossip piggybacking with this most likely.

vyzo:
Also, can we call `Config` something else? It's not an apt name when it encompasses all the fundamental actions of the protocol. Perhaps call it `Strategy`?

aschmahmann:
> You are killing gossip piggybacking with this most likely.

I'm actually think I'm doing the opposite. If we had to make a new PubSub instance for every configuration then we'd be stuck with multiple heartbeats each sending out there own piggybacked data which means that if A and B were connected by two routers then we'd end up with two messages instead of one message with both the data from A and B piggybacked onto it.

aschmahmann:
> Also, can we call `Config` something else? It's not an apt name when it encompasses all the fundamental actions of the protocol. Perhaps call it `Strategy`?

I don't think it encompasses all the fundamental actions of the protocol (I'd argue that graft + prune is the main component of gossipsub). However, as I don't have strong feelings about the naming here, I'm fine with `Strategy` or something more specific like `XStrategy` where X is a little more informative (especially since I've introduced an `EmitterStrategy` already).

gossipsub.go
```diff
@@ -209,20 +250,28 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+		}
+	}
+}
+
+func (gs *GossipSubRouter) AddGossipPeers(tosend map[peer.ID]struct{}, topics []string, withFloodSubPeers bool) {
+	for _, topic := range topics {
 		// any peers in the topic?
 		tmap, ok := gs.p.topics[topic]
 		if !ok {
 			continue
```
aschmahmann, line 279: @vyzo is it intentional that during a `Publish` we do not initialize `gs.fanout[topic]` if there are no peers in the topic (and we're not subscribed/there's no `gs.mesh[topic]`)? By not doing this we ensure that we will not rebroadcast (via emitGossip) our published message even once we come across new peers. This is less relevant in the current implementation since the republishing window is relatively short (5 heartbeat intervals which is 500ms), however it might still be worth correcting this even for our existing implementation.

vyzo: They won't be added to fanout unless there is a new message published.
Also, the heartbeat is 1 _second_, so 5 intervals are 5s.

aschmahmann: My bad on the interval timing, was looking at heartbeat initial delay by accident. Ok, so the following sequence of events in the current implementation is intentional?

**Scenario 1:**
1. A publishes M to GossipSub topic T, but has no peers
2. B subscribes to T
3. B connects to A
4. Nothing happens, B never receives M (even if within the cache history period)

**Scenario 2:**
1. A publishes M to GossipSub topic T, but has no peers
2. B1...B20 subscribes to T (important that it's more than `GossipSubD=6`, as mentioned [above](https://github.com/libp2p/go-libp2p-pubsub/pull/171#discussion_r273141427))
3. B1...B20 connects to A
4. Some of the B's receive M and some do not

**Scenario 3:**
1. A publishes M to GossipSub topic T, but has no peers
2. Wait 6 seconds
2. B1...B20 subscribes to T
3. B1...B20 connects to A
4. Nothing happens, none of the B's receives M

vyzo: Yes, that's fine.

aschmahmann: That psudo-use of a history buffer seems pretty suspect to me, but that's fine I can preserve that behavior and just modify the persistence version.

To clarify, I don't really see what's gained by having B not receive M in scenario 1.

vyzo: It's not designed to be a reliability mechanism; it's designed to help the network connect when the mesh is not fully connected for one reason or the other.

vyzo:
Closing as this is stale.

bitcard:
mark

PR #174, opened 2019-04-11
[WIP] Feat/metrics

bigs:


floodsub.go
```diff
@@ -8,6 +8,7 @@ import (
 	host "github.com/libp2p/go-libp2p-host"
 	peer "github.com/libp2p/go-libp2p-peer"
 	protocol "github.com/libp2p/go-libp2p-protocol"
+	"github.com/libp2p/go-libp2p-pubsub/metrics"
```
vyzo: named import please.

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
+package metrics
+
+import (
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+)
+
+var (
+	MOutgoingMsgs = stats.Int64("pubsub/outgoing_msg_size", "Outgoing message size", "By")
+	MIncomingMsgs = stats.Int64("pubsub/incoming_msg_size", "Incoming message size", "By")
```
vyzo: it's not a size!

pubsub.go
```diff
@@ -8,15 +8,16 @@ import (
...
 
-	pb "github.com/libp2p/go-libp2p-pubsub/pb"
-
 	logging "github.com/ipfs/go-log"
 	crypto "github.com/libp2p/go-libp2p-crypto"
 	host "github.com/libp2p/go-libp2p-host"
 	inet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
 	protocol "github.com/libp2p/go-libp2p-protocol"
+	"github.com/libp2p/go-libp2p-pubsub/metrics"
```
vyzo: named import please

vyzo (commented):
you need to instrument gossipsub as well.

vyzo:
We also need to track the validator pipeline. Can you add metrics for validation throttling and failures?
The first one will tell us if we have a message rate problem, and the second will tell us if there are misbehaving nodes and possibly an attack.

comm.go
```diff
@@ -44,6 +46,9 @@ func (p *PubSub) handleNewStream(s inet.Stream) {
 			return
 		}
 
+		measure := metrics.MIncomingMsgs.M((int64)(rpc.Size()))
+		go stats.Record(context.Background(), measure)
```
anacrolix: I would recommend against doing this with the go keyword. It removes profiling and tracking context from the runtime, and if you do get performance issues in the Record, you have no backpressure on the number of goroutines. There's also the idea of sequencing. At this point in the code, you want that measure updated, not at some undetermined point in the future. It might be a minor point here in the code, but it sets a precedent that's difficult to back out of.

floodsub.go
anacrolix: I would recommend against naming it. The conventions in Go are clear, we're just being a nuisance by deviating.

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
+package metrics
+
+import (
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+)
+
+var (
+	MOutgoingMsgs = stats.Int64("pubsub/outgoing_msg_size", "Outgoing message size", "By")
```
anacrolix: There are constants for the dimension (last argument), see https://github.com/libp2p/go-libp2p-kad-dht/pull/297/files#diff-e8d467baf1a83863d66232dc09f788e4 for an example.

Also I would recommend removing the prefixing `M`, and not naming your views. See also https://github.com/libp2p/go-libp2p-kad-dht/pull/297/files#diff-e8d467baf1a83863d66232dc09f788e4 for an example. I have some good reasons if you'd like to hear them.

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+
+	PeersGaugeView = &view.View{
+		Name:        "pubsub/peers",
+		Description: "Pubsub peers the host is connected to",
+		Measure:     MPeers,
+		Aggregation: view.LastValue(),
+	}
+)
+
+func Register() error {
```
anacrolix: I'd just expose all the views together, rendering this unnecessary. See the example and other comments.

pubsub.go
```diff
@@ -301,6 +302,9 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				continue
 			}
 
+			measure := metrics.MPeers.M((int64)(len(p.peers)))
+			go stats.Record(ctx, measure)
```
anacrolix: Recommend not with go keyword as above.

anacrolix (changes requested):
I think it's worthwhile making some changes to mirror https://github.com/libp2p/go-libp2p-kad-dht/pull/297/files#diff-e8d467baf1a83863d66232dc09f788e4, see the other comments. We're setting precedents here that we don't want to get wrong.

anacrolix:
Also there are race conditions and test timeouts in the CI :(. Nevermind, looks like those are well known.

floodsub.go
vyzo: I must insist. We use them everywhere in the code and I would like to be consistent.
Also, idiomatic or not, import naming is part of the language and it makes searching for symbols much easier.

metrics/metrics.go
lanzafame: @anacrolix 
> I have some good reasons if you'd like to hear them.

Yes.

floodsub.go
lanzafame: With @vyzo  on this one as once we register metrics from another `metrics` package, i.e. dht, we are going to need to name it anyway.

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+	return view.Register(
```
lanzafame: Libraries shouldn't register views. That is for the application to do.

vyzo:
Also, we need instrumentation for _dropped_ messages, not just _sent_ messages. If we are dropping messages, we'd like to know.

metrics/metrics.go
bigs: whooooooooops

bigs:
> We also need to track the validator pipeline. Can you add metrics for validation throttling and failures?
> The first one will tell us if we have a message rate problem, and the second will tell us if there are misbehaving nodes and possibly an attack.

yep sounds good!

> Also, we need instrumentation for _dropped_ messages, not just _sent_ messages. If we are dropping messages, we'd like to know.

k!

metrics/metrics.go
bigs: so the thought is our libraries can expose functions like this to be called from the application, registering all the views in the module. i don't think it's unreasonable, but i'm definitely open to an alternative. this should never be called within the package

raulk: Tying this in with the upcoming host/service refactor, it could make sense to have a `Measurable` (or like) interface that the omnihost can cast services to and call to register default views, iff a configuration option is enabled (and it’ll probably be enabled by default).

metrics/metrics.go
bigs: oh, actually the measure *is* a size, but i can aggregate it by count as well

metrics/metrics.go
bigs: as in a slice of views?

metrics/metrics.go
anacrolix: I think you can do this with an extra view with view.Count, on the same measure. Not 100% sure.

metrics/metrics.go
lanzafame: An example from the OC GRPC plugin: https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/server_metrics.go#L91-L97

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
+package metrics
+
+import (
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+)
+
+var (
+	MOutgoingMsgs = stats.Int64("pubsub/outgoing_msg_size", "Outgoing message size", stats.UnitBytes)
```
lanzafame, line 9: Since the measure is already package namespaced, it doesn't add much value to prefix with `M`.

metrics/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+	MIncomingMsgs = stats.Int64("pubsub/incoming_msg_size", "Incoming message size", stats.UnitBytes)
+	MTopics       = stats.Int64("pubsub/topics", "Number of topics currently subscribed", stats.UnitDimensionless)
+	MPeers        = stats.Int64("pubsub/peers", "Number of pubsub peers", stats.UnitDimensionless)
+
+	OutgoingMsgCountView = &view.View{
```
lanzafame, line 14: I am not sure exactly what would make sense, but I would like to at least have the option to differentiate these views by `peer_id` see [here](https://github.com/libp2p/go-libp2p-kad-dht/pull/317/files#diff-a4f6a21f1d4b814bb22ae5c16546906dR18) and [here](https://github.com/libp2p/go-libp2p-kad-dht/pull/317/files#diff-a4f6a21f1d4b814bb22ae5c16546906dR48).

lanzafame (changes requested):

comm.go
```diff
@@ -44,6 +46,9 @@ func (p *PubSub) handleNewStream(s inet.Stream) {
 			return
 		}
 
+		measure := metrics.MIncomingMsgs.M((int64)(rpc.Size()))
+		stats.Record(context.Background(), measure)
```
vyzo, line 50: Please abstract both these calls behind a single `metrics` interface. I don't want to see this dance sprinkled everywhere in the code and it shouldn't detail on opencensus details whatsoever.

comm.go
```diff
@@ -5,7 +5,9 @@ import (
 	"context"
 	"io"
 
+	metrics "github.com/libp2p/go-libp2p-pubsub/metrics"
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	stats "go.opencensus.io/stats"
```
vyzo, line 10: please hide this behind a `metrics` interface, I don't want to depend on opencensus directly anywhere in the code except the `metrics` module.

floodsub.go
```diff
@@ -4,10 +4,12 @@ import (
 	"context"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	stats "go.opencensus.io/stats"
```
vyzo, line 7: here too.

gossipsub.go
```diff
@@ -6,11 +6,13 @@ import (
 	"math/rand"
 	"time"
 
+	metrics "github.com/libp2p/go-libp2p-pubsub/metrics"
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 
 	host "github.com/libp2p/go-libp2p-host"
 	peer "github.com/libp2p/go-libp2p-peer"
 	protocol "github.com/libp2p/go-libp2p-protocol"
+	stats "go.opencensus.io/stats"
```
vyzo, line 15: and here.

vyzo (changes requested):
Please abstract opencensus completely inside the `metrics` module.
The pubsub/floodsub/gossipsub code should have no dependency whatsoever on opencensus and the metrics record should be a single call that hides details.

comm.go
vyzo: Also, do we care about the size of messages? We care about the _number_ of messages a lot more.

metrics/metrics.go
vyzo: We definitely want count, we don't care about sizing as much but we do care about _number of messages_.

metrics/metrics.go
anacrolix: Yeah. I think in the future, it's going to become tedious to gather up these views from everywhere, and we'll come full circle to registering them all into a common registry anyway. But at least in the interest of making it easier now, I think a slice of views is the most convenient.

comm.go
anacrolix: @vyzo I wonder if it's worth abstracting what's already supposed to be an abstraction. It's already painful enough trying to use Prometheus through the keyhole that is OC. If OC is not a good abstraction, why are we using it?

anacrolix (approved):

comm.go
vyzo: It doesn't matter, it's just metrics -- but I don't want to pollute the code with details of the metrics system.

bigs: going to update what we're actually measuring here, so don't worry about that. imo, wrapping OC would be a lot of duplicated effort leading us straight back to go-metrics-interface, at which point all we've really done is create a new dependency. this means every measure would have to be accompanied by a bespoke function or we make a new interface

vyzo: Just turn the two functions into a single metrics call; I don't mind having duplication in the `metrics` package, I just don't want to see the god damn details of the metrics subsystem littered in the code!

vyzo: sorry about the tone, I didn't mean to sound so harsh. But I do feel strongly about it, the metrics details are just that: implementation details and I shouldn't have to write two lines of that just to increment a counter.

bigs: no worries and yeah, it does feel a bit ugly. i'll put up some proposals

raulk: OC *is* the abstraction layer and we've picked it because it's an open standard. Introducing an abstraction layer over an abstraction layer not only has runtime cost, but it's redundant.

raulk: Let's be [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) and encapsulate common OC rituals behind a package (github.com/libp2p/go-libp2p-core/metrics).

Let's not create _an abstraction_ but a helper toolbox.

raulk:
Discussion about (not) abstracting the abstraction here: https://github.com/libp2p/go-libp2p-pubsub/pull/174#discussion_r275699146

comm.go
vyzo: I would argue that it's a very poor interface. I don't want to care about the details of metrics in the code, so hide it behind the dedicated module that handles them. Whether that's `core/metrics` or the `metrics` package which we have in here, it's immaterial.

vyzo: Also, the runtime cost here is negligible. I am not asking for a new abstraction layer, I want a better interface that doesn't pollute the code with _crap_ :p

raulk: Thoughts:

1. We don't want to _pollute_ the code with **repetitive** ceremonies. But if the OC call is a one-liner, we are not going to put it behind an indirection. There's no reason to hide OC -- but there is a reason to be DRY.

2. Stats bookkeeping will be done _everywhere, all the time_. The runtime cost will not be negligible and we should check all calls the library we cook up are inlined (`-gcflags="-m"`) and that we're _not allocating_ beyond what OC needs.

vyzo: Note that I am not asking for a generic abstraction here, I am asking to make named methods that hide the details of the OC ugliness.
Something like `metrics.inrementMessageCounter()` that hides the two lines is good enough for me.

vyzo: But if we have a nicer abstraction, that would obviously be acceptable.

Bottom line: I don't want to have to care about `M` or contexts here.

vyzo:
we have tracing which can gather all the metrics we need; no need for this.

Issue #173, opened 2019-04-11
Validators are called multiple times for the same message


iulianpascalau:
Hello

There seem to be an issue in which validators are called multiple times for the same message (same peer, same seqno). After placing some trace print lines in pubsub.go here is what I have found out:
suppose we have a message m with an id (id = peerID+seqNo)
1. first time message m enters the pushmsg func, it queries seenMessage func which will return false so it can continue. It launches the validator and exits through the return on line 648;
2. it might happen that the same message comes again, enters the same pushmsg func, queries again seenMessage func which will return again false, calls the validator and return on line 648;
3. after the first validator launched finish its job, it will write on chan sendMsg the send request which will trigger the call of publishMessage from line 392 which will mark the message m as seen;
4. if the same message m comes again, it will not be revalidated as the seenMessages is updated.

The problem was identified only recently, the commit that changed the calling of markSeen method order was 67eece25262fcbc36c93daf67da5de2daa01de49 from Jan 21, 2019. Before this change, validator func was called only once for a message.

vyzo:
The change you mention was made to close a hole that would lead to a potential censorship/cache poisoning  attack.
The problem was that any peer that could predict the message id of another peer (an observation of a prior message is sufficient), could launch an attack where they generated invalid messages with the next message ids of the targeted peer.
If those message were marked prior to validation (which was the case before the patch), then the cache would be poisoned and valid messages from the targeted peer would be rejected.
Hence we decided to mark messages after validation.
Note that we were fully aware that this would lead to a single message potentially being validated multiple times, but decided to take that cost.

iulianpascalau:
Ok, got it. Maybe, if signing on messages becomes mandatory, message.Signature might become the msgID and markSeen could be called in pushMsg func.

vyzo:
That's not a bad idea.

vyzo:
#176 fixes the issue -- it now ensures that user validators are invoked at most once for each message by marking them as seen as soon as the signature has been verified.

iulianpascalau:
Great news! Thanks.

Issue #172, opened 2019-04-11
bandwidth limiting


yutongp:
Hi libp2p engs,

In our project, we use libp2p-pubsub for our p2p layer. Today we did a DoS test, we sent 3000 TPS requests into the network and the all network hang(all 30 nodes stuck).
We investigated our program profile a bit and found out 45% of cpu timing is working on pubsub message verification and that basically consumed all our 4 cores CPU resource for each node in our network. Btw, we launched the attack in one single machine(4 cores also)
<img width="260" alt="Screen Shot 2019-04-10 at 9 24 46 PM" src="https://user-images.githubusercontent.com/915849/55930727-9062c880-5bd7-11e9-8ddc-35e930cf1248.png">

I posted a similar issue in go-libp2p-connmgr: https://github.com/libp2p/go-libp2p-connmgr/issues/37.

Since pubsub already has a `Blacklist` feature, and the issue is caused by msg verfication in pubsub, should pubsub provide ability to block large bandwidth node to flood the network?

Thanks.

vyzo:
I am a little reluctant to automatically blacklist high bandwidth nodes in the library, this should be done by user decree. 

You can also try to tune down the validation throttle, which will start dropping messages that can't be validated because the queue is full.

vyzo:
But in general I am in agreement that we should detect the situation and at least notify the user, which can then take action and potentially blacklist peers.

raulk:
I believe you’re seeing two different ~interrelated,~ issues that interplay.
1. Pubsub validators are assumed to be lightweight.
2. We don’t do traffic shaping, yet.

For 1, there’s #169 (if you really need to do heavy validation). For 2, we are evolving the connection manager and eventually it’ll become a traffic shaper.

Would you or your team be willing to help us out with any of those? :-)

yutongp:
> You can also try to tune down the validation throttle, which will start dropping messages that can't be validated because the queue is full.

We investigated the validation throttle configuration. It doesn't help in our experiment. We used 500 and 10, in both cases our CPUs climbed to 100% quickly under our load test. I think the reason is blew code always run in "full speed"
```go
		select {
		case p.validateThrottle <- struct{}{}:
			go func() {
				p.validate(vals, src, msg)
				<-p.validateThrottle
			}()
		default:
			log.Warningf("message validation throttled; dropping message from %s", src)
		}
```
[...]

PR #176, opened 2019-04-25
rework validator pipeline

### LINKED ISSUE #103: Optimize Validator Pipeline
Our validation pipeline is asynchronous: it spawns a goroutine for each incoming message, which may then get throttled based on topic and global limit throttles.
This creates a performance problem for message signature validation (#97) as we now have to enter the validation pipeline for every message when using signing (planned to become the default in ipfs and filecoin).

Proposed action:
- pre-spawn NUMCPU goroutines to handle the validation front-end.
- all signatures should be validated in the validation front-end goroutines
- create a distinction between synchronous and asynchronous validators (which may block and be otherwise slow due to network effects), such that:
  - synchronous validators are applied in the validation front-end
  - asynchronous validators are handled in a freshly spawned goroutine, limited by the throttles.

### LINKED ISSUE #169: Offloading messages for async validation
From @arnetheduck (Nimbus, ETH 2.0 client):

> @raulk we discussed topic validation in libp2p as a way to prevent bad information from spreading across the gossipsub network, though from what I can tell, the block propagation filtering method in libp2p that you pointed me to is synchronous (https://github.com/libp2p/go-libp2p-pubsub/blob/bfd65a2f6b810c5b4ad2cfe6bb9cc792fd7a0171/floodsub_test.go#L360). this might not sit well with block validation, where we might want to prevent gossiping a block until we've verified it based on data that we receive later.  how would you recommend we approach this? 
> 
> here's the scenario in detail:
> * over gossip, we receive a block whose parent we're missing
> * worst case, this means we cannot yet tell if it's a good / useful block or not
> * we don't want the block to be gossiped further until we've recovered its parent to ensure that it's sane. once we do know it's sane, we want to pass it on.

To summarise:

1. Validation can be costly or not feasible in some scenarios to perform sync.
2. Is it feasible to consume the message, do validation offline, then republish it? How does that affect message caches, duplicate detection across the network (e.g. if we send the message to peers who had already seen it -- and possibly even propagated it if they had more complete data than us), do we generate a new message ID?
3. What are the differences on the wire between publishing a message afresh, and spreading a gossiped message?

In a nutshell: is it possible to offload a message from the pubsub router for async validation, then resume its gossiping conditionally?

vyzo:
- [X] Prespawn goroutines for signature validation, avoiding new goroutines for each message.
- [X] Ensure user validators are invoked at most once for each message by marking them as seen as soon as the signature verification completes.
- [X] Support asynchronous validation that can take a long time to complete ~and needs to be offloaded~.
- [X] Support inline (synchronous) validation for simple validators.

See #172 #173 #169
Closes #103

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math/rand"
+	"runtime"
 	"sync/atomic"
 	"time"
 
@@ -91,6 +92,9 @@ type PubSub struct {
 
 	// topicVals tracks per topic validators
 	topicVals map[string]*topicVal
+
+	// validateQ is the front-end to the validation pipeline
+	validateQ chan *validateReq
 
 	// validateThrottle limits the number of active validation goroutines
 	validateThrottle chan struct{}
@@ -184,6 +188,7 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		topics:           make(map[string]map[peer.ID]struct{}),
 		peers:            make(map[peer.ID]chan *RPC),
 		topicVals:        make(map[string]*topicVal),
+		validateQ:        make(chan *validateReq, 32),
 		blacklist:        NewMapBlacklist(),
 		blacklistPeer:    make(chan peer.ID),
 		seenMessages:     timecache.NewTimeCache(TimeCacheDuration),
@@ -209,6 +214,11 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	h.Network().Notify((*PubSubNotif)(ps))
 
 	go ps.processLoop(ctx)
+
+	numcpu := runtime.NumCPU()
+	for i := 0; i < numcpu; i++ {
+		go ps.validateWorker()
+	}
 
 	return ps, nil
 }
@@ -631,16 +641,8 @@ func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
 	}
 
 	if len(vals) > 0 || msg.Signature != nil {
-		// validation is asynchronous and globally throttled with the throttleValidate semaphore.
-		// the purpose of the global throttle is to bound the goncurrency possible from incoming
-		// network traffic; each validator also has an individual throttle to preclude
-		// slow (or faulty) validators from starving other topics; see validate below.
 		select {
-		case p.validateThrottle <- struct{}{}:
-			go func() {
-				p.validate(vals, src, msg)
-				<-p.validateThrottle
-			}()
+		case p.validateQ <- &validateReq{vals, src, msg}:
 		default:
 			log.Warningf("message validation throttled; dropping message from %s", src)
 		}
@@ -650,7 +652,20 @@ func (p *PubSub) pushMsg(vals []*topicVal, src peer.ID, msg *Message) {
 	p.publishMessage(src, msg.Message)
 }
 
+func (p *PubSub) validateWorker() {
+	for {
+		select {
+		case req := <-p.validateQ:
+			p.validate(req.vals, req.src, req.msg)
+		case <-p.ctx.Done():
+			return
+		}
+	}
+}
+
 // validate performs validation and only sends the message if all validators succeed
+// signature validation is performed synchronously, while user validators are invoked
+// asynchronously, throttled by the global validation throttle.
 func (p *PubSub) validate(vals []*topicVal, src peer.ID, msg *Message) {
 	if msg.Signature != nil {
 		if !p.validateSignature(msg) {
@@ -660,13 +675,19 @@ func (p *PubSub) validate(vals []*topicVal, src peer.ID, msg *Message) {
 	}
 
 	if len(vals) > 0 {
-		if !p.validateTopic(vals, src, msg) {
-			log.Warningf("message validation failed; dropping message from %s", src)
-			return
+		select {
+		case p.validateThrottle <- struct{}{}:
+			go func() {
+				p.doValidateTopic(vals, src, msg)
+				<-p.validateThrottle
+			}()
+		default:
+			log.Warningf("message validation throttled; dropping message from %s", src)
 		}
+		return
 	}
 
-	// all validators were successful, send the message
+	// no user validators and the signature was valid, send the message
 	p.sendMsg <- &sendReq{
 		from: src,
 		msg:  msg,
@@ -681,6 +702,18 @@ func (p *PubSub) validateSignature(msg *Message) bool {
 	}
 
 	return true
+}
+
+func (p *PubSub) doValidateTopic(vals []*topicVal, src peer.ID, msg *Message) {
+	if !p.validateTopic(vals, src, msg) {
+		log.Warningf("message validation failed; dropping message from %s", src)
+		return
+	}
+
+	p.sendMsg <- &sendReq{
+		from: src,
+		msg:  msg,
+	}
 }
 
 func (p *PubSub) validateTopic(vals []*topicVal, src peer.ID, msg *Message) bool {
@@ -881,6 +914,13 @@ func (p *PubSub) ListPeers(topic string) []peer.ID {
 // BlacklistPeer blacklists a peer; all messages from this peer will be unconditionally dropped.
 func (p *PubSub) BlacklistPeer(pid peer.ID) {
 	p.blacklistPeer <- pid
+}
+
+// validation requests
+type validateReq struct {
+	vals []*topicVal
+	src  peer.ID
+	msg  *Message
 }
 
 // per topic validators
--- a/pubsub.go
+++ b/pubsub.go
@@ -22,8 +22,7 @@ import (
 )
 
 const (
-	defaultValidateTimeout     = 150 * time.Millisecond
-	defaultValidateConcurrency = 100
+	defaultValidateConcurrency = 1024
 	defaultValidateThrottle    = 8192
 )
 
@@ -227,7 +226,7 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 }
 
 // WithValidateThrottle sets the upper bound on the number of active validation
-// goroutines.
+// goroutines across all topics. The default is 8192.
 func WithValidateThrottle(n int) Option {
 	return func(ps *PubSub) error {
 		ps.validateThrottle = make(chan struct{}, n)
@@ -966,7 +965,8 @@ type Validator func(context.Context, peer.ID, *Message) bool
 // ValidatorOpt is an option for RegisterTopicValidator.
 type ValidatorOpt func(addVal *addValReq) error
 
-// WithValidatorTimeout is an option that sets the topic validator timeout.
+// WithValidatorTimeout is an option that sets a timeout for an (asynchronous) topic validator.
+// By default there is no timeout in asynchronous validators.
 func WithValidatorTimeout(timeout time.Duration) ValidatorOpt {
 	return func(addVal *addValReq) error {
 		addVal.timeout = timeout
@@ -974,7 +974,8 @@ func WithValidatorTimeout(timeout time.Duration) ValidatorOpt {
 	}
 }
 
-// WithValidatorConcurrency is an option that sets topic validator throttle.
+// WithValidatorConcurrency is an option that sets the topic validator throttle.
+// This controls the number of active validation goroutines for the topic; the default is 1024.
 func WithValidatorConcurrency(n int) ValidatorOpt {
 	return func(addVal *addValReq) error {
 		addVal.throttle = n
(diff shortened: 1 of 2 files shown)


vyzo:
test failures are known, will fix in a separate pr.

pubsub.go
```diff
@@ -660,14 +690,27 @@ func (p *PubSub) validate(vals []*topicVal, src peer.ID, msg *Message) {
...
+	}
+
 	if len(vals) > 0 {
-		if !p.validateTopic(vals, src, msg) {
-			log.Warningf("message validation failed; dropping message from %s", src)
-			return
+		select {
+		case p.validateThrottle <- struct{}{}:
+			go func() {
+				p.doValidateTopic(vals, src, msg)
```
Stebalien: Doesn't this defeat the point of having a worker pool? Is there any reason to spawn this as a goroutine?

pubsub.go
```diff
@@ -209,6 +217,11 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 
 	go ps.processLoop(ctx)
 
+	numcpu := runtime.NumCPU()
+	for i := 0; i < numcpu; i++ {
```
yutongp: can this `numcpu` be a configurable number?

pubsub.go
```diff
@@ -183,6 +190,7 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		topics:           make(map[string]map[peer.ID]struct{}),
 		peers:            make(map[peer.ID]chan *RPC),
 		topicVals:        make(map[string]*topicVal),
+		validateQ:        make(chan *validateReq, 32),
```
mhchia: Should the size of `validateQ` the same as or greater than `defaultValidateThrottle`? Though it seems `defaultValidateThrottle` is the potential max number of the concurrent goroutines previously.

pubsub.go
vyzo: Yes, it kind of does. But at the moment all registered validators are asynchronous which means they can block, and we need to execute them in a goroutine.

Having said that, this is an artifact of WIP; I want to separate the validators to synchronous and asynchronous, and execute the synchronous ones inline.
The asynchronous can be in a goroutine or even better in an asynchronous api that doesn't use throttles at all.

pubsub.go
vyzo: We could make it configurable, but is there any benefit in using more than `NumCPU` synchronous validation workers?

pubsub.go
vyzo: I don't think so, or we will have the same problems with overloading we have previously.

pubsub.go
vyzo: Actually, since we are going to spawn a goroutine for async validators, we can still use the throttles.

pubsub.go
vyzo: So to explain: the throttle controls the number of active asynchronous validation requests, which can take a while to complete.
The queue controls the front end to the validation pipeline, which can  drop messages if we are overloaded - see #172.

pubsub.go
raulk: If the validation function is not CPU-bound, yes. That’s the case for Eth2.0 folks.

vyzo: These validators will be asynchronous and be executed in a goroutine.
This is the number of workers for the validation front-end; it handles signature verification, synchronous validators (TBD), and spawning goroutines for async validators with the throttle.

vyzo:
I've removed the WIP moniker, this is ready for review.

pubsub.go
vyzo: Added inline (synchronous) validation supprt.

pubsub.go
Stebalien: Hm. But why not just have 1024 synchronous validators (well, dynamically adjusting)?

vyzo: The asynchronous validators can block at will, I don't think we should have pinned goroutines for them. The cost of actually spawning the goroutine shouldn't be too high.

Also, if we had that many synchronous goroutines we could overtax the system doing synchronous validation.
That is we could end up validating signatures for 1024 messages concurrently which is not the intention (oversubscribes the cpu by a lot).
Perhaps we could have a second pool with prespawned (or dynamically adjustable) goroutines for the async validators but I don't think we gain much for the complexity.

pubsub.go
yutongp: If we give a number less than `numcpu`, would that help other components of the program utilize  cpu resource better?

vyzo: hrm maybe. We can add an option to make it less of a hog.

vyzo: Added option to control the number of workers.

pubsub.go
Stebalien: I see. So we're assuming that some validators will be CPU bound and others will won't. Async validators are those non-CPU bound validators.

This is really where async/await shines.

>  Perhaps we could have a second pool with prespawned (or dynamically adjustable) goroutines for the async validators but I don't think we gain much for the complexity.

We can revisit this later.

vyzo: We might want to allow multiple validators per topic: one could register an inline validator that fails fast with basic structural validation and an async validator that fetches stuff from the network for deep validation.

pubsub.go
mhchia: Big thanks for the clear explanation. Are my following understandings correct?
- `validatorQ` with the workers serve as the rate-limiting gate
- Except for the removal of the timeout, the asynchronous verification part is roughly the same as the one before this PR(except for the fact that signature verification is done synchronously by the workers).

vyzo: yes, this is correct.

raulk (commented):
Had a quick look at this, shaping up well!

Now that validation is becoming more complex, could we please encapsulate it behind a `ValidationPipeline` type?

The logic is a bit scattered over the place now, and it's difficult to navigate how the different aspects interplay with one another, e.g. `inline`, `timeout`, `throttles`, concurrency, etc.

raulk:
On a related note, do you feel strongly about async validator throttling being enforced by pubsub?

I think this may be an artificial need arising from us spawning a single goroutine per validator call -- once you do that you need a mechanism to prevent DoS, and hence you introduce a throttling mechanism.

If, however, we made async validators truly async by using channels, each validator would be in charge of accepting or rejecting jobs, and could use a different (more efficient) concurrency model instead of 1 goroutine per task.

vyzo:
> Now that validation is becoming more complex, could we please encapsulate it behind a ValidationPipeline type?

We can certainly do that. Or we can lump all validation related logic in its own file, which ~would~ may have similar effects in terms of readability.

> On a related note, do you feel strongly about async validator throttling being enforced by pubsub?

I don't have particularly strong views, other than having a throttle somewhere.
The trade-off is in the complexity of writing validation routines, which has a really simple api now.

vyzo:
I have split off the validation logic into its own component, per @raulk's request.

Note that the diff has become rather dirty because of this, and it's hard to see exactly what changed.
You might have to read each commit individually.

vyzo:
Rebased on master for test fixes in #178.

validation.go
```diff
@@ -0,0 +1,383 @@
...
+		delete(v.topicVals, topic)
+		req.resp <- nil
+	} else {
+		req.resp <- fmt.Errorf("No validator for topic %s", topic)
+	}
+}
+
+// Push pushes a message into the validation pipeline.
+// It returns true if the message can be forwarded immediately without validation.
+func (v *validation) Push(src peer.ID, msg *Message) bool {
```
raulk, line 143: nit: can we call this `Validate` for clarity?

validation.go
```diff
@@ -0,0 +1,383 @@
...
+	resp  chan error
+}
+
+// newValidation creates a new validation pipeline
+func newValidation() *validation {
+	return &validation{
+		topicVals:        make(map[string]*topicVal),
+		validateQ:        make(chan *validateReq, 32),
+		validateThrottle: make(chan struct{}, defaultValidateThrottle),
+		validateWorkers:  runtime.NumCPU(),
```
raulk, line 85: Could we document somewhere that the defaults assume CPU-intensive validation, and that the worker pool length should be adjusted for IO-bound tasks? Godoc on the `WithValidateWorkers` method should suffice.

validation.go
```diff
@@ -0,0 +1,383 @@
...
+
+		vals = append(vals, val)
+	}
+
+	return vals
+}
+
+// validateWorker is an active goroutine performing inline validation
+func (v *validation) validateWorker() {
+	for {
```
raulk, line 176: Do we want to use a waitgroup here for an orderly shutdown of workers?

validation.go
```diff
@@ -0,0 +1,383 @@
...
+		select {
+		case req := <-v.validateQ:
+			v.validate(req.vals, req.src, req.msg)
```
raulk, line 179: ~I think validate should take a context. Otherwise we risk blocking a system shutdown if a lengthy validation is inflight.~

Ah, nevermind. I see that `validate` does pass in the pubsub context to the validator itself.

validation.go
```diff
@@ -0,0 +1,383 @@
...
+	defaultValidateThrottle    = 8192
+)
+
+// Validator is a function that validates a message.
+type Validator func(context.Context, peer.ID, *Message) bool
+
+// ValidatorOpt is an option for RegisterTopicValidator.
+type ValidatorOpt func(addVal *addValReq) error
+
+// validation represents the validator pipeline
```
raulk: Some more colour on this doc would be nice. Suggestion: "the validator pipeline runs a sequence of user-configured validators per-topic. It is possible to adjust various concurrency parameters, such as the number of workers and the max number of simultaneous validations. The user can also attach inline validators that will be executed synchronously; this may be useful to prevent superfluous context-switching for lightweight tasks."

validation.go
```diff
@@ -0,0 +1,383 @@
...
+	err := verifyMessageSignature(msg.Message)
+	if err != nil {
+		log.Debugf("signature verification error: %s", err.Error())
+		return false
+	}
+
+	return true
+}
+
+func (v *validation) doValidateTopic(vals []*topicVal, src peer.ID, msg *Message) {
```
raulk, line 252: Can we remove this method and inline its code in its sole call site? (unless it's used from tests or we expect to do so) Makes it easier to read.

validation.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	return true
+}
+
+// fast path for single topic validation that avoids the extra goroutine
+func (v *validation) validateSingleTopic(val *topicVal, src peer.ID, msg *Message) bool {
+	select {
+	case val.validateThrottle <- struct{}{}:
+		ctx, cancel := context.WithCancel(v.p.ctx)
```
raulk: Is this new context necessary? `validateMsg` already creates one with a deadline if a timeout is configured.

raulk (approved):
Sorry for the long turn-around! This looks really good, @vyzo. The separation of the validation logic is clean, and I'm excited to roll this out. Just a few small items, but nothing blocking from me. Good job.

pubsub.go
raulk: Ultimately I'd like to see something like a [claim-check pattern](https://www.enterpriseintegrationpatterns.com/patterns/messaging/StoreInLibrary.html) here, but we can iterate towards that later. In summary:

1. The validator pipeline sends the message to the validator via a channel. The message includes the message ID, a deadline until which a result will be accepted, and a response channel.
2. The validator is free to manage its own goroutine pool, without the pipeline making assumptions whether the task is CPU or IO-bound.
3. The validator performs validation.
4. The validator returns the result to the pipeline via a channel, reporting the message ID and the result (OK/KO).

validation.go
vyzo: sure, why not?

validation.go
vyzo: Well, the validateWorkers perform a cpu-bound task by default, which is signature validation.
But better docs on `WithValidateWorkers` can do.

validation.go
vyzo: There may be no timeout configured (there is none by default), so we do need the context.

validation.go
vyzo: We could, but is it necessary?

validation.go
vyzo: ok, will do.

validation.go
vyzo: ha! I would never have expected a request to remove a small function in favour of inline expansion.
It will turn `validate` into a little monster though -- I find it easier to read with this in its own function.

validation.go
vyzo: Actually, it's a little confusing to call it `Validate` because it doesn't perform the validation, it simply pushes it to the validation queue.
The return value is a bool, which would be confusing as a `false` value doesn't mean the message is invalid, but rather it means that the message was scheduled for validation.

vyzo: A perhaps better name would be `ScheduleValidation`, but that's comically longer than simply calling it `Push`.

validation.go
vyzo: done.

validation.go
vyzo: improved the godoc for `WithValidateWorkers`.

vyzo:
I am going to rebase for merge.

validation.go
vyzo: Hrm, actually you are right. We don't _need_ this context. I will remove it.

vyzo: done.

validation.go
raulk: I find that `do*` functions tend to exist only for reuse purposes, i.e. they're normally inner routines you extract for reuse, but in this case it's only called once. But this is really a nit; the compiler would probably inline this anyway.

PR #181, opened 2019-05-08
Test adversarial signing

rargulati:
This PR introduces a test where an adversarial peer can sign a message with a key they
didn't originally register with.

First, an adversarial peer will allow the improperly signed message to pass through
validation (as they turn off strict verification, putting themselves at risk!).

Next, it's tested that an honest peer, with strict verification on, will never see the
malicious message, but they will see the correctly signed message.

We (@keep-network) just stripped out our custom message signing in favor of 
`go-libp2p-pubsub`'s strict message signing and verification. This test seemed 
useful to contribute to mainline.

### DIFF
--- a/floodsub_test.go
+++ b/floodsub_test.go
@@ -15,7 +15,6 @@ import (
 	protocol "github.com/libp2p/go-libp2p-protocol"
 	swarmt "github.com/libp2p/go-libp2p-swarm/testing"
 
-	//bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	bhost "github.com/libp2p/go-libp2p-blankhost"
 )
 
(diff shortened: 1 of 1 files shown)


floodsub_test.go
```diff
@@ -12,10 +12,10 @@ import (
 
 	host "github.com/libp2p/go-libp2p-host"
 	peer "github.com/libp2p/go-libp2p-peer"
+	protocol "github.com/libp2p/go-libp2p-protocol"
 	swarmt "github.com/libp2p/go-libp2p-swarm/testing"
-	//bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+
```
vyzo: can we get rid of this empty line?

vyzo (approved):
this is great, thank you! 
just a small nit about an empty line :)

floodsub_test.go
rargulati: Done in https://github.com/libp2p/go-libp2p-pubsub/pull/181/commits/3c78c237ce2e621b839bf70e3587f0e10c2e46b2 !

raulk:
@rargulati Thank you for the great contribution! 🙌 Keep them coming.

Issue #28, opened 2017-07-15
Subscribe and Relay should be separate operations.


Stebalien:
Not all subscribers will want to relay (bandwidth reasons) and not all relays will want to subscribe (they may just want to forward values without looking at them).

film42:
I'm new to ipfs and libp2p, but I was thinking this could set this at the subscription level. For a not great example:
```go
// Subscribe and relay by default.
pubsub.Subscribe("topic1")
// maybe implemented as:
td := pb.TopicDescriptor{Name: "topic1", Relay: true, Subscriber: true}
pubsub.SubscribeByTopicDescriptor(&td)

// Relay without reading messages.
pubsub.Relay("topic1")
// maybe implemented as:
td := pb.TopicDescriptor{Name: "topic1", Relay: true, Subscriber: false}
pubsub.SubscribeByTopicDescriptor(&td)

// Subscribe without relaying messages.
td := pb.TopicDescriptor{Name: "topic1", Relay: false, Subscriber: true}
pubsub.SubscribeByTopicDescriptor(&td)
```
I'm sure there's a better way to do this because having different flags makes things a bit cumbersome.

But, if you then had a wild card topic `"*"` you could become a dedicated relay or sink.

vyzo:
I think it will be a great feature to detach subscription with routing, especially with gossipsub.
We can use it to offer pubsub services in the core network, without spawning new goroutines and channels for them.

We can add two methods to the pubsub interface to this effect:
```
Join(topic string)
Leave(topic string)
```

I will work on it after gossipsub, pr as it maps nicely to the new router Join/Leave methods it adds.

aarshkshah1992:
Hey @aschmahmann @vyzo 
 
Thanks for the beautifully written [gossipsub spec](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub). 

Now, I understand that a peer publishes a message with topic `T` to peers in `fanout[T]` ONLY if it is NOT subscribed to `T` AND the message originates from itself. Even in this case, it is acting as a 'relayer' of the message even though it is not a 'subscriber' of `T`.

If my understand is correct, this issue is to enable a peer to choose one of three modes:

1. **Relayer**: Receive & forward a message to `fanout[T]` even if it hasn't subscribed to `T` AND the message does NOT originate from `T` 
2. **Subscriber**: Receive a message of type `T` without having to forward it to `mesh[T]` or gossip to `randomD(gossip[T] - mesh[T])`
[...]

PR #184, opened 2019-05-20
Add Discovery

aschmahmann:
This adds bootstrapping and the ability for rediscovery into the PubSub system as described in libp2p/go-libp2p-pubsub-router#28.

Once we go forward with this we can modify go-libp2p-pubsub-router to utilize an arbitrary discovery mechanism instead of the built in content routing based discovery. We likely will also need to implement a number of encapsulating discovery wrappers (e.g. tiered, parallel, etc.) like we have for content routing.

@raulk @vyzo @Stebalien


pubsub.go
```diff
@@ -397,13 +433,69 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+				defer cancel()
+
+				err := p.host.Connect(ctx, pi)
+				if err != nil {
+					log.Debugf("Error connecting to pubsub peer %s: %s", pi.ID, err.Error())
+					return
+				}
+
+				// delay to let pubsub perform its handshake
+				time.Sleep(time.Millisecond * 250)
```
Stebalien: We shouldn't need to wait here (it's not like we're doing anything but printing a debug line after).

pubsub.go
```diff
@@ -130,6 +150,9 @@ type PubSubRouter interface {
 	// Leave notifies the router that we are no longer interested in a topic.
 	// It is invoked after the unsubscription announcement.
 	Leave(topic string)
+	// Bootstrap gets the bootstrapping discovery parameters from the router.
+	// It also sets the mechanism for retrieving new discovery requests for the topic.
+	Bootstrap() []discovery.Option
```
Stebalien: Do routers really need to be able to configure discovery options? IMO, this is more of a global option. That is, that is, "discovery" is just a background service that:

1. Tells the network we're interested in a topic.
2. Connects to more peers when asked.

pubsub.go
```diff
@@ -710,10 +809,72 @@ func (p *PubSub) Publish(topic string, data []byte) error {
...
+	p.bootstrap(topic)
 	p.publish <- &Message{m}
+
 	return nil
 }
 
+// bootstrap handles initial bootstrapping of peers for a given topic
+func (p *PubSub) bootstrap(topic string) {
+	_, bootstrapped := p.bootstrapped.LoadOrStore(topic, struct{}{})
+	if !bootstrapped {
```
Stebalien: Any reason to call this up-front instead of just letting the router tell us when to call it? This will:

1. Force a discovery even if we have peers.
2. Not discover even if we've disconnected from all relevant peers.

If we just want to wait to connect to peers before publishing, we should consider doing that in the router. That is, have the router synchronously ask for more peers when publishing to an empty peer list.

pubsub.go
aschmahmann: This section of code was moved from https://github.com/libp2p/go-libp2p-pubsub-router/blob/76fb69228e0bb14cafe1f25c80301f97dd12871f/pubsub.go#L378. It looks from there like the idea was to let the PubSub notifier detect and add the new peer before proceeding after the bootstrap and that the same logic would apply here (since we publish immediately after this function returns). If you think it's unnecessary I would buy that though.

pubsub.go
aschmahmann: For better or worse this behavior mimics what we've been doing so far with some handles for making things better going forward. The reason I've left things this way is basically because PubSub has no persistence mechanism meaning that whoever you're connected to when you send out a message is who will receive the message and after that it's game over. So for the sake of keeping behavior the same during the refactor I've just left things as is, although if you'd like I'm happy to fix it up so that we don't force a discovery if we have peers (easily solved by passing the topic into the router's Bootstrap() command). 

> Not discover even if we've disconnected from all relevant peers.

We may actually discover even if we've disconnected from all peers (unlike the current implementation) since we have a "rediscovery" mechanism that the routers can trigger.

> we should consider doing that in the router

I'd like to put the logic in PubSub instead of the routers if we can since then we only need to write the code once for all of the routers.

>That is, have the router synchronously ask for more peers when publishing to an empty peer list.

This is a change from current behavior so as long as @vyzo is good with it so am I

pubsub.go
aschmahmann: Yes, but to be fair I'm not sure I understand discovery.Option in general.

There seem to only be two Option implementations, `Limit` and `TTL`, and `TTL` doesn't seem to be used by the one implementation in the package. I was planning on using `Limit` so that the router could ask for a particular number of peers (e.g. maybe floodsub wants more peers than gossipsub).

aschmahmann: Given our conversation regarding moving the complexity from the router level to the validator and application level perhaps the discovery options (like Limit) could be passed in on a per-topic basis. This could be done by making another register function similar to "RegisterTopicValidator", by adding fields to Subscription and utilizing a SubOpt, or something in between.

pubsub.go
Kubuxu: Sleeping is never a good option. There is no guarantee that 250ms is enough.
I understand that this is an old code but it probably should be solved in another way.

aschmahmann: Fair enough, now that this code is in pubsub instead of in go-libp2p-pubsub-router we should have enough info to not need to sleep

Stebalien: The difference is the lack of a waitgroup in this case.

pubsub.go
Stebalien: > We may actually discover even if we've disconnected from all peers (unlike the current implementation) since we have a "rediscovery" mechanism that the routers can trigger.

IMO, this should be the _only_ mechanism. That is, routers should have to say "please bootstrap me".

> I'd like to put the logic in PubSub instead of the routers if we can since then we only need to write the code once for all of the routers.

Does pubsub know enough to do this?

pubsub.go
Stebalien: This relates to https://github.com/libp2p/go-libp2p-pubsub/pull/184#discussion_r285807815. It's one of the reasons I believe the router should be the one asking for more peers.

Basically, it's up to routers to form the pubsub network. They may even have alternative/preferred discovery mechanisms.

pubsub.go
aschmahmann: yep, that's my bad. If I'm not going to remove the delay then I need to add the waitgroup back in. Any preferences @Stebalien

pubsub.go
aschmahmann: > Does pubsub know enough to do this?

Yes/It depends. PubSub knows how many peers are subscribed to a given topic and has access to a discovery mechanism. If your router had a data structure that tracked network topology then the way you would signal "I need 2 good peers or 4 bad ones" is by asking for 2 and then asking for more if what you have is insufficient.

It's also worth noting that we may want some of this to come from the application layer. For example, with a persistent pubsub we may feel comfortable publishing to fewer peers than without persistence. This could be because in a persistent pubsub even one peer can eventually get the message out, while in an ephemeral pubsub once a message is processed it is dead.

> IMO, this should be the only mechanism. That is, routers should have to say "please bootstrap me".

Not an unreasonable request. It's a bit tricky though since we can't block in the routers as they all share the same event loop. As a result, we may have to do some gross async/callback things. If you think this is the way to go (now as opposed to dealing with this PR first and adding an issue) let me know and I'll dig more into it.

pubsub.go
aschmahmann: > Basically, it's up to routers to form the pubsub network. They may even have alternative/preferred discovery mechanisms.

Maybe... under the premises of how I wrote #171 I agreed with you. However, after our conversations with @raulk and @vyzo about moving complexity from new routers into per-topic validators the landscape looks different. Maybe floodsub/gossipsub/episub cares about the discovery mechanisms, but maybe lww-persistence/DAG-persistence/validator that only propagate every third message/etc. cares about the discovery mechanisms.

If we're prepared to go back to "all interesting logic is in the routers" then I'm happy to put discovery there. However, given the addition of networking logic from the application layer (e.g. via stateful validators as we're planning on using for persistence) we might be better off with something like a TopicConfig that ends up containing all the smarts so that the discovery strategy isn't tied to floodsub/gossipsub/etc. but to a struct/function that implements that logic.

vyzo: This looks like a pretty useless method as it has empty implementations, so let's remove it.

pubsub.go
```diff
@@ -710,10 +809,72 @@ func (p *PubSub) Publish(topic string, data []byte) error {
...
+					}
+				case <-advertisingCtx.Done():
+					return
+				}
+			}
+		}()
+	}
+}
+
+type Rediscover struct {
```
vyzo: this should be called just `Discover`.

pubsub.go
```diff
@@ -710,10 +809,72 @@ func (p *PubSub) Publish(topic string, data []byte) error {
...
+
+		go func() {
+			next, err := p.discovery.Advertise(advertisingCtx, topic)
+			if err != nil {
+				log.Warningf("bootstrap: error providing rendezvous for %s: %s", topic, err.Error())
+			}
+
+			for {
+				select {
+				case <-time.After(next):
```
vyzo: Can you use a ticker for this?

gossipsub.go
```diff
@@ -410,6 +415,14 @@ func (gs *GossipSubRouter) heartbeat() {
 			}
 		}
 
+		// if we still don't have enough peers request discovering more
+		if len(peers) < GossipSubDlo {
+			rediscoverTopic := topic
+			go func() {
+				gs.p.rediscover <- &Rediscover{rediscoverTopic, []discovery.Option{discovery.Limit(GossipSubDhi)}}
```
vyzo: it's ok to discover more peers, let's not limit to just `Dhi`.

pubsub.go
```diff
@@ -13,11 +14,12 @@ import (
 
 	logging "github.com/ipfs/go-log"
 	crypto "github.com/libp2p/go-libp2p-crypto"
+	discovery "github.com/libp2p/go-libp2p-discovery"
 	host "github.com/libp2p/go-libp2p-host"
 	inet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
 	protocol "github.com/libp2p/go-libp2p-protocol"
-	timecache "github.com/whyrusleeping/timecache"
+	"github.com/whyrusleeping/timecache"
```
vyzo: please don't remove this named import.

pubsub.go
```diff
@@ -98,6 +100,24 @@ type PubSub struct {
...
+	discovery discovery.Discovery
+
+	// advertising tracks which topics are being advertised
+	advertising map[string]context.CancelFunc
+
+	// bootstrapped tracks which topics have been bootstrapped
+	bootstrapped sync.Map
+
+	// rediscover handles continuing peer rediscovery
+	rediscover chan *Rediscover
```
vyzo: `rediscover` is a silly name, let's call this just `discover`.

vyzo (changes requested):
Let's move the discovery logic to its own component/file.

vyzo:
Just to clarify the change request, I would like the discovery logic in its own component instead of piling up on the pubsub struct (and file).
See how we handle validation.

pubsub.go
aschmahmann: Yes, but the behavior changes Notice that `next` is (re)set in the next line. This basically just advertises whenever the TTL on our previous advertisement expires. Another option would be to change this to refresh every X hours and make X configurable by topic.

Let me know what you think

gossipsub.go
aschmahmann: Given that Dhi, Dlo, etc. are not global variables instead of construction parameters I figured this was likely to be yet another global variable and I'm open to what you'd like that to be. I'd prefer there to be a limit instead of an "unlimited" discovery option since there's no explicit contract on the Discover channel even closing.

Yes, we have a context currently set to 10 seconds for Discovery, but it feels a bit awkward for waiting 10 seconds to potentially be the default (e.g. if I did a joint DHT + Rendezvous search). Having stateful clients that buffer the "extra" clients seems sufficient and since we can't process/handle more than `Dhi` peers anyway I figure let's let the discovery client worry about excess peers.

pubsub.go
aschmahmann: umm... sure. Although I was actually planning on doing the opposite. From what I can tell since the go-libp2p-core refactor all of the named imports have been removed in favor of unnamed imports (if the names would match and not cause problems). e.g. look at go-libp2p-swarm

pubsub.go
aschmahmann: ok

pubsub.go
aschmahmann: ok

aschmahmann:
> Just to clarify the change request, I would like the discovery logic in its own component instead of piling up on the pubsub struct (and file).
> See how we handle validation.

Seems reasonable, however, it like makes fixing https://github.com/libp2p/go-libp2p-pubsub/pull/184#discussion_r289939817 even more difficult (i.e. without looking like a mess of private channels between the PubSub and discover structs). You ok if we punt on that for now and just re-add the waitgroup?

pubsub.go
aschmahmann: As indicated above it would be great to be able to say "persistent channels only need to wait for 1-2 peers during bootstrapping instead of 12". Originally I was going to do this via a new router (as in #171), which means that the implementations here would be empty and there would not.

Given that, per our discussion, I've now rewritten #171 to use stateful validators + external protocols this doesn't look super useful. **However**, I'd like to replace this with a PubSub function for per-topic configurations (similar in nature to how topic validators, which we're also using to influence message propagation, are per-topic).

aschmahmann: Plan is to punt on this for this PR. While it hurts that bootstrapping before publishing can be unnecessarily slow (e.g. waiting for the 10s timeout to find 100 peers when we know we only need 5), having bootstrapping for subscribe be asynchronous takes away the most painful part of the experience.

gossipsub.go
aschmahmann: I added a global var `GossipSubNPeers = 100` to replace `GossipSubDhi` here. That sufficient?

pubsub.go
vyzo: Let's use a timer and reset it then.

gossipsub.go
vyzo: Probably ok for now.

pubsub.go
aschmahmann: Clarification: Because `time.After` uses a ticker internally, this is really about having a single ticker and calling `ticker.Reset(newDuration)` and therefore removing the extra allocation every time we advertise.

Easy enough fix.

pubsub.go
aschmahmann: for now we will use a policy of named imports for non-libp2p and non-standard packages, so `timecache` will go back in

pubsub.go
Kubuxu: Make also sure to close tickers in cases of ctx being done and so on.
We had an out of memory problem before because of unclosed tickers staying in memory for the duration of the tick.

aschmahmann: interesting would've thought GC would handle that. Thanks for the heads up.

Kubuxu: GC doesn't handle that because the timer creation spaws a goroutine, this goroutine then reference to the ticker.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+	ongoingDiscovery map[string]struct{}
+
+	// doneDiscovery handles completion of a discovery request
+	doneDiscovery chan string
+}
+
+// newValidation creates a new validation pipeline
+func newDiscover() *discover {
+	return &discover{
+		bootstrapped:     sync.Map{},
```
vyzo: this is unnecessary as it's the default initializer.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+	advertising map[string]context.CancelFunc
+
+	// bootstrapped tracks which topics have been bootstrapped
+	bootstrapped sync.Map
+
+	// discover handles continuing peer discovery
+	discover chan *Discover
+
+	// ongoingDiscovery tracks ongoing discovery requests
+	ongoingDiscovery map[string]struct{}
```
vyzo: call this just `ongoing`.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// doneDiscovery handles completion of a discovery request
+	doneDiscovery chan string
```
vyzo: and call this just `done`.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+				d.ongoingDiscovery[topic] = struct{}{}
+				go func() {
+					d.handleDiscovery(d.p.ctx, topic, discover.opts)
+					d.doneDiscovery <- topic
+				}()
+			}
+		case topic := <-d.doneDiscovery:
+			delete(d.ongoingDiscovery, topic)
+		case <-d.p.ctx.Done():
+			log.Info("pubsub discoverloop shutting down")
```
vyzo: no need to log that.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+			return
+		}
+	}
+}
+
+// advertise advertises this node's interest in a topic to a discovery service
+func (d *discover) advertise(topic string) {
+	if d.discovery != nil {
```
vyzo: why this check? It shouldn't start at all if the `discovery` object is `nil`.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+func (d *discover) stopAdvertise(topic string) {
+	if advertiseCancel, ok := d.advertising[topic]; ok {
+		advertiseCancel()
+		delete(d.advertising, topic)
+	}
+}
+
+// bootstrap handles initial bootstrapping of peers for a given topic
+func (d *discover) bootstrap(async bool, topic string, dOpts ...discovery.Option) {
+	newMux := &sync.Mutex{}
```
vyzo: I don't understand the use of this mutex; it doesn't seem to be protecting any critical section as it's a brand new object each time.

discovery.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+
+		case <-discoverCtx.Done():
+			log.Infof("pubsub discovery for %s timeout", topic)
+			return
+		}
+	}
+
+	if len(findPeers) > 0 {
+		req := &reqNewPeerNotifier{initPeers: findPeers, resp: make(chan chan peer.ID, 1)}
```
vyzo: why are you doing this? the event loop will receive connection notifications anyway.

pubsub.go
```diff
@@ -72,6 +75,18 @@ type PubSub struct {
 	// a notification channel for when our peers die
 	peerDead chan peer.ID
 
+	// tracks the channels subscribed to new peer notifications
+	newPeerNotif map[chan peer.ID]struct{}
+
+	// setup new peer notifier
+	addPeerNotif chan *reqNewPeerNotifier
```
vyzo: what's this?

pubsub.go
```diff
@@ -72,6 +75,18 @@ type PubSub struct {
...
+
+	// remove peer notifier
+	rmPeerNotif chan chan peer.ID
+
+	// peers that we have connected to, but have not received a Hello from
+	inflightPeers map[peer.ID]struct{}
```
vyzo: no, we shouldn't be tracking those. The connection notifications will issue hello.

pubsub.go
```diff
@@ -72,6 +75,18 @@ type PubSub struct {
...
 	peerDead chan peer.ID
 
+	// tracks the channels subscribed to new peer notifications
+	newPeerNotif map[chan peer.ID]struct{}
+
+	// setup new peer notifier
+	addPeerNotif chan *reqNewPeerNotifier
+
+	// remove peer notifier
+	rmPeerNotif chan chan peer.ID
```
vyzo: and this?

pubsub.go
```diff
@@ -329,15 +364,35 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
 			delete(p.peers, pid)
+			delete(p.inflightPeers, pid)
 			for _, t := range p.topics {
 				delete(t, pid)
 			}
 
 			p.rt.RemovePeer(pid)
+			p.notifyNewPeer(pid)
+
+		case add := <-p.addPeerNotif:
```
vyzo: this logic is completely redundant, we already handle this with connection notifications.

randomsub.go
```diff
@@ -2,7 +2,6 @@ package pubsub
 
 import (
 	"context"
-
```
vyzo: don't remove this line.

vyzo (changes requested):
It's great progress to see the discovery stuffs in its own module, but I am totally confused about this new peer notification stuff you are adding.
That looks completely redundant, as we already have logic to handle this with connection notifications.

discovery.go
aschmahmann: It's just a question of where to put the `nil` checks. Either I put them in the `discover` functions, or I put them in pubsub (e.g. in `handleAddSubscription`) I opted to remove any extra lines of code from pubsub and put them here instead.

If you disagree I can move it, just let me know.

discovery.go
aschmahmann: It ensures that bootstrapping only occurs once per topic. `sync.Map.LoadOrStore` ensures that an item will only be added to the map once. This adding to the map indicates that bootstrapping has started, but we have no way to indicate it has completed.

The mutex does this by being Locked until bootstrapping completes at which point it is released. We could do this in other ways such as by having a second sync.Map also that indicates "bootstrapping has completed".

If you have a preference I'm open to suggestions.

discovery.go
aschmahmann: This request does two things.
1) I need to check not just inbound peers, but peers I'm already subscribed to. So this allows me to fill the channel with the intersection of peers I care about and peers I'm already connected to. This could be done with more memory usage by either sending all peers instead of the intersection or by trying to utilize the ListPeers method. Using the ListPeers method may be racy and a bit tricky to work around, but if we're ok being not perfectly precise here we can probably figure this out.
2) We need to be notified about incoming messages on a unique channel per topic/bootstrapping goroutine. We could also have a single discover channel that receives these messages and dispatches them to the goroutines. The way it was done seemed more flexible/reusable since it's more about peer-state then discovery, but this could be moved around.

If you'd like this to go away that's fine, just let me know which path you'd like to take.

pubsub.go
aschmahmann: The pubsub discovery module is asking to be notified of which peer streams we've connected to/failed to connect to. We need this to know when we're done trying to bootstrap. There may also be other uses to having a channel tracking new peers (currently we have `newPeerNotif map[chan peer.ID]struct{}` but it could easily be `newPeerNotif map[chan {peer.ID, Subscribed|Disconnected|etc.}]struct{}`

pubsub.go
aschmahmann: we track successful connections, but not really unsuccessful ones. We need the inflight peers to properly handle those.

pubsub.go
aschmahmann: as with any event subscription service you should be able to unsubscribe. this is called when we're done bootstrapping saying "no need to tell me about peer connections anymore"

pubsub.go
aschmahmann: I am not sure I understand where the redundancy comes in. Are you saying `notifyNewPeer` there will never convey information not handled in `notifyNewPeer` elsewhere (e.g. in the newPeerStream and newPeerErr channels)? Or is this a higher level "general redundancy"? Either way more explanation would help me understand.

aschmahmann:
> It's great progress to see the discovery stuffs in its own module, but I am totally confused about this new peer notification stuff you are adding.
> That looks completely redundant, as we already have logic to handle this with connection notifications.

This is needed in order to solve https://github.com/libp2p/go-libp2p-pubsub/pull/184#discussion_r289939817. In particular, if we initialize pubsub and the first thing we do is Publish to a topic then we'll start bootstrapping that topic before we actually send the message. If we send the message too soon it'll just be dropped since there won't be anyone in the network (or a large enough portion of the network) to hear the message. We used to just wait a little bit (250ms) between the end of `network.Connect` and when PubSub has received a `Subscribe` message and hope that was enough. This peer notification mechanism let's us actually know when we've finished running our bootstrap function.

pubsub.go
vyzo: No we shouldn't have to care about in flight peers and all the details of discovery, this leaks through the abstraction barrier.
The only thing we care about is whether bootstrap is complete or not and that can be handled with a single event in the pubsub mainloop.

vyzo:
The pubsub mainloop should have no details of discovery leaking.
All it needs is a single event, please track the peers internally in discovery if you have to.

pubsub.go
vyzo: this is unnecessary.

pubsub.go
vyzo: again, this is leaking through the abstraction barrier -- it's irrelevant to the mainloop.

pubsub.go
vyzo: we don't care about this in the mainloop.

discovery.go
vyzo: ok

discovery.go
vyzo: but you create a new mutex every time you call bootstrap, there is nothing protected by this.

discovery.go
vyzo: please take this out of the pubsub mainloop.

discovery.go
aschmahmann: clarification: the map signifies "have we started bootstrapping" and the mutex controls making sure we're only going to bootstrap once,

After discussing with @vyzo we're going to see how this looks if we use maps of channels instead of maps of mutexes.

discovery.go
```diff
@@ -0,0 +1,368 @@
+package pubsub
+
+import (
+	"context"
+	"github.com/libp2p/go-libp2p-core/discovery"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"sync"
+	"time"
```
vyzo: please separate the stdlib from our imports, leave an empty line to do so.

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+	newPeerNotif map[chan peer.ID]*peerEventTracker
+
+	// setup new peer notifier
+	addPeerNotif chan *reqNewPeerNotifier
+
+	// remove peer notifier
+	rmPeerNotif chan chan peer.ID
+}
+
+// newValidation creates a new validation pipeline
```
vyzo: incorrect comment.

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+	for {
+		select {
+		case discover := <-d.discoverQ:
+			topic := discover.topic
+
+			d.bootstrappedMux.RLock()
+			bootstrapDone := d.bootstrapped[topic]
+			d.bootstrappedMux.RUnlock()
+			select {
+			case _ = <-bootstrapDone:
```
vyzo: you don't need the dummy variable assignment, just `case <-bootstrapDone:`

vyzo: note that you also need to check whether you got a value, the nil value will trigger as done here.

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+			sentPeers := &peerEventTracker{peers: make(map[peer.ID]struct{})}
+			d.newPeerNotif[ch] = sentPeers
+			add.resp <- ch
+			go func() {
+				topicPeers := d.p.ListPeers(add.topic)
+				sentPeers.mux.RLock()
+				sendPeers := peerIntersect(topicPeers, sentPeers.peers)
+				sentPeers.mux.RUnlock()
+
+				sentPeers.mux.Lock()
```
vyzo: since you are RLocking and immediately write locking afterwards, just use a write lock.

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+		return
+	}
+
+	d.bootstrappedMux.RLock()
+	doneCh, bootstrapStarted := d.bootstrapped[topic]
+	d.bootstrappedMux.RUnlock()
+
+	if bootstrapStarted {
+		if !async {
+			_ = <-doneCh
```
vyzo: no need for the dummy assignment, just `<-doneCh`

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+	} else {
+		d.bootstrappedMux.Lock()
+		doneCh, bootstrapStarted := d.bootstrapped[topic]
+		if bootstrapStarted {
+			d.bootstrappedMux.Unlock()
+			if !async {
+				_ = <-doneCh
```
vyzo: here too

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	d.discoverQ <- &discoverReq{topic, opts}
+}
+
+func (d *discover) handleDiscovery(ctx context.Context, topic string, opts []discovery.Option) {
+	discoverCtx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	peerChan, err := d.discovery.FindPeers(discoverCtx, topic, opts...)
```
vyzo: `peerCh`

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+		return
+	}
+
+	findPeers := make(map[peer.ID]struct{})
+
+loop:
+	for {
+		select {
+		case pi, more := <-peerChan:
+			if pi.ID == d.p.host.ID() || pi.ID == "" {
```
vyzo: why would you get the local peer here?

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+			}
+
+		case <-discoverCtx.Done():
+			log.Infof("pubsub discovery for %s timeout", topic)
+			return
+		}
+	}
+
+	if len(findPeers) > 0 {
+		req := &reqNewPeerNotifier{topic: topic, resp: make(chan chan peer.ID, 1)}
```
vyzo: this is a mouthful, better name please!

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+
+func (d *discover) NotifyNewPeer(pid peer.ID) {
+	if d.discovery == nil {
+		return
+	}
+
+	select {
+	case d.peerNotifier <- pid:
+	default:
+		log.Error("dropped msg")
```
vyzo: better log here, what message did you drop?

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+	for _, pid := range peerArr {
+		if _, ok := peerSet[pid]; ok {
+			peerBuffer = append(peerBuffer, pid)
+		}
+	}
+
+	return peerBuffer
+}
+
+type reqNewPeerNotifier struct {
```
vyzo: pretty ugly type name!

discovery.go
```diff
@@ -0,0 +1,368 @@
...
+	topic string
+	resp  chan chan peer.ID
+}
+
+type peerEventTracker struct {
```
vyzo: that too.

gossipsub.go
```diff
@@ -293,7 +297,6 @@ func (gs *GossipSubRouter) Leave(topic string) {
 	log.Debugf("LEAVE %s", topic)
 
 	delete(gs.mesh, topic)
-
```
vyzo: please don't remove this empty line.

gossipsub.go
```diff
@@ -357,6 +360,7 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
 func (gs *GossipSubRouter) heartbeatTimer() {
 	time.Sleep(GossipSubHeartbeatInitialDelay)
 	select {
+
```
vyzo: and don't add this empty line.

pubsub.go
```diff
@@ -283,6 +296,7 @@ func (p *PubSub) processLoop(ctx context.Context) {
 
 			if p.blacklist.Contains(pid) {
 				log.Warning("ignoring connection from blacklisted peer: ", pid)
+				p.disc.NotifyNewPeer(pid)
```
vyzo: why notify for a blacklisted peer?

pubsub.go
```diff
@@ -305,13 +319,15 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				log.Warning("closing stream for blacklisted peer: ", pid)
 				close(ch)
 				s.Reset()
+				p.disc.NotifyNewPeer(pid)
```
vyzo: here too.

pubsub.go
```diff
@@ -305,13 +319,15 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
 				continue
 			}
 
 			p.rt.AddPeer(pid, s.Protocol())
 
 		case pid := <-p.newPeerError:
 			delete(p.peers, pid)
+			p.disc.NotifyNewPeer(pid)
```
vyzo: A little confused about notifying for a new peer when we got an error. The method is probably wrongly named.

discovery.go
aschmahmann: I can, I was mostly concerned about `peerIntersect` taking some time if we have lots of peers. Concerned about lock contention/clogging up the internal discovery event loop. If you think it's no big deal I can just replace the RWMutex with a regular one and call it a day, just figured this was a low cost - medium reward type of optimization.

discovery.go
aschmahmann: These are peers coming from the `Discovery` interface. Unless we make it an assertion that all discovery services will not return your Peer ID then we have to do this. For example, I advertise in my Peer ID in DHT/Rendezvous and then I receive it during a query.

If you'd prefer to add a comment to the discovery interface asserting the above that's fine by me. Let me know.

discovery.go
aschmahmann: @vyzo I don't think that's correct. nil channels block on read, so we'll go to the default and continue. (https://play.golang.org/p/GTDxZ6uLRzG)

discovery.go
aschmahmann: dropped the peer discovery notification message, will make the log message more detailed.

pubsub.go
aschmahmann: We need to know if when a bootstrap attempt failed so we can remove them from the queue of peers we're waiting on. Blacklisting is a type of failure, we can tag the successes and failures differently if we want, but since this is all internal I don't know if there's really a rush to do this now,

pubsub.go
aschmahmann: See above. Feel free to come up with a new name. We're notifying when a new peer connection attempt has completed, whether with success or failure.

discovery.go
vyzo: what's the contention model here?

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+
+	if _, ok := d.advertising[topic]; ok {
+		return
+	}
+	d.advertising[topic] = cancel
+
+	go func() {
+		next, err := d.discovery.Advertise(advertisingCtx, topic)
+		if err != nil {
+			log.Warningf("bootstrap: error providing rendezvous for %s: %s", topic, err.Error())
```
vyzo, line 189: shouldn't we short-circuit here as well?

vyzo (commented):
this is almost ready to merge I would say, let's get another pass from @Stebalien as well.

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+	p *PubSub
+
+	// discovery assists in discovering and advertising peers for a topic
+	discovery discovery.Discovery
+
+	// advertising tracks which topics are being advertised
+	advertising map[string]context.CancelFunc
+
+	// bootstrapped tracks which topics have been bootstrapped
+	bootstrapped map[string]chan struct{}
```
Stebalien: Why do we have a separate "bootstrap"?

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+	// discoverQ handles continuing peer discovery
+	discoverQ chan *discoverReq
+
+	// ongoing tracks ongoing discovery requests
+	ongoing map[string]struct{}
+
+	// done handles completion of a discovery request
+	done chan string
+
+	// peerNotifier processes all peer notifications
```
Stebalien: What's a peer notification?

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+}
+
+// Start attaches the discovery pipeline to a pubsub instance and starts event loop
+func (d *discover) Start(p *PubSub) {
+	if d.discovery == nil {
+		return
+	}
+
+	d.p = p
```
Stebalien, line 87: Ideally, this would check if d.p is not nil and refuse to start.

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+			case <-bootstrapDone:
+			default:
+				continue
+			}
+
+			if _, ok := d.ongoing[topic]; !ok {
+				d.ongoing[topic] = struct{}{}
+				go func() {
+					d.handleDiscovery(d.p.ctx, topic, discover.opts)
+					d.done <- topic
```
Stebalien: Should select on the context. This will hang on shutdown otherwise.

discovery.go
Stebalien: No? This seems right.

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+		case pid := <-d.peerNotifier:
+			d.notifyNewPeer(pid)
+		case <-d.p.ctx.Done():
+			return
+		}
+	}
+}
+
+// Advertise advertises this node's interest in a topic to a discovery service
+func (d *discover) Advertise(topic string) {
```
Stebalien, line 173: Can we explicitly note that this service isn't thread-safe?

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	findPeers := make(map[peer.ID]struct{})
+
+loop:
+	for {
+		select {
+		case pi, more := <-peerCh:
+			if pi.ID == d.p.host.ID() || pi.ID == "" {
+				if more {
```
Stebalien: This check should happen at the very top. That is:

```
case pi, ok := <-peerCh:
  if !ok {
    break loop
  }
...
```

`ok == false` means that the channel is closed and `pi` is invalid (well, the zero value).

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+				defer cancel()
+
+				err := d.p.host.Connect(ctx, pi)
+				if err != nil {
+					log.Debugf("Error connecting to pubsub peer %s: %s", pi.ID, err.Error())
+					return
+				}
+			}(pi)
+
+			if !more {
```
Stebalien: See above.

discovery.go
```diff
@@ -0,0 +1,369 @@
...
+			select {
+			case pid := <-peerCh:
+				if _, ok := findPeers[pid]; ok {
+					delete(findPeers, pid)
+					log.Debugf("Connected to pubsub peer %s", pid)
+					if len(findPeers) == 0 {
+						return
+					}
+				}
+			case <-discoverCtx.Done():
```
Stebalien: We're going to wait the full timeout unless we magically manage to connect to every single peer.

Stebalien: Anyways, it seems like we're tracking the wrong thing here. The goal is to find peers for the topic, not connect to all the peers that happened to be advertising the topic. What if we just kept a discovery worker running in the background till the pubsub router told us to turn it off? That is, `Discover(topic)` and `StopDiscover(topic)`. I'm not sure how well that would integrate with the current pubsub implementatoin but it would let us entirely drop the "new peer" logic.

Stebalien (changes requested):

discovery.go
Stebalien: Note: this is non-blocking. While not perfect, this module should work well enough (although I'd like to consider this as I think it would simplify the logic here significantly).

discovery.go
Stebalien: Ah, I see. I believe we've discussed this. This is probably fine for now but this isn't really the correct way to handle this. For example, I _shouldn't_ have to do _any_ bootstrapping if I already have peers on the topic.

discovery.go
aschmahmann: @Stebalien I can change the if statement to be `if d.discovery == nil || p == nil` if you'd like. If so should I add the same null check into `func (v *validation) Start(p *PubSub)`?

discovery.go
aschmahmann: We can certainly add in some comments that `Advertise` and `StopAdvertise` aren't thread-safe. Hopefully that doesn't imply that that the other functions (e.g. `Discover` and `Boostrap`) are totally thread-safe, they are thread-safe as long `d.discovery` isn't changed during use.

discovery.go
aschmahmann: @Stebalien While I agree with you in principle that this is correct, any change to this behavior potentially breaks people using PubSub currently if they just copy-pasted the bootstrap code from go-libp2p-pubsub-router.

If we don't think this is happening, then I'm perfectly fine telling all the standard non-persistent messaging schemes to figure out the "ready condition" on their own (persistent schemes don't really care since the initially sent messages won't get dropped). This would obviously make the code here which is a really nice plus. It's possible that #190 could be used as events to trigger the "ready condition", but is there anything else we should expose (so users don't just keep running ListPeers in a loop or waiting a fixed amount of time)?

@vyzo do you know how people are bootstrapping pubsub currently? Will changing behavior from go-libp2p-pubsub-router to make bootstrapping non-blocking be a problem?

aschmahmann: Also, we do not have to connect to every peer we just need to have a response from `connect` on every peer. For example, if connects fail by going to dead IP addresses then we'll get a Peer Dead notification which will also count as a completed attempt.

Stebalien: Discussed in-person. The resolution was to:

1. Drop all the "wait to bootstrap" code.
2. Provide options to the `Publish` function that wait for peers.
3. Trigger a search for more peers on publish, if we don't have enough.

discovery.go
Stebalien: My suggestion was to make sure that `Start` hasn't already been called by failing if `d.p` has already been set.

aschmahmann: Should I do the same to the validation pipeline as well?

discovery.go
aschmahmann: @Stebalien How's this modified version?

1. Drop all the "wait to bootstrap" code.
2. _Provide options to a new `SetTopicParameters` function (@vyzo has expressed concern about breaking the `Publish` function in the past, but we could definitely do that instead if people prefer)._
3. Trigger a search for more peers on publish, if we don't have enough. _(block until we have enough as defined in 2)_
4. _Trigger a search for more peers on continuously, if we don't have enough. This helps with discovering new updates._

I haven't figured out if it makes sense to have the continuously search function built into the Routers (as we do currently in the PR for gossipsub) or passed in by the developer. I'm thinking the developer should define the behavior since some protocols (like floodsub) are super oblivious to network conditions and have no idea how many peers they want to have. However, I can certainly be convinced otherwise.

Stebalien: > 1

:+1:

> 2

We could do that but it wouldn't be a _breaking_ change if we use a variable number of functional options (well, not that breaking).

> 3

Sgtm.

> 4

Regardless, we need a backoff on the continuous search. Probably something like:

1. Every time we search, double the delay up to some max.
2. Every time we loose a peer, reset this delay (if we're below the delay).

Given something like that, we can probably set the floodsub limit to something like 5.

discovery.go
Stebalien: Ideally? I mean, it's not critical, just a nice sanity check.

discovery.go
Stebalien: Although... if we block _at all_ inside publish, we really should have it take a context. Which will be a breaking change.

Honestly, we should seriously consider just making that change. Alternatively, we could just not do this here and, instead:

1. Add a that `Bootstrap(topic)` function.
2. Provide an external helper function `BootstrapWait(pubsub, topic, npeers)` and call it before calling publish. However, I'd _really_ prefer to make publish do this.

aschmahmann: Even if a little gross, that change doesn't need to necessarily be breaking since we could make the context specific to the bootstrap function passed into `Publish`. The result would probably be that after a timeout we send the message even if we don't think we're bootstrapped.

Stebalien: Sure, but I'd _rather_ break things than do something like that.

aschmahmann: fair enough. @vyzo are you onboard?

vyzo: sure, i am fine with breaking things here.

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+	// ongoing tracks ongoing discovery requests
+	ongoing map[string]struct{}
+
+	// done handles completion of a discovery request
+	done chan string
+}
+
+type IsDiscovered func(numPeers int) bool
+
+func CheckNumPeers(num int) IsDiscovered {
```
Stebalien: What does this do? (needs docs and maybe a better name)

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+	discoverQ chan *discoverReq
+
+	// ongoing tracks ongoing discovery requests
+	ongoing map[string]struct{}
+
+	// done handles completion of a discovery request
+	done chan string
+}
+
+type IsDiscovered func(numPeers int) bool
```
Stebalien: ???

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (d *discover) discoverLoop() {
+	for {
+		select {
+		case discover := <-d.discoverQ:
+			topic := discover.topic
+
+			if _, ok := d.ongoing[topic]; ok {
+				discover.done <- struct{}{}
```
Stebalien, line 150: Should we really mark this as done? Doesn't that make this channel completely useless?

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+// Advertise advertises this node's interest in a topic to a discovery service. Advertise is not thread-safe.
+func (d *discover) Advertise(topic string) {
+	if d.discovery == nil {
+		return
+	}
+
+	advertisingCtx, cancel := context.WithCancel(d.p.ctx)
```
Stebalien, line 178: We need to move this below the check below. Technically, not calling cancel could leak a goroutine in some cases.

Stebalien: Try running `golangci-lint` against this PR.

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+// Discover searches for additional peers interested in a given topic
+func (d *discover) Discover(topic string, opts ...discovery.Option) {
+	if d.discovery == nil {
+		return
+	}
+
+	d.discoverQ <- &discoverReq{topic, opts, make(chan struct{})}
```
Stebalien: Channel needs a buffer.

discovery.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// Bootstrap attempts to bootstrap to a given topic. Returns true if it was already bootstrapped, false otherwise.
```
Stebalien: What does "already bootstrapped" mean? Does this not wait for bootstrapping to finish?

Stebalien (changes requested):

discovery.go
aschmahmann: refactored a bit hopefully this is a bit better now. lmk

discovery.go
aschmahmann: done, although didn't fix all existing linting errors.

discovery.go
aschmahmann: in the Bootstrap function we wait for done before we try again to see if we're bootstrapped.

discovery.go
aschmahmann: refactored

discovery.go
aschmahmann: refactored, is now `MinTopicSize` option. Current docs for `MinTopicSize`:
```
// MinTopicSize returns a function that checks if a router is ready for publishing based on the topic size.
// The router ultimately decides the whether it is ready or not, the given size is just a suggestion.
```

discovery.go
```diff
@@ -0,0 +1,295 @@
...
+		return
+	}
+
+	d.p = p
+	go d.discoverLoop()
+	go d.pollTimer()
+}
+
+func (d *discover) pollTimer() {
+	time.Sleep(DiscoveryPollInitialDelay)
```
Stebalien: Nit: this should probably select on the pubsub context.

discovery.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for {
+		// Check if ready for publishing
+		bootstrapped := make(chan bool, 1)
+		select {
+		case d.p.eval <- func() {
+			done, _ := ready(d.p.rt, topic)
+			bootstrapped <- done
+			close(bootstrapped)
```
Stebalien: nit: no need to close this.

discovery.go
```diff
@@ -0,0 +1,295 @@
...
+
+		select {
+		case <-disc.done:
+		case <-d.p.ctx.Done():
+			return false
+		case <-ctx.Done():
+			return false
+		}
+
+		time.Sleep(time.Millisecond * 100)
```
Stebalien: Context.

pubsub.go
```diff
@@ -73,7 +82,10 @@ type PubSub struct {
 	peerDead chan peer.ID
 
 	// The set of topics we are subscribed to
-	myTopics map[string]map[*Subscription]struct{}
+	mySubs map[string]map[*Subscription]struct{}
+
+	// The set of topics we are interested in
+	myTopics map[string]*Topic
```
Stebalien, line 88: `Join` _should_ subscribe from a pubsub standpoint so we can relay pubsub messages on that topic. We should just make `mySubs` and `myTopics` the same thing.

However, I'm fine leaving this for now (but we should probably have a TODO).

pubsub.go
```diff
@@ -668,6 +727,48 @@ func (p *PubSub) publishMessage(from peer.ID, pmsg *pb.Message) {
...
+
+	resp := make(chan *Topic, 1)
+	t.p.addTopic <- &addTopicReq{
+		topic: t,
+		resp:  resp,
+	}
+	returnedTopic := <-resp
+
+	if returnedTopic != t {
+		return returnedTopic, fmt.Errorf("topic already exists")
```
Stebalien: Let's be consistent. We're either going to completely share topic handles and allow multiple services to join the same topic with no errors _or_ we return an error and don't return the topic handle.

I really _don't_ want to return one service's topic handle to a different service.

discovery.go
```diff
@@ -0,0 +1,295 @@
...
+		select {
+		case d.discoverQ <- disc:
+		case <-d.p.ctx.Done():
+			return false
+		case <-ctx.Done():
+			return false
+		}
+
+		select {
+		case <-disc.done:
```
Stebalien, line 269: We're going to spawn a bunch of goroutines. We _really do_ need to listen on peer join/leave events to determine when we're done bootstrapping. At the moment:

1. This is going to fire once the DHT query completes, regardless of whether or not we're done connecting to peers.
2. We're going to hit that 100ms sleep.
3. We're then going to restart the loop, see that we aren't "ready" because we haven't finished connecting yet.
4. We're going to spawn another discovery process.
5. We're going to spawn a bunch of goroutines to connect to the _same_ peers.

---

The right way to do this is:

1. When we want more peers, signal that we want to discover more peers.
2. When we don't need any more peers, signal that we're happy with the number of peers we have.

---

Note: We can probably also find a way to hack in a fix but, at the very least, we can't spawn a bunch of goroutines connecting to the same peers.

Stebalien (changes requested):

discovery.go
aschmahmann: That's fine, I was mostly pattern matching from gossipsub

https://github.com/libp2p/go-libp2p-pubsub/blob/baa91df4dc5f380b8fc1c0b2af18a7229b975831/gossipsub.go#L386-L387.

Should I fix this in gossipsub as well?

discovery.go
aschmahmann: fair enough

discovery.go
aschmahmann: If you'd like, wasn't sure it was worth setting up a timer here but np.

pubsub.go
aschmahmann: My understanding is that PubSub is supposed to be allowed to `Publish` without `Subscribing` (e.g. see the fanout logic in gossipsub which makes no sense otherwise). So we do in fact need to keep these things separate unless we change the PubSub semantics.

From a programming perspective I think it's totally reasonable to have a `subs map[*Subscription]struct{}` field in `Topic`, I just figured it would be easier to read the PR if I left it as it was. Happy to change it if you're game though (just give a 👍)

pubsub.go
aschmahmann: Sure, we're assuming a single/exclusive topic handle here (as described in #198). I'm happy to remove the error (it felt a bit wrong anyway), but do you think it's useful for a service to know that when they `Join` a topic that they weren't the first ones?

Seems like it could potentially come in handy, but certainly isn't necessary.

discovery.go
aschmahmann: Emitting an "I don't need more peers" signal is likely much more invasive then just checking from the outside. However, if we're certain that the "how many peers" functions that we care about is static per PubSub instance then we can use the PeerJoin events instead of `time.Sleep(100 ms)` (although if the number of peers we wait on can change over time this won't work).

> We're going to spawn another discovery process.

This won't be very expensive since it should be backed by a backoff cache of some sort anyhow.

> We're going to spawn a bunch of goroutines to connect to the same peers.

I'm fine replacing this with storing a local cache of queried peers. I only left it this way bc I though when we talked about this previously that you didn't want to store that state locally.

discovery.go
Stebalien: Let's do that in a followup patch.

discovery.go
Stebalien: It's a short timer but `Bootstrap` is synchronous so I'd expect a canceled context to abort immediately.

pubsub.go
Stebalien: Hm. Yeah, that does make sense. I'm a bit concerned that, if you aren't _subscribed_ to a topic, peers might cut the connection because they don't know you're publishing.

I _would_ like to add an option to `Join` for auto-subscribing (but throwing away the results) but let's punt on that.

pubsub.go
Stebalien: (documenting the results of a sync call)

I was suggesting that we remove the _topic_, not the error. The alternative is to make topics multi-subscriber friendly but that would involve a bunch of work to handle conflicting validators, reference counts, etc.

discovery.go
Stebalien: We discussed this on a call. It's going to be fine as-is as long as we remember which peers we're currently dialing and/or have recently dialed.

pubsub.go
aschmahmann: That's fine we can return nil. The tradeoff will be that we'll need to expose a private function that allows us to get another reference to the topic so that we can continue to use the existing `Publish` and `Subscribe` functions.

Sound good? Also, should I put a deprecation warning on `Publish` and `Subscribe`?

pubsub.go
aschmahmann: Makes sense, we'll punt for now. What you're recommending seems like a part of #28.

pubsub.go
Stebalien: SGTM and yes, we should deprecate (IMO).

topic.go
```diff
@@ -0,0 +1,239 @@
...
+	done := make(chan struct{}, 1)
+	t.p.eval <- func() {
+		tmap := t.p.topics[t.topic]
+		for p := range tmap {
+			h.evtLog[p] = PeerJoin
+		}
+
+		t.evtHandlerMux.Lock()
+		t.evtHandlers[h] = struct{}{}
+		t.evtHandlerMux.Unlock()
```
vyzo, line 48: this blocks the event loop, so we need to make sure there is not much contention here.

gossipsub.go
```diff
@@ -94,6 +94,35 @@ func (gs *GossipSubRouter) RemovePeer(p peer.ID) {
...
+		if gs.peers[p] == FloodSubID {
+			fsPeers++
+		}
+	}
+
+	// gossipsub peers
+	gsPeers = len(gs.mesh[topic])
+
+	if suggested == 0 {
+		suggested = GossipSubD
```
vyzo: I think `GossipSubDlo` might be better. Also, is there a `Publish` option that supplies the `suggested` count?

aschmahmann: I'm fine changing it, figured that we'd be better off proactively getting more peers between `GossipSubDlo` and `GossipSubD` so that if we ever drop below `Dlo` mesh peers we already have peers available for upgrade.

Yes there is a `Publish` option. Look at `WithReadiness` and `MinTopicSize` as they're used in the discovery tests.

topic.go
aschmahmann: Absolutely, but I don't think there shouldn't be much mutex contention. The only places it's used are:
1) here
2) `topic.sendNotification` (as an RLock) which is only called by functions in the event loop so no contention
3)  `topic.Cancel` (as a Lock that protects a map deletion)

discovery.go
```diff
@@ -190,14 +194,18 @@ func (d *discover) Bootstrap(ctx context.Context, topic string, ready RouterRead
 		return true
 	}
 
+	t := time.NewTimer(time.Hour)
+	if !t.Stop() {
```
Stebalien, line 237: This is the most pessimistic check I've ever seen. I like it :laughing:.

Stebalien (approved):
:rocket:

discovery.go
```diff
@@ -0,0 +1,310 @@
...
+	d.p = p
+	go d.discoverLoop()
+	go d.pollTimer()
+}
+
+func (d *discover) pollTimer() {
+	select {
+	case <-time.After(DiscoveryPollInitialDelay):
+	case <-d.p.ctx.Done():
+	}
```
vyzo, line 110: shouldn't we return here?

discovery.go
vyzo: wait what? this seems completely unnecessary.

discovery.go
```diff
@@ -0,0 +1,310 @@
...
+
+	// done handles completion of a discovery request
+	done chan string
+}
+
+// MinTopicSize returns a function that checks if a router is ready for publishing based on the topic size.
+// The router ultimately decides the whether it is ready or not, the given size is just a suggestion.
+func MinTopicSize(size int) RouterReady {
+	return func(rt PubSubRouter, topic string) (bool, error) {
+		return rt.EnoughPeers(topic, size), nil
```
vyzo, line 77: this needs to be called in the event loop, at least for gossipsub.

vyzo (commented):
This looks good; just a minor bug and and open question about the `MinTopicSize` returend function being called potentially outside the event loop.

discovery.go
aschmahmann: It is likely unnecessary given how long the timer is set for. However, it is basically 0 performance cost for properly using the `Timer.Stop()` idiom as is described in the documentation.

discovery.go
aschmahmann: ya, might as well.

discovery.go
```diff
@@ -0,0 +1,310 @@
...
+	if !t.Stop() {
+		<-t.C
+	}
+
+	for {
+		// Check if ready for publishing
+		bootstrapped := make(chan bool, 1)
+		select {
+		case d.p.eval <- func() {
+			done, _ := ready(d.p.rt, topic)
```
aschmahmann, line 246: @vyzo The only time MinTopicSize is called is right here. As you can it is in the event loop.

discovery.go
aschmahmann: See: https://github.com/libp2p/go-libp2p-pubsub/pull/184/files/45117e8a47af41fd65eaa03ae8b6298adbe96b68#r339047141

discovery.go
vyzo: ok.

discovery.go
```diff
@@ -0,0 +1,310 @@
...
+				continue
+			}
+
+			findPeers[pi.ID] = struct{}{}
+
+			go func(pi peer.AddrInfo) {
+				ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+				defer cancel()
+
+				err := d.p.host.Connect(ctx, pi)
```
aschmahmann: Thanks @Stebalien for the review. In order to solve https://github.com/libp2p/go-libp2p-pubsub/pull/184#discussion_r330846676, we should first get https://github.com/libp2p/go-libp2p-discovery/pull/26 merged. That PR has a `BackoffConnector` object that we should be able to use here instead of only relying on the underlying dialer to backoff (which you mentioned it probably won't completely do).

I can push a commit now that handles the last few things here (@vyzo wanted the minimum number of peers to be lower, I need to merge in the latest changes from master, and I'll link to the commit for the discovery PR and use the backoff connector). Should all be taken care of by the time you wake up 😄

gossipsub.go
aschmahmann: @vyzo going to make some hopefully end-of-the-road changes. Did you still think this was important or do you agree that `GossipSubD` instead of `GossipSubDlo` actually make more sense?

vyzo: I think we should use `GossipSubDlo` as it makes more sense.

discovery.go
vyzo: Let's get https://github.com/libp2p/go-libp2p-discovery/pull/26 merged and incorporate this here.
We are almost done other than that.

gossipsub.go
aschmahmann: np

discovery.go
aschmahmann: done. Also, when you make comments in the future could you comment on the files tab instead of on a particular iteration. For some reason your comments don't show "outdated" after an update.

subscription.go
```diff
@@ -2,35 +2,19 @@ package pubsub
...
-}
-
-type PeerEvent struct {
-	Type EventType
-	Peer peer.ID
+	err      error
 }
 
+// Topic returns the topic string associated with the Subscription
 func (sub *Subscription) Topic() string {
```
aschmahmann, line 18: @vyzo no need to do this now, but wanted to point out that it might be worth deprecating this function and/or replacing it with one that returns a `Topic` instead of a string.

subscription.go
```diff
@@ -49,6 +33,8 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
 	}
 }
 
+// Cancel closes the subscription. If this is the last active subscription then pubsub will send an unsubscribe
+// announcement to the network.
 func (sub *Subscription) Cancel() {
```
aschmahmann, line 38: @vyzo this comment work for you?

pubsub.go
```diff
@@ -203,6 +222,9 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	h.Network().Notify((*PubSubNotif)(ps))
 
 	ps.val.Start(ps)
+	if err := ps.disc.Start(ps); err != nil {
+		return nil, err
+	}
```
aschmahmann: @vyzo I just changed the `Start` function to return an error since creating a cache can throw an error. In practice this should never happen though.

Should I move things around here (e.g. move `ps.disc.Start` at least before `val.Start` if not earlier so that any other goroutines also get cancelled?

discovery.go
```diff
@@ -0,0 +1,301 @@
...
+	if d.discovery == nil || p == nil {
+		return nil
+	}
+
+	d.p = p
+
+	minBackoff, maxBackoff := time.Second*10, time.Hour
+	rng := rand.New(rand.NewSource(rand.Int63()))
+	backoff := discimpl.NewExponentialBackoff(minBackoff, maxBackoff, discimpl.FullJitter, time.Second, 5.0, 0, rng)
+	conn, err := discimpl.NewBackoffConnector(p.host, 100, time.Second*10, backoff)
```
aschmahmann: @vyzo @Stebalien Any suggestions on numbers here or if I should expose `WithDiscoveryDialingBackoff` as a PubSub option?

Stebalien: 1. We _should_ expose it.
2. I'm not sure what this timeout is (exponential, fixed, etc?). Given the amount of time/effort it usually takes to dial, I'd go for a default 2 minute backoff.

aschmahmann: There's the timeout which is how long we'll try to dial someone before calling it quits, and there's the backoff which is how long we should wait before dialing at all. Timeout is fixed and I don't really see a good reason to change that now and the backoff follows some function (in this case exponential + some randomness).

The 10 second fixed timeout number was grabbed from https://github.com/libp2p/go-libp2p-pubsub-router/blob/b47a5e34c4bd854836440a59faddab595499592e/pubsub.go#L539 which is how we've been doing bootstrapping thus far. No problem changing this to 2 minutes though if you think that's more reasonable.

Will expose it, but how do the exponential numbers look for backoff? The current numbers for backoffs after attempt N look like:

1: 10 seconds
2: Range [10, 25] seconds
3: [10, 125] seconds
4: [10 seconds, ~10.4 minutes]
5: [10 seconds, ~52 minutes]
6+: [10 seconds, 1 hour]

pubsub.go
aschmahmann: I moved it up.

discovery.go
aschmahmann: Done. We're going to try these numbers for now. They are externally configurable and we can also always change the defaults later.

discovery.go
aschmahmann: Yep, looks like we're at the end. @Stebalien let's make sure to merge the other PR first so that this one can be updated to use the latest commit.

Stebalien (approved):

Stebalien:
The history is nasty enough we should probably rebase and squash (or at least rebase to get rid of the merges).

Stebalien (approved):

Stebalien:
:rocket: 
:clap::clap::clap:

bitcard:
mark

PR #185, opened 2019-05-26
migrate to consolidated types.

raulk:
Applies the core refactor to this module (https://github.com/libp2p/go-libp2p/issues/602).


raulk:
Wow, the pubsub tests are slow. Maybe we can `t.Parallel()`ize?

PR #189, opened 2019-06-06
make message large message handling robust and noisy

Stebalien:
1. Reject outgoing messages that are too large.
2. Drop outgoing RPCs that are too large and _log_.
3. Increase the max RPC size to 1MiB+64KiB (to allow for 1MiB messages).

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -70,11 +70,20 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			continue
 		}
 
+		if out.Size() > maxRPCSize {
+			log.Errorf(
+				"dropping RPC outbound message that's too large: %d > %d",
+				out.Size(),
+				maxRPCSize,
+			)
+			continue
+		}
+
 		select {
 		case mch <- out:
 		default:
-			log.Infof("dropping message to peer %s: queue full", pid)
 			// Drop it. The peer is too slow.
+			log.Infof("dropping message to peer %s: queue full", pid)
 		}
 	}
 }
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -358,7 +358,7 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
 	}
 
 	// push control messages that need to be retried
-	ctl := out.GetControl()
+	ctl = out.GetControl()
 	if ctl != nil {
 		gs.pushControl(p, ctl)
 	}
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -342,15 +342,25 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
 		return
 	}
 
-	select {
-	case mch <- out:
-	default:
-		log.Infof("dropping message to peer %s: queue full", p)
-		// push control messages that need to be retried
-		ctl := out.GetControl()
-		if ctl != nil {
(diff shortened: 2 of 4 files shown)


vyzo:
travis fails with:
```
./gossipsub.go:361:6: no new variables on left side of :=
```

pubsub.go
```diff
@@ -696,7 +703,12 @@ func (p *PubSub) GetTopics() []string {
 }
 
 // Publish publishes data to the given topic.
+//
+// The message data must be less than the maximum message size, 1MiB.
 func (p *PubSub) Publish(topic string, data []byte) error {
+	if len(data) > 0 {
```
vyzo: BUG: that should be
```
if len(data) > maxMessageSize
```

pubsub.go
```diff
@@ -25,6 +25,13 @@ var (
 	TimeCacheDuration = 120 * time.Second
 )
 
+const (
+	// maxRPCSize is the maximum size of any RPC protobuf.
+	maxRPCSize = maxMessageSize + (64 * 1024) // a message + 64 KiB.
+	/// maxMessageSize is the maximum size of outbound message.
+	maxMessageSize = 1 << 20 // 1MiB
```
vyzo, line 30: probably invert the order of these two declarations.

vyzo (changes requested):
there is a bug...

floodsub.go
```diff
@@ -70,11 +70,20 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			continue
 		}
 
+		if out.Size() > maxRPCSize {
```
vyzo, line 73: Is this check really necessary here?
We already drop large messages on publish and can't read messages larger than maxRCPSize anyway.

gossipsub.go
```diff
@@ -342,15 +342,25 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
 
-	select {
-	case mch <- out:
-	default:
-		log.Infof("dropping message to peer %s: queue full", p)
-		// push control messages that need to be retried
-		ctl := out.GetControl()
-		if ctl != nil {
-			gs.pushControl(p, ctl)
+	if out.Size() > maxRPCSize {
```
vyzo, line 345: here too, this check seems unnecessary.

floodsub.go
Stebalien: I guess it should be fine in floodsub. My worry is gossipsub.

gossipsub.go
Stebalien: I'm worried we could add too much gossip. That's why I have that check. But I can remove it if you think it's overkill.

vyzo: ~but the check is done before we even add the gossip!~

floodsub.go
vyzo: well, let's remove it from here as it's redundant, and rethink the gossipsub case.

gossipsub.go
vyzo: ok, this check is reasonable on second thought, if we are adding too much gossip (over 64KB) we probably want to know.

floodsub.go
Stebalien: This can't hurt and future proofs us against, e.g., batching too many messages in a single RPC and then silently failing when the remote side drops it.

PR #190, opened 2019-06-07
Add the ability to handle newly subscribed peers

aschmahmann:
Per conversations with @vyzo @raulk and @Stebalien we are refactoring the content of #171 which does 2 things.

1. Gives developers control over what messages are broadcast
2. Gives developers an "OnJoin" method that can specify an action to do when first connecting to a peer (e.g. send a specific message, utilize an external protocol, etc.)

Instead of doing this by making routers extensible, we will be leaving router extensibility for a later date and solve the issues above by:

1. Using stateful validators + application layer publishing to effectively control which messages are broadcast. This can happen outside of the `go-libp2p-pubsub` package. 
2. Adding a mechanism for PubSub to expose when it connects to a new peer

This PR deals with the latter.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -454,19 +454,11 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
 	}
 
 	tmap := p.topics[sub.topic]
-	inboundBufSize := len(tmap)
-	if inboundBufSize < 32 {
-		inboundBufSize = 32
-	}
 
-	sub.ch = make(chan *Message, 32)
-	sub.joinCh = make(chan peer.ID, inboundBufSize)
-	sub.leaveCh = make(chan peer.ID, 32)
+	for p := range tmap {
+		sub.evtBacklog[p] = PEER_JOIN
+	}
 	sub.cancelCh = p.cancelCh
-
-	for pid := range tmap {
-		sub.joinCh <- pid
-	}
 
 	p.myTopics[sub.topic][sub] = struct{}{}
 
@@ -581,11 +573,7 @@ func (p *PubSub) subscribedToMsg(msg *pb.Message) bool {
 func (p *PubSub) notifyLeave(topic string, pid peer.ID) {
 	if subs, ok := p.myTopics[topic]; ok {
 		for s := range subs {
-			select {
-			case s.leaveCh <- pid:
-			default:
-				log.Infof("Can't deliver leave event to subscription for topic %s; subscriber too slow", topic)
-			}
+			s.sendNotification(PeerEvent{PEER_LEAVE, pid})
 		}
 	}
 }
@@ -605,11 +593,7 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 				if subs, ok := p.myTopics[t]; ok {
 					peer := rpc.from
 					for s := range subs {
-						select {
-						case s.joinCh <- peer:
-						default:
-							log.Infof("Can't deliver join event to subscription for topic %s; subscriber too slow", t)
-						}
+						s.sendNotification(PeerEvent{PEER_JOIN, peer})
 					}
 				}
 			}
@@ -712,6 +696,11 @@ func (p *PubSub) SubscribeByTopicDescriptor(td *pb.TopicDescriptor, opts ...SubO
 
 	sub := &Subscription{
 		topic: td.GetName(),
+
+		ch:         make(chan *Message, 32),
+		peerEvtCh:  make(chan PeerEvent, 32),
+		evtBacklog: make(map[peer.ID]EventType),
+		backlogCh:  make(chan PeerEvent, 1),
 	}
 
 	for _, opt := range opts {
--- a/pubsub.go
+++ b/pubsub.go
@@ -78,6 +78,9 @@ type PubSub struct {
 	// topics tracks which topics each of our peers are subscribed to
 	topics map[string]map[peer.ID]struct{}
 
+	// a set of notification channels for newly subscribed peers
+	newSubs map[string]chan peer.ID
+
 	// sendMsg handles messages that have been validated
 	sendMsg chan *sendReq
 
@@ -418,6 +421,7 @@ func (p *PubSub) handleRemoveSubscription(sub *Subscription) {
 
 	sub.err = fmt.Errorf("subscription cancelled by calling sub.Cancel()")
 	close(sub.ch)
+	close(sub.inboundSubs)
 	delete(subs, sub)
 
 	if len(subs) == 0 {
@@ -447,6 +451,7 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
 		subs = p.myTopics[sub.topic]
 	}
 
+	sub.inboundSubs = make(chan peer.ID, 32)
 	sub.ch = make(chan *Message, 32)
 	sub.cancelCh = p.cancelCh
 
@@ -570,7 +575,19 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 				p.topics[t] = tmap
 			}
 
-			tmap[rpc.from] = struct{}{}
+			if _, ok = tmap[rpc.from]; !ok {
+				tmap[rpc.from] = struct{}{}
+				if subs, ok := p.myTopics[t]; ok {
+					inboundPeer := rpc.from
+					for s := range subs {
+						select {
+						case s.inboundSubs <- inboundPeer:
+						default:
+							log.Infof("Can't deliver event to subscription for topic %s; subscriber too slow", t)
+						}
+					}
+				}
+			}
 		} else {
 			tmap, ok := p.topics[t]
 			if !ok {
--- a/pubsub.go
+++ b/pubsub.go
@@ -78,9 +78,6 @@ type PubSub struct {
 	// topics tracks which topics each of our peers are subscribed to
 	topics map[string]map[peer.ID]struct{}
 
-	// a set of notification channels for newly subscribed peers
-	newSubs map[string]chan peer.ID
-
 	// sendMsg handles messages that have been validated
 	sendMsg chan *sendReq
 
@@ -426,9 +423,7 @@ func (p *PubSub) handleRemoveSubscription(sub *Subscription) {
 	}
 
 	sub.err = fmt.Errorf("subscription cancelled by calling sub.Cancel()")
-	close(sub.ch)
-	close(sub.joinCh)
-	close(sub.leaveCh)
+	sub.close()
 	delete(subs, sub)
 
 	if len(subs) == 0 {
@@ -608,10 +603,10 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 			if _, ok = tmap[rpc.from]; !ok {
 				tmap[rpc.from] = struct{}{}
 				if subs, ok := p.myTopics[t]; ok {
-					inboundPeer := rpc.from
+					peer := rpc.from
 					for s := range subs {
 						select {
-						case s.joinCh <- inboundPeer:
+						case s.joinCh <- peer:
 						default:
 							log.Infof("Can't deliver join event to subscription for topic %s; subscriber too slow", t)
 						}
--- a/pubsub.go
+++ b/pubsub.go
@@ -456,7 +456,7 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
 	tmap := p.topics[sub.topic]
 
 	for p := range tmap {
-		sub.evtBacklog[p] = PEER_JOIN
+		sub.evtBacklog[p] = PeerJoin
 	}
 	sub.cancelCh = p.cancelCh
 
@@ -573,7 +573,7 @@ func (p *PubSub) subscribedToMsg(msg *pb.Message) bool {
 func (p *PubSub) notifyLeave(topic string, pid peer.ID) {
 	if subs, ok := p.myTopics[topic]; ok {
 		for s := range subs {
-			s.sendNotification(PeerEvent{PEER_LEAVE, pid})
+			s.sendNotification(PeerEvent{PeerLeave, pid})
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


vyzo (commented):
This looks good. Shouldn't we also notify when peers _leave_ the topic however?

subscription.go
```diff
@@ -31,3 +33,16 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
 func (sub *Subscription) Cancel() {
 	sub.cancelCh <- sub
 }
+
+func (sub *Subscription) NextSubscribedPeer(ctx context.Context) (peer.ID, error) {
```
vyzo: maybe call this `NextPeer`.

vyzo:
An issue here is that if we subscribe _after_ peers have been discovered, there will be no notifications.
I think we might want to emit notifications for all known peers when we first subscribe.

subscription.go
aschmahmann: Sure and what would you like the next leave event to be called? I thought `NextPeerJoin` and `NextPeerLeave` might have a symmetry, but your call.

aschmahmann:
While you could already get the existing peers with the `FindPeers` command, I think you're right that this would likely be useful for many developers. I'll likely just make the buffered channel large enough to fill with the initial peers and fill it up on initialization.

aschmahmann:
> This looks good. Shouldn't we also notify when peers _leave_ the topic however?

@vyzo That's fine by me. Any thoughts on how important receiving all the leave messages is? In particular, while joins might be slow enough that concurrent joins will not get dropped due to channel buffer overflow disconnecting from the internet might instead lead to mass connectivity dropping and therefore some missed `NextPeerLeave` events.

Also, wanted to confirm that blacklisting based disconnects should count towards `Leave` events also.

pubsub.go
```diff
@@ -333,8 +336,9 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			}
 
 			delete(p.peers, pid)
-			for _, t := range p.topics {
-				delete(t, pid)
+			for t, tmap := range p.topics {
+				delete(tmap, pid)
+				p.notifySubscriberLeft(t, pid)
```
vyzo: this is going to emit phantom notifications, as it might not be in the map.

pubsub.go
```diff
@@ -392,8 +396,9 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			if ok {
 				close(ch)
 				delete(p.peers, pid)
-				for _, t := range p.topics {
-					delete(t, pid)
+				for t, tmap := range p.topics {
+					delete(tmap, pid)
+					p.notifySubscriberLeft(t, pid)
```
vyzo: same here.

pubsub.go
```diff
@@ -560,6 +579,18 @@ func (p *PubSub) subscribedToMsg(msg *pb.Message) bool {
 	return false
 }
 
+func (p *PubSub) notifySubscriberLeft(topic string, pid peer.ID) {
```
vyzo: Can we call this something else? This is a mouthful, call it `notifyLeave`.

subscription.go
```diff
@@ -2,13 +2,16 @@ package pubsub
...
 type Subscription struct {
-	topic    string
-	ch       chan *Message
-	cancelCh chan<- *Subscription
-	err      error
+	topic       string
+	ch          chan *Message
+	cancelCh    chan<- *Subscription
+	inboundSubs chan peer.ID
+	leavingSubs chan peer.ID
```
vyzo: naming again, call these two `joinCh` and `leaveCh`.

vyzo (commented):
There is a race to read join and leave notifications, as they are emitted in different channels.
I think we need a single channel for both notifications and use a struct with a field that specifies the notification type.

pubsub.go
aschmahmann: oops, pretty silly mistake there. Thanks

pubsub.go
```diff
@@ -418,6 +427,8 @@ func (p *PubSub) handleRemoveSubscription(sub *Subscription) {
 
 	sub.err = fmt.Errorf("subscription cancelled by calling sub.Cancel()")
 	close(sub.ch)
+	close(sub.joinCh)
+	close(sub.leaveCh)
```
vyzo: let's move these three closures into a `close` method in the subscription.

pubsub.go
```diff
@@ -78,6 +78,9 @@ type PubSub struct {
 	// topics tracks which topics each of our peers are subscribed to
 	topics map[string]map[peer.ID]struct{}
 
+	// a set of notification channels for newly subscribed peers
+	newSubs map[string]chan peer.ID
```
vyzo: is this actually used?

pubsub.go
```diff
@@ -570,13 +605,29 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 				p.topics[t] = tmap
 			}
 
-			tmap[rpc.from] = struct{}{}
+			if _, ok = tmap[rpc.from]; !ok {
+				tmap[rpc.from] = struct{}{}
+				if subs, ok := p.myTopics[t]; ok {
+					inboundPeer := rpc.from
```
vyzo: call this just `peer`.

subscription.go
```diff
@@ -2,19 +2,36 @@ package pubsub
 
 import (
 	"context"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type EventType uint8
```
vyzo: just `int`.

subscription.go
```diff
@@ -2,19 +2,36 @@ package pubsub
...
+
+const (
+	UNKNOWN EventType = iota
```
vyzo: this is unnecessary.

vyzo (commented):
this is looking good, just a few minor points.

pubsub.go
aschmahmann: ok, private method right?

pubsub.go
aschmahmann: oops, didn't think about it bc I thought Go would flag the unused variables. My bad.

pubsub.go
aschmahmann: sure, didn't want to obscure the `peer` pkg, but that's fine

subscription.go
aschmahmann: sure, any particular reason why so I know for future enums?

subscription.go
aschmahmann: My understanding is that this is best practice so that an empty `EventType` doesn't have a valid value by accident. Is that incorrect/incomplete?

vyzo: it's not, it's just redundant.

vyzo (commented):
only thing remaining to fix is the enum, other than that it looks good.

subscription.go
```diff
@@ -31,3 +47,33 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+
+func (sub *Subscription) close() {
+	close(sub.ch)
+	close(sub.joinCh)
+	close(sub.leaveCh)
+}
+
+// NextPeerEvent returns the next event regarding subscribed peers
+// Note: There is no guarantee that the Peer Join event will fire before
+// the related Peer Leave event for a given peer
```
Stebalien: Should note that this is lossy.

Stebalien: Actually, this doesn't even need to be lossy. We can have this function:

1. Return all connected peers on start.
2. Guarantee that it returns all connect/disconnect events.

We can do this by keeping track of peers we've reported as subscribed internally. If we fall behind, pubsub can close the event channel. If we see that the event channel has been closed, we can re-register for events and, at the same time, diff our local list of tracked peers against the master list.

---

Also, we really should guarantee ordering of these events. Is that not possible?

aschmahmann: Good idea. The reason we couldn't guarantee ordering of events was because we didn't want the `PeerLeave` events filling up the event queue to block `PeerJoin` events from making it in so we had to process them independently before we could combine them at the output.

However, your scheme should allow us to do both and the only tradeoff is that we won't see peers that quickly Join and Leave, but that feels pretty irrelevant.

vyzo: sounds good to me!

aschmahmann: @vyzo @Stebalien did this, need an updated review. I'd like more eyes to confirm my strategy isn't racy, although I added some tests to try and help a bit with that.

subscription.go
```diff
@@ -31,3 +51,81 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+	for k, v := range sub.evtBacklog {
+		evt := PeerEvent{Peer: k, Type: v}
+		delete(sub.evtBacklog, k)
+		return evt, true
+	}
+	return PeerEvent{}, false
+}
+
+// NextPeerEvent returns the next event regarding subscribed peers
+// Guarantees: Peer Join and Peer Leave events for a given peer will fire in order.
```
aschmahmann: Does the comment here (and the one on the next line) fit the bill?

subscription.go
```diff
@@ -31,3 +51,81 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+// Unless a peer both Joins and Leaves before NextPeerEvent emits either event
+// all events will eventually be received from NextPeerEvent.
```
aschmahmann, line 94: Does the comment here fit the bill?

aschmahmann:
Latest idea for how to do this:

We effectively want an infinite length buffer so that no Joins/Leaves are lost. However, we can be slightly efficient by ignoring when the same peer Joins and Leaves before we even ask for a notification.

Strategy:
Have a channel that receives updates, if the channel is full empty it into the buffer which is a map protected by a mutex.
When asked for an update we pull from the buffer if it's not empty, otherwise we pull from the channel.

There are some side points around how to get this right and look clean that I could definitely use some review on.

Stebalien:
> Have a channel that receives updates, if the channel is full empty it into the buffer which is a map protected by a mutex.
> When asked for an update we pull from the buffer if it's not empty, otherwise we pull from the channel.

Maybe? I'm concerned about races reading the channel. _Someone_ needs to drain the channel and we can't have both the subscriber and pubsub do this (or risk out of order events).

If we want to go with this approach, we could just put everything into the map up front (and use a channel for signaling).

subscription.go
```diff
@@ -31,3 +51,83 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+	sub.eventMx.Lock()
+	evt, ok := sub.pullFromBacklog()
+	sub.eventMx.Unlock()
+
+	if ok {
+		return evt, nil
+	}
+
+	select {
+	case evt, ok := <-sub.peerEvtCh:
```
Stebalien: This is racy. We could be in the middle of draining this queue into the map and read an leave event before the join (now in the map).

subscription.go
```diff
@@ -31,3 +52,65 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+
+func (sub *Subscription) sendNotification(evt PeerEvent) {
+	sub.backlogMx.Lock()
+	defer sub.backlogMx.Unlock()
+
+	sub.addToBacklog(evt)
+
+	select {
+	case sub.backlogCh <- struct{}{}:
+	default:
```
Stebalien: uber nit: I'd just move this to the `addToBacklog` function. We shouldn't call it  in the `else` case anyways.

subscription.go
```diff
@@ -31,3 +52,65 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+	}
+	return PeerEvent{}, false
+}
+
+// NextPeerEvent returns the next event regarding subscribed peers
+// Guarantees: Peer Join and Peer Leave events for a given peer will fire in order.
+// Unless a peer both Joins and Leaves before NextPeerEvent emits either event
+// all events will eventually be received from NextPeerEvent.
+func (sub *Subscription) NextPeerEvent(ctx context.Context) (PeerEvent, error) {
+	sub.nextEventMx.Lock()
```
Stebalien: Ok, so, I was wrong. This has a problem:

1. Thread 1: Call `NextPeerEvent(ctxa)`
2. Thread 1: Take lock.
3. Thread 2: Call `NextPeerEvent(ctxb)`
4. Thread 3: Cancel `ctxb`.

I think we need to go the loop route where any thread taking from the backlog puts the flag back in the backlog.

Stebalien (changes requested):
(sorry, we missed a case)

subscription.go
aschmahmann: which `else` case? 

I left them separate mostly as a strategy for separating out the code components so we could play around with a few options. However, given that `addToBacklog` is only called once, and it's by `sendNotification` if we wanted to we could just combine these functions. Having them separate is only useful if it helps with code readability.

subscription.go
aschmahmann: right, if you cancel because we're using mutexes and not channels nothing's going to happen. I guess we can always keep putting the flag in.

However, it does mean that if for some reason a consumer spun up two goroutines to read events then a Join and a Leave could come out of order. Do we care about this case?

Stebalien: No, that's their problem. I mean, we could just say "only do this on one thread" but... I don't like weird contracts like that.

subscription.go
Stebalien: https://github.com/libp2p/go-libp2p-pubsub/pull/190/files#diff-384bf0d04fad588e8589bdda0252018bR78

subscription.go
```diff
@@ -31,3 +51,69 @@ func (sub *Subscription) Next(ctx context.Context) (*Message, error) {
...
+		sub.evtLogMx.Lock()
+		evt, ok := sub.pullFromEventLog()
+		if ok {
+			// make sure an event log signal is available if there are events in the event log
+			if len(sub.evtLog) > 0 {
+				select {
+				case sub.evtLogCh <- struct{}{}:
+				default:
+				}
+			}
```
aschmahmann, line 106: @Stebalien we thinking this works? I should only need to push a signal if there are events in the log right?

Stebalien: LGTM.

Stebalien (approved):
:shipit:

vyzo (approved):
alright, let's merge this.

PR #196, opened 2019-08-13
minor bug fix: on join, source peers from gossip[topic] if insufficient peers in fanout[topic]

### LINKED ISSUE #194: Not filling mesh peers if there are less than `D` of them when joining a topic
### What is wrong?
In `GossipSubRouter.Join()` if there are fanout peers but not enough of them, it won't `selects the remaining number of peers (D-x) from peers.gossipsub[topic]` as suggested in the [spec](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub#topic-membership).
https://github.com/libp2p/go-libp2p-pubsub/blob/4221a3966bf2bf3f915fcd6623c6bde688d99556/gossipsub.go#L269-L274

vyzo:
Closes #194 

This is a very minor bug, as the mesh would get filled in the next heartbeat.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -268,6 +268,17 @@ func (gs *GossipSubRouter) Join(topic string) {
 
 	gmap, ok = gs.fanout[topic]
 	if ok {
+		if len(gmap) < GossipSubD {
+			// we need more peers; eager, as this would get fixed in the next heartbeat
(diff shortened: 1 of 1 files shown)


raulk (approved):
nit: would be good to add a test for this.

PR #199, opened 2019-09-23
Refactor to allow out-of-tree router implementations

vyzo:
By popular demand, this refactors the pubsub system to allow out-of-tree router implementations.
This is accomplished by introducing a `PubSubControl` interface, which is the pubsub view presented to routers. 
It is currently sufficient to implement our in-tree routers, but we may want to extend it in the future.
It also adds a new field to the protobuf, `extData`, to be used by out-of-tree protocols for their own purposes.


pubsub.go
```diff
@@ -133,6 +134,35 @@ type PubSubRouter interface {
...
+	// If the peer was unknown, it returns ErrUnknownPeer.
+	SendMessage(p peer.ID, msg *RPC) error
+
+	// SeenMessage returns true if the message identified by mid has been seen before
+	// by the pubsub system.
+	SeenMessage(mid string) bool
+
+	// Eval schedules a thunk for evaluation in the pubsub process loop.
+	// Returns false if the evaluation was aborted because the pubsub system has been shutdown.
+	Eval(thunk func()) bool
```
Stebalien, line 157: This is funky.

* Are we allowed to use the rest of the interface outside of a "thunk"?
* Are we using this to protect gossipsub state? If so, what if we just used a lock?

pb/rpc.proto
```diff
@@ -12,6 +12,7 @@ message RPC {
 	}
 
 	optional ControlMessage control = 3;
+    optional bytes extData = 4; // protocol extension data
```
Stebalien: gofmt.

Stebalien: https://developers.google.com/protocol-buffers/docs/proto3#any?

Stebalien (changes requested):
:heart:

I believe we need to export `from` from `pubsub.RPC`.

pb/rpc.proto
vyzo: oops! emacs did this.

pubsub.go
vyzo: The interface in general is unsafe to use outside of the event loop, this should probably be specified in the documentation; will fix.

vyzo:
Good point, the `from` field will have to be exported.

pb/rpc.proto
Stebalien:
```el
(use-package go-mode
  :config (gofmt-before-save))
```

:wink:

vyzo: Any is probably a bad idea because it needs metadata, which may not be supported by some programming languages.
Also, our protobuf is proto2 so we don't want to pollute with proto3 constructs.

pubsub.go
Stebalien: Got it. I wonder if we should put _that_ part into a separate interface and pass it to the thunk? That should make it clear that that's the only safe way to use it.

Stebalien: Does `Eval` _wait_ for the thunk to eval?

pb/rpc.proto
vyzo: I do have a gofmt save hook, but this is a protobuf file!

pubsub.go
vyzo: It waits for the evaluation to start, but it doesn't wait for completion.

pb/rpc.proto
Stebalien: :roll:

So, I'm just cheating and using spacemacs. `(add-to-list 'protobuf-mode-hooks (lambda () (setq-local indent-tabs-mode t)))`?

pubsub.go
vyzo: Marked the unsafe outside the event loop methods as such.

pb/rpc.proto
vyzo: fixed the indentation; the hook looks good, i might adopt it :p

pubsub.go
Stebalien: I'd still like to consider moving these to a separate interface and passing it into the thunk. I _really_ wish go had some kind of goroutine local storage so we could just make all this automatic but it doesn't.

pubsub.go
```diff
@@ -817,3 +850,48 @@ func (p *PubSub) UnregisterTopicValidator(topic string) error {
 	p.rmVal <- rmVal
 	return <-rmVal.resp
 }
+
+// PubSubControl view
+var ErrUnknownPeer = errors.New("unknown peer")
+var ErrQueueFull = errors.New("queue full")
+
+func (p *PubSubControlView) PeersForTopic(topic string) (map[peer.ID]struct{}, bool) {
+	tmap, ok := p.topics[topic]
```
Stebalien, line 859: Copying the peer list into a slice shouldn't really cost us anything in the grand scheme of things. I'd do that to avoid exposing internal state.

pubsub.go
vyzo: Well, the interface gets attached in `Router.Attach` and can get used without explicitly re-entering with `Eval` -- all callbacks to the router from the pubsub system are inside the event loop and that would be awkward to use.
I think the contract should be "you call the pubsub system inside the event loop; if you have a goroutine that is outside (eg a timer, like in gossipsub), use Eval to re-enter the event loop."
We should probably document this in the `Router` interface.

Stebalien: Discussed synchronously: All the router callbacks are _also_ called on the main thread so we'd have to feed this new interface in there as well. TL;DR: not worth it.

Instead, we should:

1. Carefully document this.
2. Maybe call this `InvokeInMainThread` (or something less java-y).

pubsub.go
```diff
@@ -133,6 +134,37 @@ type PubSubRouter interface {
 	Leave(topic string)
 }
 
+// PubSubControl is the control interface presented by the pubsub system to routers.
+type PubSubControl interface {
+	// PeersForTopic returns the set of peers in a topic; the boolean indicator will be false
+	// if the topic is unknown.
+	// This method is not safe to call outside the pubsub process loop.
+	PeersForTopic(topic string) (map[peer.ID]struct{}, bool)
```
raulk, line 142: Hm, would a router want to limit their view of topic membership by only seeing peers that support their routing protocol? The reason I ask is because all other methods can be regarded as _events_, except for this one, which is pure state access. It seems to break the abstraction a bit.

I'd be more inclined to have the router and control layers be event-based, with no shared state. That could remove the need for the synthetic `Eval()` method -- which is only needed to protect state at an adjacent layer...

I see this model working as follows:

1. On startup, routers enrol with the control layer, providing a predicate that, given a `[]protocol.ID`, returns whether the router can interact with that peer or not.
2. On every new peer that advertises the umbrella pubsub protocol, we call the predicate on all routers, and emit a `NewPeer` event to routers that signal support.
3. The control layer tracks which routers are handling which peers. When a peer goes away, we emit an event relevant routers.
4. For topic membership control messages (join or leave), the control layer emit events to the supporting routers.
5. Routers send messages by pushing events on a channel that the control layer is looping over.
6. ...

This approach would insulate the state of each layer and we'd remove the need for cross-boundary scope changes.

pubsub.go
raulk: I think this is a smell. Alternative design proposal in comments above.

pubsub.go
raulk: I'm assuming this method is on the hot path, hence the intention to avoid copying by introducing the thunk scheduling to guard state...

raulk (changes requested):
TL;DR: I think the `Eval()` is a smell. Its goal is to curtail state leakage caused the preexisting design. An event-based model would be cleaner and more comfortable to develop against.

pubsub.go
vyzo: this is a major redesign i'd rather not tackle in this pr!

vyzo: Also, programming this with events makes it a lot harder to reason about; I really like the simplicity of a lock-free event loop.

pubsub.go
vyzo: it is on the hot path indeed.

pubsub.go
raulk: Hmmm, I get that, but pubsub/gossipsub is relied on by major downstream adopters, and I'm happy to invest time and effort in that refactor. I think it would make the code more decoupled, elegant, accessible for contributors, and less spaghetti.

Isn't the lock-free event loop inherently mono-threaded, anyway? It can become a major bottleneck.

If instead we let routers keep their own state locally, and be responsible for mutating in reaction to events they receive from the control layer, I think it makes the system more functional and a lot easier to reason about.

aschmahmann: 👍 that having an `Eval()` function on the interface seems ridiculous (also, this being ridiculous was the reason given for #171 not being it's own router). If there's a short term need for users that don't want to fork go-libp2p-pubsub just to get their work done I'd rather be even grosser and just export the fields in PubSub so people know how unstable/temporary the state is.

A more approachable, and less intrusive, approach might be to invert a small amount of the design. For example `PeersForTopic` could easily be part of the `PubSubRouter` interface instead utilizing a `PubSubControl` interface. If certain structs/functions are really the same for various different routers then we could have some `PubSubRouterBase` struct that could be embedded in the individual routers.

vyzo:
closing as we decided to refactor internals before allowing out of tree routers.

Issue #200, opened 2019-09-30
Subscription.Cancel() hangs forever if PubSub exited


lukesolo:
Function Cancel() blocks forever because no one listens to sub.cancelCh if PubSub has exited.

```golang
func (sub *Subscription) Cancel() {
	sub.cancelCh <- sub
}
```

How to reproduce:

```golang
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/libp2p/go-libp2p"
	pubsub "github.com/libp2p/go-libp2p-pubsub"
)

func main() {
	ctx, cancel := context.WithCancel(context.Background())

	host, _ := libp2p.New(ctx)
	gsub, _ := pubsub.NewGossipSub(ctx, host)
	sub, _ := gsub.Subscribe("topic")

	cancel()
	time.Sleep(time.Second)

	fmt.Println("cancelling subscription")
	sub.Cancel() // will hang forever
	fmt.Println("subscription cancelled")
}
```

I think that Subscription should have the same context as a GossipSub, so that you could check it in the Cancel function

```golang
func (sub *Subscription) Cancel() {
	select {
		case sub.cancelCh <- sub:
		case <-sub.ctx.Done():
	}
}
```

vyzo:
Yes indeed, this is a bug; care for a patch?

lukesolo:
@vyzo This is not a critical bug for me (at least now), but I made a pull request with a simple fix, and now I think what to do with the other methods which get context as an argument. Should they check both inner and outer context for cancellation in select statement?

vyzo:
If they are doing an operation with the event loop only, because that's only when they can deadlock on shutdown.

vyzo:
Closing as the problem in the issue has been fixed by #201

PR #201, opened 2019-09-30
add PubSub's context to Subscription

lukesolo:
Fix for #200

This patch fixes the issue, but what about functions which get outer context as an argument, e.g. `Next(ctx context.Context)`. Should these functions check only outer context for cancellation or inner too?


vyzo (approved):
thank you!

Issue #202, opened 2019-10-03
Fix flaky tests   


aarshkshah1992:
Some tests are flaky & do not reliably pass locally or on CI (#141 for example). 
As discussed with @aschmahmann, the two we know to be flaky are:
 
1. TestGossipsubGossip
2. TestGossipsubGossipPiggyback

Please feel free to point out more in the comments.

Also, please can someone assign this issue to me ?

aschmahmann:
TestGossipsubControlPiggyback I suspect is also flaky given that it is skipped with `t.Skip("travis regularly fails on this test")`.

I can try and do some more looking into this, but I suspect that most of the tests that utilize `time.Sleep` are flaky to some degree, but most are not currently causing major issues and will therefore be difficult to detect. If these sleeps are actually the problem then replacing them with either the new PubSub events or polling internal data structures (e.g. something like `for condition{condition=checkPrivateState(); time.Sleep(100ms)}`) should help.

aarshkshah1992:
@vyzo @aschmahmann  @Stebalien 

The first [PR](https://github.com/libp2p/go-libp2p-pubsub/pull/203) makes the flaky test wait for mesh construction & subscription processing rather than sleeping & hoping it has finished. However, that didn't do much. 

The tests fail because Travis has a timeout of 10 minutes per test & not all subscribers receive the message before that. I'm not being able to think of how to dig into this issue. What more should we do/investigate for these tests ?

aschmahmann:
The Travis 10 minute timeout shouldn't be an issue. However, it does look like not all the subscribers are receiving the messages.

My guess is that this is happening because messages are getting dropped here:
https://github.com/libp2p/go-libp2p-pubsub/blob/d28851004cfcbccceea98f8eb409c3c01f882097/gossipsub.go#L345-L348

If we can confirm this is true (this can be a simple as throwing a panic instead of logging an error, or adding some internal variable to gossipsub that we can check) that would be great. At that point we can figure out what to do next that might include:

1) Diving more deeply into the events that cause the messages to drop (maybe there's a bug somewhere to expose here) 
[...]

PR #203, opened 2019-10-04
Try to fix flaky tests by waiting for subscriptions & mesh to be ready

aarshkshah1992:
For #202 

@aschmahmann 

In spite of many repeated runs, I was unable to reproduce the flakiness. However, I can intuitively see that the mesh overlay is not guaranteed to be in a steady state if we just sleep & hope for the best. 

Two main changes made to all the 3 tests mentioned in the issue are:

1. Poll for the incoming subscription messages on all peers to be processed so that `PubSub.topics` reflects the correct state
2. Poll for the overlay mesh to have ATLEAST `Dlow` peers rather than sleeping for 2 seconds & hoping the corresponding heartbeats get processed in the meantime 

Let me know what you think. If this approach sounds reasonable, we can do something similar for the other tests.

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -23,22 +23,68 @@ func getGossipsubs(ctx context.Context, hs []host.Host, opts ...Option) []*PubSu
 	return psubs
 }
 
-func waitForMeshConstruction(topic string, psubs []*PubSub) {
+func waitForMeshConstruction(t *testing.T, parentCtx context.Context, topic string, psubs []*PubSub) {
+	mChan := make(chan int, 1)
+	defer close(mChan)
 	for _, ps := range psubs {
+		// Do not wait for more than 10 seconds  for one peer to catch up
+		ctx, cancel := context.WithTimeout(parentCtx, 10*time.Second)
+		defer cancel()
 		grt := ps.rt.(*GossipSubRouter)
-		for len(grt.mesh[topic]) < GossipSubDlo {
-			time.Sleep(100 * time.Millisecond)
+
+	LOOP:
+		for {
+			select {
+			case <-ctx.Done():
+				t.Fatalf("err: timed out waiting for mesh construction, host: %s", ps.host.ID())
+			case ps.eval <- func() {
+				mChan <- len(grt.mesh[topic])
+			}:
+				select {
+				case <-ctx.Done():
+					t.Fatalf("err: timed out waiting for mesh construction, host: %s", ps.host.ID())
+				case mLen := <-mChan:
+					if mLen < GossipSubDlo {
+						time.Sleep(100 * time.Millisecond)
+					} else {
+						break LOOP
+					}
+				}
+			}
 		}
 	}
 }
 
 // wait for all incoming subscription messages to be processed
 // this method should ONLY be called if all connected peers subscribe to the same topic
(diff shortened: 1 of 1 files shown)


gossipsub_test.go
```diff
@@ -23,6 +23,26 @@ func getGossipsubs(ctx context.Context, hs []host.Host, opts ...Option) []*PubSu
 	return psubs
 }
 
+func waitForMeshConstruction(topic string, psubs []*PubSub) {
```
aarshkshah1992: Races with the heartbeat thread for GossipsubRouter.mesh[T]. Let me think of something.

gossipsub_test.go
```diff
@@ -23,6 +23,26 @@ func getGossipsubs(ctx context.Context, hs []host.Host, opts ...Option) []*PubSu
...
+		grt := ps.rt.(*GossipSubRouter)
+		for len(grt.mesh[topic]) < GossipSubDlo {
+			time.Sleep(100 * time.Millisecond)
+		}
+	}
+}
+
+// wait for all incoming subscription messages to be processed
+// this method should ONLY be called if all connected peers subscribe to the same topic
+func waitForSubscriptionProcessing(topic string, psubs []*PubSub) {
```
aarshkshah1992: Races with PubSub.handleIncomingRPC when it tries to update Pubsub.topics[T].

gossipsub_test.go
aschmahmann: yep, if your function is very computationally inexpensive you can generally use something along the lines of this:

```
        done := make(chan resultType,1)
	p.eval <- func() {
               done <- myFn()
	}
        return <-done
```

Note: the channel closure is only going to be needed if your function needs to be synchronous. Also, you may want to add in `select` statements that check for an expired context if you want to reuse these functions a bunch.

gossipsub_test.go
aschmahmann: See above: PubSub is currently setup to basically have a few long running goroutines with event loops in them instead of using mutexes. There's an internal `p.eval` channel that will run arbitrary functions for you inside of the event loop so you don't run into races.

aarshkshah1992: Aha... that is brilliant. Let me fix this.

gossipsub_test.go
```diff
@@ -704,7 +726,7 @@ func TestGossipsubGraftPruneRetry(t *testing.T) {
 }
 
 func TestGossipsubControlPiggyback(t *testing.T) {
-	t.Skip("travis regularly fails on this test")
+	//t.Skip("travis regularly fails on this test")
```
aarshkshah1992, line 775: @aschmahmann I'm not being able to understand the last line of this test & of `TestGossipsubGossip`

```
// and wait for some gossip flushing
	time.Sleep(time.Second * 2)
```

Please can you explain what this is for ? Why is it important to flush the remaining gossip/control messages once the test is complete ?

aschmahmann: Not really sure, perhaps it's just supposed to help close things down when running consecutive tests, even though that should be covered by cancelling the context 🤷‍♂.

Hopefully either @vyzo can shed some light or perhaps there's an answer to be found in the first gossipsub PR #67.

vyzo:
closing as it doesn't fix the tests and needs quite a bit of work.

PR #204, opened 2019-10-06
gossip methods: renames and predicate adjustment

raulk:
This PR proposes minor renames for accuracy.

It also uses the predicate on `ps.getPeers` to exclude the exclusion set, to sure we source as many as `D` peers to push gossip to. With the current logic, we might select peers that later end up being excluded, resulting in us gossiping to less than `D` peers.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -527,13 +527,15 @@ func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}
 		return
 	}
 
-	gpeers := gs.getPeers(topic, GossipSubD, func(peer.ID) bool { return true })
-	for _, p := range gpeers {
-		// skip mesh peers
+	// Send gossip to D peers, skipping over the exclude set.
+	gpeers := gs.getPeers(topic, GossipSubD, func(p peer.ID) bool {
 		_, ok := exclude[p]
-		if !ok {
-			gs.pushGossip(p, &pb.ControlIHave{TopicID: &topic, MessageIDs: mids})
-		}
+		return !ok
+	})
+
+	// Emit the IHAVE gossip to the selected peers.
+	for _, p := range gpeers {
+		gs.pushGossip(p, &pb.ControlIHave{TopicID: &topic, MessageIDs: mids})
 	}
 }
 
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -436,6 +436,8 @@ func (gs *GossipSubRouter) heartbeat() {
 			}
 		}
 
+		// 2nd arg are mesh peers excluded from gossip. We already push
+		// messages to them, so its redundant to gossip IHAVEs.
 		gs.emitGossip(topic, peers)
 	}
 
@@ -472,6 +474,8 @@ func (gs *GossipSubRouter) heartbeat() {
 			}
 		}
 
+		// 2nd arg are fanout peers excluded from gossip. We already push
+		// messages to them, so its redundant to gossip IHAVEs.
 		gs.emitGossip(topic, peers)
 	}
 
@@ -515,7 +519,9 @@ func (gs *GossipSubRouter) sendGraftPrune(tograft, toprune map[peer.ID][]string)
 
 }
 
-func (gs *GossipSubRouter) emitGossip(topic string, peers map[peer.ID]struct{}) {
+// emitGossip emits IHAVE gossip advertising items in the message cache window
+// of this topic.
+func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}) {
 	mids := gs.mcache.GetGossipIDs(topic)
 	if len(mids) == 0 {
 		return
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -547,7 +555,7 @@ func (gs *GossipSubRouter) flush() {
 	}
 }
 
-func (gs *GossipSubRouter) pushGossip(p peer.ID, ihave *pb.ControlIHave) {
+func (gs *GossipSubRouter) queueGossip(p peer.ID, ihave *pb.ControlIHave) {
```
aarshkshah1992: Would be nice to have a comment saying that the queued gossip will be sent/piggy-backed with the next set of control messages or flushed during the next heartbeat.

aarshkshah1992:
Spec change PR https://github.com/libp2p/specs/pull/219

aschmahmann:
@raulk No problem with these changes, however when I've spoken to @vyzo about changing the predicate in the past he has mentioned that this is an intentional change. So we may need to first decide whether the real issue is docs or implementation (i.e. libp2p/docs#66).

My understanding of the argument to leave the predicate alone is that we'd actually like the effective `D` peers we message to decrease as the percentage of peers we know about are ones in our mesh increases. We do this because the whole point of gossip is to send out messages to peers who likely have not seen our message (note: if a peer sends us a GRAFT between message send and the first gossip they are guaranteed to not to receive the message from us), and so if we have 18 peers with 12 in our mesh we don't really feel the need to send the messageID to all 6 of the remaining peers.

If we are onboard with the above logic (or if @vyzo has other points that I'm missing) we **must** fix the documentation (@vyzo disagrees and says this is just an implementation optimization), otherwise people implementing simulations will achieve different results from us. Additionally, if we wanted we could make the logic a little more predictable by using the proposed predicate (filtering based on existing peers), but setting the number to be based on the fraction of peers not in the mesh e.g. `D*(len(topicPeers) - len(meshPeers))/len(topicPeers)`

No complaint about simply adopting this PR if we do not think the above argument is sufficient. However, we should also make sure to change it in the simulations so that they stay consistent.

EDIT: Also, if we do this change we should make sure to propagate this information to the JS and Rust implementations.

raulk:
@aschmahmann Thoughts:

* Yeah, if we're seeking a balancing effect here it needs to be formally captured. Right now it's an unspoken truth.
* This is asymmetrically biased. If we're above D, we penalise/suppress some gossip, but if we're below D, we do not boost gossip.
* To me, this seems like the inverse heuristic of what should be happening, and it can have a detrimental emergent effect.
    * If a peer is meshing above D, it's probably because it's a well-advertised and discoverable peer for that topic.
    * As a result, it is likely a hot node in the dissemination graph for that topic.
    * For the benefit of the network, we'd want that peer to emit _**more** gossip_ to nodes that didn't make it into its mesh, not less.

---

I _think_ Rust already behaves the way proposed herein: https://github.com/libp2p/rust-libp2p/pull/898/files#diff-88fd6ff4aa853343221bf015d3c5a32bR768 (follow the breadcrumbs).

gossipsub-js seems to not impose an upper limit on gossip: https://github.com/ChainSafe/gossipsub-js/blob/a6bd1850085176b912bc9e1c350122e34c8eefed/src/index.js#L547-L556

raulk:
Let me be more precise:

> Yeah, if we're seeking a balancing effect here it needs to be formally captured. Right now it's an unspoken truth.

This behaviour *is* specified in the specs (https://github.com/libp2p/specs/pull/219). If this is deliberate, we should explain the motivation.

aschmahmann:
> gossipsub-js seems to not impose an upper limit on gossip: 

@raulk I'm confused. The line right before the ones you linked to limits the gossip to `constants.GossipSubD`

gossipsub.go
```diff
@@ -515,19 +519,23 @@ func (gs *GossipSubRouter) sendGraftPrune(tograft, toprune map[peer.ID][]string)
...
+// of this topic.
+func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}) {
 	mids := gs.mcache.GetGossipIDs(topic)
 	if len(mids) == 0 {
 		return
 	}
 
-	gpeers := gs.getPeers(topic, GossipSubD, func(peer.ID) bool { return true })
+	// Send gossip to D peers, skipping over the exclude set.
+	gpeers := gs.getPeers(topic, GossipSubD, func(p peer.ID) bool {
```
vyzo, line 531: this is a subtle change as it will send slightly more gossip, but I am ok with it.

vyzo (approved):

gossipsub.go
vyzo: I might sound a little pedantic, but `queue` is not a verb; we _enqueue_, which is synonymous with pushing to a queue.

vyzo:
Let's fix the naming of `queueGossip`; the old name was correct, as pushing to a queue is synonymous with enqueuing in common programming language lingo.
Regardless, if you are confused by `push`, call it `enqueue`, not `queue`.

gossipsub.go
raulk: It is pedantic, but it's also accurate! Changing.

raulk: @aarshkshah1992

> Would be nice to have a comment saying that the queued gossip will be sent/piggy-backed with the next set of control messages or flushed during the next heartbeat.

That's part of the overall design of gossipsub; a one-off comment is not the appropriate place to convey that aspect to the user.

gossipsub.go
raulk: 👍

vyzo:
Great, merging.

PR #207, opened 2019-10-09
add sender id to message

adam-hanna:
Unless I'm mistaken, there's no way to get the sender ID from a message.

`msg.GetFrom()` returns the originator's id, not the most recent sender. There could be use cases where knowing who sent this message is valuable. For instance, I'm building a small benchmark tool and I need to calculate last delivery hop. I do so by logging recipient host id and sender id. My analyzer then parses the logs and calculates each message path.

I don't know if my simple implementation accomplishes this goal correctly or if there is a better way but would love to hear your thoughts.

Thanks!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,5 @@ require (
 	github.com/multiformats/go-multistream v0.1.0
 	github.com/whyrusleeping/timecache v0.0.0-20160911033111-cfcb2f1abfee
 )
+
+go 1.13
--- a/pubsub.go
+++ b/pubsub.go
@@ -135,10 +135,15 @@ type PubSubRouter interface {
 
 type Message struct {
 	*pb.Message
+	senderID peer.ID
 }
 
 func (m *Message) GetFrom() peer.ID {
 	return peer.ID(m.Message.GetFrom())
+}
+
+func (m *Message) GetSenderID() peer.ID {
+	return m.senderID
 }
 
 type RPC struct {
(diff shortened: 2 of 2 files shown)


adam-hanna:
broken, one sec...

adam-hanna:
ok, I believe this is a more correct implementation.

This isn't ready to be merged, it was mostly opened to show a possible implementation. Also, I did not write any tests. After discussion, I'm happy to create a PR that could be ready for merge.

go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	github.com/multiformats/go-multistream v0.1.0
 	github.com/whyrusleeping/timecache v0.0.0-20160911033111-cfcb2f1abfee
 )
+
+go 1.13
```
adam-hanna, line 15: I didn't do this. It must've been added via go?

pubsub.go
```diff
@@ -135,12 +135,17 @@ type PubSubRouter interface {
 
 type Message struct {
 	*pb.Message
+	senderID peer.ID
```
vyzo, line 138: this is available in the `RPC` struct and will be exposed to routers in the refactoring in #199.
It is also visible to validators as it is passed in the validation function.

go.mod
vyzo: It's done automatically be go; please remove this line.

vyzo (commented):
Not sure if we want that, see notes on exposure.

vyzo:
To clarify, I don't think this is relevant (the direct source of the message) to subscribers, as it is entirely non deterministic.
Routers and validators, who need it, have access to it.

adam-hanna:
Ah yes, I agree!

I didn't realize it was already made available on the validator. That worked! Thanks!

Closing...

vyzo:
Note that we did end up adding it in #218, turns out it is needed for filecoin.

Issue #198, opened 2019-09-18
Topic handles and ownership


Stebalien:
# Motivation

* It's not possible to forward messages on a topic without also subscribing.
* It's not possible to instruct pubsub to remain connected to a topic's "mesh" without also subscribing. That means calling `Publish` without first calling `Subscribe` will often just drop the message.
* It's difficult to reliably register/manage validators in the presence of multiple services as validators are registerd per-topic, globally (https://github.com/libp2p/go-libp2p-pubsub/issues/58).

# Proposal

* Introduce a `Join` function that returns a `Topic` handle (which can be closed to leave the topic). Joining a topic handle subscribes to the topic internally, performs any discovery necessary to find peers participating in the topic, and joins the topic mesh.
* Make topic ownership _exclusive_. That is, only a single service will be able to "join" and/or participate in a topic at a time.
* Bind validators to topic handles, unregistering them when the topic handle is closed.

```
type PubSub interface {
    // Join joins the topic. Messages can then be sent/received via the returned topic handle.
    // Only one service may join a topic at a time.
    Join(topic) (Topic, error)
}
type Topic interface {
    // Subscribe returns a subscription to messages on the topic.
    Subscribe(ops) Subscription
    // Publish publishes messages on the topic.
    Publish(message) error
    // Close leaves the topic.
    Close() error
    ...
}
```

# Discussion

There are two somewhat orthogonal parts to this proposal: topic handles and exclusivity.

## Topic Handles

The motivation for the topic handle is pretty clear. Unless we're in a fully connected network, `Publish` can't currently be used independently of `Subscribe` and topic handles try to resolve this issue.

## Exclusivity

The concern here was: how do we make it easy for multiple services to reliably use a single pubsub service at the same time. The specific concern was validator registration.

We landed on three options:

1. Global validators applied to a _class_ of topics (e.g., using pattern matching).
2. Per-subscription validators and independent subscriptions.
3. Per-subscription validators and exclusive subscriptions.

### Global Validators

This is the option I proposed in #58. The user would have to setup global topic validators and would be responsible for managing conflicts up-front. This is pretty close to how DHT validators work.

However, unlike the DHT, we expect services using pubsub to add new validators for new topic types at runtime. A single over-zealous validator registered by one service could interfere with other services.

### Independent Subscriptions

Next we considered associating validators with subscriptions (or topic handles). The idea was to:

1. Allow multiple services to "join" the same topic.
[...]

PR #208, opened 2019-10-11
Allow peer to be in Relay ONLY mode for a topic

aarshkshah1992:
For #28 

@aschmahmann  As discussed, this PR will allow a peer to 'Relay' for a topic without having to 'subscribe'/'consume' messages for that topic. However, a 'Subscriber' will always be a 
'Relayer'.

Please let me know if the tests make sense/we need more tests.

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -363,6 +363,145 @@ func TestGossipsubGossip(t *testing.T) {
 	time.Sleep(time.Second * 2)
 }
 
+func TestRelayOnlyPeerForwardsMessage(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	hosts := getNetHosts(t, ctx, 8)
+	psubs := getPubsubs(ctx, hosts)
+
+	connect(t, hosts[0], hosts[1])
+	connect(t, hosts[1], hosts[2])
+	connect(t, hosts[2], hosts[3])
+	connect(t, hosts[1], hosts[4])
+
+	connect(t, hosts[0], hosts[5])
+	connect(t, hosts[5], hosts[6])
+	connect(t, hosts[6], hosts[7])
+	/*
+		R= Relay ONLY
+		S = Subscriber
+
+			[0](R) -> [1](R) -> [2](R) -> [3](S)
+			 |      L->[4](S)
+			 v
+			[5](R)
+			 |
+			 v
+			[6](R) -> [7](S)
+	*/
+
+	var subs []*Subscription
+	for i, ps := range psubs {
+		if i == 3 || i == 4 || i == 7 {
+			ch, err := ps.Subscribe("foo")
+			if err != nil {
+				panic(err)
+			}
+			subs = append(subs, ch)
+		} else {
+			_, err := ps.Join("foo")
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+
+	// wait for heartbeats to build mesh
+	time.Sleep(time.Second * 2)
+
+	// assert that  published message is received
+	for i := 0; i < 100; i++ {
+		msg := []byte(fmt.Sprintf("%d it's not a floooooood %d", i, i))
+
+		owner := rand.Intn(len(psubs))
+
+		psubs[owner].Publish("foo", msg)
+
+		for _, sub := range subs {
+			got, err := sub.Next(ctx)
+			if err != nil {
+				t.Fatal(sub.err)
+			}
+			if !bytes.Equal(msg, got.Data) {
+				t.Fatal("got wrong message!")
+			}
+		}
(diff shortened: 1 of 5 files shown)


pubsub.go
```diff
@@ -684,6 +758,50 @@ func (p *PubSub) Subscribe(topic string, opts ...SubOpt) (*Subscription, error)
...
 
+type RelayOpt func(r *Relay) error
+
+// Join returns a new Relay for the given topic which will enable us to
+// propagate messages we receive to other peers without having to 'consume'/'subscribe' them.
+// For now, we ONLY support Relay without subscription.
+// We do NOT allow subscription without Relay.
+// Note that Join is not an instanteneous operation. It may take some time
+// before it is processed by the pubsub main loop and propagated to our peers.
+func (p *PubSub) Join(topic string, opts ...RelayOpt) (*Relay, error) {
```
aarshkshah1992, line 769: Since `Subscribe` = `Join` + subscription specific stuff,  we need to create a `Relay` object every time `Join` is called in order to reference count subscribers so we can announce we are leaving the topic if we have no more subscribers & "un-cancelled" RelayOnly objects pending.  

However, this means that will create multiple `Relay` objects even if the  PubSub client/peer directly calls `Join`(without calling `Subscribe`) more than once in order to become a Relayer. I'm not sure if that makes sense because a peer is either RelayOnly or not. Should I add a comment saying that `Join` should ONLY be called once by the client?

gossipsub_test.go
```diff
@@ -363,6 +363,145 @@ func TestGossipsubGossip(t *testing.T) {
 	time.Sleep(time.Second * 2)
 }
```
aarshkshah1992, line 365: Need to add some tests for Floodsub as well.

gossipsub_test.go
```diff
@@ -363,6 +363,145 @@ func TestGossipsubGossip(t *testing.T) {
...
+				t.Fatal("got wrong message!")
+			}
+		}
+	}
+}
+
+func TestGossipsubGossipWithSomeRelayOnlyPeers(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	hosts := getNetHosts(t, ctx, 15)
```
aarshkshah1992, line 436: Change number of hosts to 20 for consistency.

pubsub.go
vyzo: We should have a single relay object and reference count; Subscribing increases the reference count, unsubscribe decreases.
In fact, I am not even sure we _need_ a relay object at all, just a reference count in the topic map.

aschmahmann:
@aarshkshah1992 thanks so much for taking a stab at this!

Brief initial thoughts (which I hope to expand on when I've got a little more time) are that this seems a little overkill. Granted this is likely because you did the natural thing and emulated what the rest of the code looks like, but since the relay option (as @vyzo and @raulk specified in #28) seems to really just be about not allocating the Subscription message buffer we could probably significantly shrink the scope.

Perhaps it would make sense to just add a flag to subscription like `receiveMsgs` and then have some if statements that don't instantiate the channel (or send to it) if the flag is false. We could even wrap a Relay struct around a Subscription and just not expose the `Next` function. This would also give us events for Relays, which we could also have an option to turn off if we wanted.

@vyzo any thoughts?

pubsub.go
aschmahmann: > In fact, I am not even sure we need a relay object at all, just a reference count in the topic map.

I don't even think we need a reference count at all. If we take the strategy from #198 and https://github.com/libp2p/go-libp2p-pubsub/pull/184#discussion_r330393936 then there should really only be one topic handle and that topic handle should be able to optionally be a relay.

vyzo: Yeah, that's a valid strategy.

vyzo:
I don't think we need a flag in the subscription -- we can simply not have a subscription object at all when we are only relaying.

aschmahmann:
> we can simply not have a subscription object at all when we are only relaying.

@vyzo the flag helps cut down on the complexity of managing another struct (the relay struct) and adding many more lines of boilerplate code to pubsub. However, if you really prefer that approach I'm not going to block on it.

vyzo:
That's not what I meant -- I don't think we need either a relay object or to instantiate a subscription object for this.

raulk:
Sorry to put a stick on the wheels, but I think https://github.com/libp2p/go-libp2p-pubsub/issues/198 comes before this. Once that design is implemented, it will make marking a topic as relay only much more natural.

Stebalien:
Topic handles now exist.

aarshkshah1992:
Starting work on this again now that the amazing work on Topic handlers has landed.

aarshkshah1992:
Closing this in favour of #229

Issue #211, opened 2019-10-14
Question: expected behavior on empty mesh peers?


ZenGround0:
I've noticed that the following situation can occur during gossipsub publish:

1. There are peers in the pubsub topic map for this topic [this ok is true](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L217)
2. This router is in the mesh for the topic [this ok is true](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L230)
3. The router has not added any peers to the mesh [gmap and tosend are empty](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L252)

This leads to the counterintuitive result that peers that are not in the mesh will end up publishing, but peers that are in the mesh (without peers) will not end up publishing.  

Is this behavior expected?

vyzo:
If there are no peers in the mesh, there is nothing to publish ... except we might have floodsub peers.
That latter may be a bug, otherewise it is expected behaviour.

Regardless, we may want some kind of event notification for when we have peers so that the application knows when to start publishing.

ZenGround0:
> Regardless, we may want some kind of event notification for when we have peers so that the application knows when to start publishing.

That would be amazing for my use case

aschmahmann:
> Regardless, we may want some kind of event notification for when we have peers so that the application knows when to start publishing.

#184 allows the Publish function to block on waiting for a bootstrapped condition to be met (e.g. connected to a certain number of peers).

@ZenGround0 is that sufficient? It feels much better then an event, since a node could lose all its peer connections and then we would need to fire a "we don't have enough peers event" and manage that (worry about what happens if we drop an event, etc.).

vyzo:
But that's for peers that are not subscribed, no? We need something for peers in the mesh.

aschmahmann:
No, it looks for peers in the mesh. See https://github.com/aschmahmann/go-libp2p-pubsub/blob/45117e8a47af41fd65eaa03ae8b6298adbe96b68/gossipsub.go#L97 for more details.

vyzo:
We can still emit events regardless of blocking publish -- so that the application can decide exactly when it can start publishing.
Blocking is a little rude for many use cases.
It is a slightly different use case than the pubsub-router.

aschmahmann:
[...]

PR #213, opened 2019-10-14
Empty gmap fans out too

ZenGround0:
I don't know if #211 is expected behavior, but if it is not then this will correct it.  If #211 is expected please close this down.


vyzo (commented):
this is not a bug -- the gmap may be empty because we are the first peer in the topic and it will eventually get filled in the heartbeat.

Issue #215, opened 2019-10-18
Expose the peer that delivered the message


aschmahmann:
It may be useful for peers that want to layer protocols on top of pubsub to know which peer actually delivered a message to them.

We already have this information available in the `from` field of the RPC:

https://github.com/libp2p/go-libp2p-pubsub/blob/2247a54a8cfc7f84d54d28429a60fa4ce89d5c43/pubsub.go#L144-L149

so this would mean making the information publicly available in the Message struct:

https://github.com/libp2p/go-libp2p-pubsub/blob/2247a54a8cfc7f84d54d28429a60fa4ce89d5c43/pubsub.go#L136-L138

aschmahmann:
@raulk @vyzo @ZenGround0

Do we think this is sufficiently reusable such that we're better off recommending this approach rather than the hackier, but already available, approach of using the validators? Note, that the validators take the peer that delivered the message as an input so we actually already have an avenue for getting the `from` into application space.

vyzo:
Using the validators is just too hacky, let's expose `from`.

raulk:
Closed in #218.

PR #218, opened 2019-10-18
Expose the peer that propagates a message to the recipient

aschmahmann:
This is a solution for #215. We were doing a lot of passing around `(src, Message)` to avoid exposing the source next to the message, so I hope that solving #215 is a good thing 😬.

I haven't removed all the places where we pass both `src` and `Message`, but got most of the internal ones.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,5 @@ require (
 	github.com/multiformats/go-multistream v0.1.0
 	github.com/whyrusleeping/timecache v0.0.0-20160911033111-cfcb2f1abfee
 )
+
+go 1.13
--- a/pubsub.go
+++ b/pubsub.go
@@ -79,7 +79,7 @@ type PubSub struct {
 	topics map[string]map[peer.ID]struct{}
 
 	// sendMsg handles messages that have been validated
-	sendMsg chan *sendReq
+	sendMsg chan *Message
 
 	// addVal handles validator registration requests
 	addVal chan *addValReq
@@ -135,6 +135,7 @@ type PubSubRouter interface {
 
 type Message struct {
 	*pb.Message
+	ReceivedFrom peer.ID
 }
 
 func (m *Message) GetFrom() peer.ID {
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	github.com/multiformats/go-multistream v0.1.0
 	github.com/whyrusleeping/timecache v0.0.0-20160911033111-cfcb2f1abfee
 )
+
+go 1.13
```
raulk, line 15: Accidental version bump.

pubsub.go
```diff
@@ -734,10 +736,11 @@ func (p *PubSub) GetTopics() []string {
 // Publish publishes data to the given topic.
 func (p *PubSub) Publish(topic string, data []byte) error {
 	seqno := p.nextSeqno()
+	hostID := p.host.ID()
```
raulk: nit: just `id` is fine. `hostID` sounds alien, as hosts don't have IDs, but rather peers do.

raulk (approved):
This looks like a pretty good rationalisation to me anyway, even if we didn't have to address #218.

go.mod
aschmahmann: What's the deal with the go version being in the go.mod? Any reason we should remove it? GoLand keeps adding it for me automatically whenever I compile.

raulk: The go1.13 toolchain keeps doing this. On re-reading the [modules docs released with go1.12](https://golang.org/doc/go1.12#modules), I think it's fine to keep this directive.

I thought it specified the minimum go toolchain version that's needed to build this module, and that builds with earlier versions (e.g. go1.12) would simply fail. That's not the case, though. Builds will succeed unless they use language features available in go1.13 that aren't available in the toolchain in use. When that happens, this directive allows the compiler to give better hints when barking.

raulk: Conclusion: ok to keep this.

raulk:
Let's wait for @vyzo to sign-off too.

vyzo (approved):
LGTM

vyzo:
Test failure is systemic, @aschmahmann you need to `go fmt`!

```
*** go fmt ./...

ERROR: some files not gofmt'ed:

./floodsub_test.go
```

aschmahmann:
> @aschmahmann you need to go fmt!

lol that's what I get for trying to rename a variable using the Github text editor

PR #225, opened 2019-11-01
Topic handler bug fixes

aschmahmann:
Fixes the issues named in #224 namely:

Calling Topic.Close() throws an NRE 😨
Calling Topic.Close() still allows some calls to be utilized (e.g. can still Publish after Close)

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -789,9 +789,13 @@ func (p *PubSub) tryJoin(topic string, opts ...TopicOpt) (*Topic, bool, error) {
 	}
 
 	resp := make(chan *Topic, 1)
-	t.p.addTopic <- &addTopicReq{
+	select {
+	case t.p.addTopic <- &addTopicReq{
 		topic: t,
 		resp:  resp,
+	}:
+	case <-t.p.ctx.Done():
+		return nil, false, t.p.ctx.Err()
 	}
 	returnedTopic := <-resp
 
@@ -848,7 +852,11 @@ type topicReq struct {
 // GetTopics returns the topics this node is subscribed to.
 func (p *PubSub) GetTopics() []string {
 	out := make(chan []string, 1)
-	p.getTopics <- &topicReq{resp: out}
+	select {
+	case p.getTopics <- &topicReq{resp: out}:
+	case <-p.ctx.Done():
+		return nil
+	}
 	return <-out
 }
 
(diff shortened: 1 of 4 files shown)


topic.go
```diff
@@ -10,20 +11,33 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
+// errTopicClosed is returned if a Topic is utilized after it has been closed
+var errTopicClosed = errors.New("this Topic is closed, try opening a new one")
```
aschmahmann: @Stebalien export?

Stebalien: Yes.

topic.go
```diff
@@ -148,13 +174,31 @@ func WithReadiness(ready RouterReady) PubOpt {
...
 // Does not error if the topic is already closed.
 func (t *Topic) Close() error {
+	t.mux.Lock()
+	defer t.mux.Unlock()
+	if t.closed {
+		return nil
+	}
+
 	req := &rmTopicReq{t, make(chan error, 1)}
 	t.p.rmTopic <- req
```
Stebalien: https://github.com/libp2p/go-libp2p-pubsub/issues/226

Stebalien (approved):

topic.go
aschmahmann: This was true for all previous PubSub functions and the new Topic ones. I've pushed an update, lmk if that's what you had in mind.

Three questions about fixing this:
1) Do I want custom `ErrPubSubClosed` messages or is the context error enough (both for the PubSub functions and for the Topic functions)?
2) Should I set `Topic.closed=true` on any/all of the pubsub context errors?
3) It seems more reasonable to return an empty array then nil for `GetTopics` and `ListPeers` right?

pubsub.go
```diff
@@ -848,7 +852,11 @@ type topicReq struct {
 // GetTopics returns the topics this node is subscribed to.
 func (p *PubSub) GetTopics() []string {
 	out := make(chan []string, 1)
-	p.getTopics <- &topicReq{resp: out}
+	select {
+	case p.getTopics <- &topicReq{resp: out}:
+	case <-p.ctx.Done():
+		return []string{}
```
vyzo: we can just return `nil` here.

pubsub.go
```diff
@@ -880,16 +888,23 @@ type listPeerReq struct {
...
 func (p *PubSub) ListPeers(topic string) []peer.ID {
 	out := make(chan []peer.ID)
-	p.getPeers <- &listPeerReq{
+	select {
+	case p.getPeers <- &listPeerReq{
 		resp:  out,
 		topic: topic,
+	}:
+	case <-p.ctx.Done():
+		return []peer.ID{}
```
vyzo: here too, why not just `nil`?

vyzo (approved):
LGTM modulo the nit about returning `nil` for empty lists.
But that's not normative, feel free to fix or leave it as is.

pubsub.go
aschmahmann: Figured that it would be less breaking to return empty then nil in case some code doesn't check for `nil`. If you think `nil` is better though just give the say so and I'll fix it.

topic.go
aschmahmann: @vyzo  any thoughts on the first 2 questions here? Would like to get this pushed out ASAP since the `Close` bug is particularly bad.

pubsub.go
vyzo: It's more idiomatic to return `nil`.

topic.go
vyzo: 1. either works, it's a very small thing.
2. probably not, I don't see what we would gain from this.

PR #227, opened 2019-11-04
tracing support

vyzo:
~Currently only the tracing scaffolding, but this should be enough for initial review pass.~

TBD:
- [x] implement tracer methods with particular event types
- [x] implement tracing to file
- [x] implement tracing to peer


Stebalien (dismissed):
This seems too "not invented here".

Stebalien:
(we discussed this on a call but I misunderstood some important context)

If we were trying to trace across machines, I'd understand maybe doing this ourselves. Given that we're _just_ tracing events locally, this seems way over the top. Why not just use Zap?

If the answer is "performance", have you tested that? I really doubt logging is going to be a bottleneck given encrypted transports, signing, verifying, etc.

Stebalien:
Unless the concern is bandwidth. That's the one thing this buys us over a general-purpose logger like zap.

trace.go
```diff
@@ -0,0 +1,336 @@
...
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+)
+
+// Generic event tracer interface
+type EventTracer interface {
+	Trace(evt interface{})
```
Stebalien: `pb.Message`?

Stebalien:
Ok, I've discussed this with raul and looked at it a bit more. This still feels like a lot of code for something so simple but I understand why we're using protobufs (gRPC & bandwidth).

vyzo:
Yes, we _are_ trying to trace across machines, it's just that only the local tracer has been written so far.
Tracing to peer (and gRPC based tracing) are forthcoming.
Also note, bandwidth is a real concern here.

trace.go
vyzo: Yeah, we could do that to avoid the cast.

vyzo: done, it takes a `*pb.TraceEvent` now.

Stebalien:
> Yes, we are trying to trace across machines

I'm talking about including trace IDs in messages. However, message IDs should give us that anyways.

vyzo:
> I'm talking about including trace IDs in messages. 

The peer ID of the originating peer is included in the message.

vyzo:
Rebased on master.

floodsub.go
```diff
@@ -39,11 +40,16 @@ func (fs *FloodSubRouter) Protocols() []protocol.ID {
 
 func (fs *FloodSubRouter) Attach(p *PubSub) {
 	fs.p = p
+	fs.tracer = p.tracer
 }
 
-func (fs *FloodSubRouter) AddPeer(peer.ID, protocol.ID) {}
+func (fs *FloodSubRouter) AddPeer(p peer.ID, proto protocol.ID) {
+	fs.tracer.AddPeer(p, proto)
```
Stebalien, line 47: Should this be in the _router_ or in the part that calls `AddPeer`?

floodsub.go
```diff
@@ -39,11 +40,16 @@ func (fs *FloodSubRouter) Protocols() []protocol.ID {
...
+}
 
-func (fs *FloodSubRouter) RemovePeer(peer.ID) {}
+func (fs *FloodSubRouter) RemovePeer(p peer.ID) {
+	fs.tracer.RemovePeer(p)
```
Stebalien, line 51: ditto.

tracer.go
```diff
@@ -0,0 +1,290 @@
+package pubsub
```
Stebalien, line 1: Double checking, this is moving to a new repo?

Stebalien (commented):
After my initial "why are we doing this" reaction, this LGTM (modulo a few questions/nits).

floodsub.go
vyzo: We could have it in the control substrate, it just felt more natural to do it in the router.
Other than that there is no particular reason for the choice.

tracer.go
vyzo: I think we can keep this file here, so that we have a one-stop construction of the pubsub system without requiring an external dependency.

floodsub.go
vyzo: same, it could be in either place, just felt more natural to do it in the router.

tracer.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	t.mx.Unlock()
+
+	select {
+	case t.ch <- struct{}{}:
+	default:
+	}
+}
+
+func (t *basicTracer) Close() {
```
Stebalien, line 55: Let's make this a bit safer.

tracer.go
```diff
@@ -0,0 +1,297 @@
...
+	for {
+		_, ok := <-t.ch
+
+		// nil out the buffer to gc events when swapping buffers
+		for i := range buf {
+			buf[i] = nil
+		}
+
+		// wait a bit to accumulate a batch
+		time.Sleep(time.Second)
```
Stebalien: We could end up with a pretty large buffer this way and/or drop messages. Is there a better way to do this?

tracer.go
```diff
@@ -0,0 +1,297 @@
...
+	var batch pb.TraceEventBatch
+
+	gzipW := gzip.NewWriter(s)
+	w := ggio.NewDelimitedWriter(gzipW)
+
+	for {
+		_, ok := <-t.ch
+
+		// nil out the buffer to gc events when swapping buffers
+		for i := range buf {
```
Stebalien: Shouldn't we do this at this at the end?

tracer.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+
+			gzipW.Reset(s)
+		}
+	}
+}
+
+func (t *RemoteTracer) connect() error {
+	for {
+		ctx, cancel := context.WithTimeout(t.ctx, time.Minute)
```
Stebalien: Can't we just call open a stream? We shouldn't have to dial separately.

Stebalien (changes requested):

tracer.go
vyzo: that's the buffer we are swapping in, it's already been consumed.

tracer.go
vyzo: We can, but the host will have to know how to reach the trace peer. The traced is not announcing itself to the dht, so it wouldn't be findable in general.

tracer.go
vyzo: Sure.

tracer.go
Stebalien: Yeah, but we could nil it out earlier to avoid holding on to the memory.

tracer.go
vyzo: We could check the size of the buffer and sleep only if the batch is smaller than a threshold.
We can also reduce the sleep time and poll a few times, say up to a second.

tracer.go
Stebalien: Ah. Shouldn't we add the the addresses to the peerstore on start with an expiration of "never" (https://github.com/libp2p/go-libp2p-core/blob/master/peerstore/peerstore.go#L45)?

tracer.go
vyzo: Oh, that's a good point. Ok, I'll move the niling out right after we write it.

vyzo: done.

tracer.go
vyzo: I added a check for the size of the batch, and polling every 100ms (for up to 1s) in order to accumulate buffer in safer way.

tracer.go
vyzo: that would work too; can do.

vyzo: done.

tracer.go
vyzo: I added a closed state variable, which gets set when we first Close, and is also consulted before adding to the event buffer.
It should be safe and idempotent now, no potential panics because of a closed channel.

tracer.go
```diff
@@ -160,12 +173,15 @@ type RemoteTracer struct {
...
-	pi   peer.AddrInfo
+	peer peer.ID
 }
 
 // NewRemoteTracer constructs a RemoteTracer, tracing to the peer identified by pi
 func NewRemoteTracer(ctx context.Context, host host.Host, pi peer.AddrInfo) (*RemoteTracer, error) {
-	tr := &RemoteTracer{ctx: ctx, host: host, pi: pi, basicTracer: basicTracer{ch: make(chan struct{}, 1), lossy: true}}
+	tr := &RemoteTracer{ctx: ctx, host: host, peer: pi.ID, basicTracer: basicTracer{ch: make(chan struct{}, 1), lossy: true}}
+	for _, addr := range pi.Addrs {
+		host.Peerstore().AddAddr(pi.ID, addr, peerstore.PermanentAddrTTL)
```
Stebalien: `host.Peerstore().AddAddrs(pi.ID, pi.Addrs, ...)`?

Stebalien: (we can do them all at once)

tracer.go
```diff
@@ -187,15 +203,17 @@ func (t *RemoteTracer) doWrite() {
...
+		// deadline for batch accumulation
+		deadline := time.Now().Add(time.Second)
 
+	again:
 		t.mx.Lock()
+		if len(t.buf) < MinTraceBatchSize && time.Now().Before(deadline) {
+			t.mx.Unlock()
+			time.Sleep(100 * time.Millisecond)
+			goto again
+		}
```
Stebalien, line 212: for loop?

```go
t.mx.Lock()
for len(t.buf) < MinTraceBatchSize && time.Now().Before(deadline) { 
  t.mx.Unlock()
  time.Sleep(100 * time.Millisecond)
  t.mx.Lock()
}
```

Stebalien (approved):
nits but LGTM

tracer.go
vyzo: ah, good point; will do.

tracer.go
vyzo: the alergy to gotos strikes again! Ok, I'll write it as a for loop.

raulk:
Hey, would’ve liked to review this properly before it was merged to master :-( Didn’t get a chance due to other things, but I’d have appreciated a mention and a heads-up warning that the merge train was leaving.

vyzo:
@raulk you were extremely busy, so didn't want to burden you.
Feel free to make comments about potential improvements, and we can address them.
Note that the grpc trace client will live out of tree, as we don't want a dependency on grpc for mainline.

Issue #220, opened 2019-10-19
[Roadmap issue] Short-term hardening


raulk:
Features needed by Filecoin and ETH2, in order of priority:

- [X] Expose the peer that propagated the message to us (#218)
- [X] Subscription options, which enables...
- [x] Topic handles (#198), so we can...
- [ ] Configurable per-topic max message sizes (#209).
- ~[ ] Duplicate consumption (#216), which is blocked by...~
- ~[ ] Validator cache (we don’t want to validate dupes over and over again, requires message signing to be enabled).~
- [x] Configurable outbound peer queue sizes (#217).
- [ ] Make Publish() report problems / coalesce queued messages (#217).
- [ ] Mark topics as relay-only (#28).
- [ ] Relative prioritization of traffic per topic (needs to be designed and thought carefully) (#210).

vyzo:
Duplicate consumption can be implemented by the application through a custom tracer, just like the score subsystem plugs into the tracer.
The validator cache, which is complex, is then unecessary.

PR #229, opened 2019-11-15
Peers should be able to ONLY Relay for a topic

aarshkshah1992:
For #28  as enlisted in #220 

@aschmahmann @vyzo @Stebalien Please take a look.

Relay only messages will also pass through the validator pipeline as discussed in https://github.com/libp2p/go-libp2p-pubsub/issues/198#issuecomment-543798308)


aarshkshah1992:
ping @aschmahmann

vyzo:
closing as we need to revisit those things in light of v1.1.

Issue #217, opened 2019-10-18
Published messages can be dropped silently


raulk:
Currently if the outbound message queues for all peers in a mesh/fanout are full, gossipsub drops the message silently, which has already bitten us various times (in stress testing, and during actual usage).

This happens here: https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L359

# Proposal

- [ ] Extend the signature of `Publish()` so it can take functional options to configure the reliability characteristics we want.
    - [ ] `WaitUntilQueued(One | All | float64)` => blocks the publish until the message has been queued to as many peers as the arg indicates. Can be implemented easily.
    - [ ] `WaitUntilSent(One | All | float64)` => blocks the publish until the message has actually been dispatched to as many peers as the arg indicates. Requires deeper changes, but reliability is higher.

- [ ] Aggregate RPCs waiting in the queue. Messages are associative and can be merged, in theory.

- [x] Make queue size configurable. It's currently hardcoded to 32, and cannot be changed.

# Affected users

Filecoin, ETH2 gossipsub testing.

raulk:
cc @whyrusleeping

aschmahmann:
> Extend the signature of Publish() so it can take functional options to configure the reliability characteristics we want.

Note this is already available in #184

hsanjuan:
>  Make queue size configurable. It's currently hardcoded to 32, and cannot be changed.

woa I noticed this yesterday in a completely unrelated thing. What I saw is that hammering gossipsub results in messages seemingly not sent around (or dropped on receive).

aschmahmann:
TLDR: This the proposal does not fully solve this problem, but it's not clear it has to. If we want to more fully solve this problem there are more things we can talk through.

@hsanjuan, yep there are some things we can do here but there are some protocol level problems to tackle if we actually wanted to "solve" this problem (see: #197 for more about the problem).

For example, even if `WaitUntilSent` was implemented such that it waited until the other party confirmed receipt of the data a couple things could still go wrong.

1) The receiver could crash, leaving the message effectively unsent
2) The receiver could not have `WaitUntilSent` activated (or not with the same parameters) meaning that while your peers may receive the data their peers might not.
   - We sometimes have conventions where people associate the same options with the same topics, but it's possible we may want to formalize this in some way. (e.g. `topicID = fn(topicName + options)`)

[...]

PR #230, opened 2019-11-15
Configurable outbound peer queue sizes

aarshkshah1992:
For #217 as enlisted in #220 

Low hanging fruit :)  The other tasks enlisted in #217 will be next.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -3,6 +3,7 @@ package pubsub
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/rand"
 	"sync"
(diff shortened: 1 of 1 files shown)


pubsub.go
```diff
@@ -232,6 +237,18 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	return ps, nil
 }
 
+// WithPeerOutboundQueueSize is an option to set the buffer size for outbound messages to a peer
+// We start dropping messages to a peer if the outbound queue if full
+func WithPeerOutboundQueueSize(size int) Option {
+	return func(p *PubSub) error {
+		if size < 0 {
```
vyzo: It can't be zero either.

aarshkshah1992: I see what you mean.  Because we write the `HelloPacket` to the queue before we start the go-routine that reads from it & dispatches it. Thanks for pointing this out.

aarshkshah1992: @vyzo  Have addressed this. Let me know if you have any concerns before we merge this.

vyzo (approved):
LGTM

PR #231, opened 2019-11-15
add ValidatorData field to Message

vyzo:
Useful for passing data from the validation to the application.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -151,6 +151,7 @@ type PubSubRouter interface {
 type Message struct {
 	*pb.Message
 	ReceivedFrom peer.ID
+	VaidatorData interface{}
 }
 
 func (m *Message) GetFrom() peer.ID {
@@ -688,7 +689,7 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 			continue
 		}
 
-		msg := &Message{pmsg, rpc.from}
+		msg := &Message{pmsg, rpc.from, nil}
 		p.pushMsg(msg)
 	}
 
--- a/topic.go
+++ b/topic.go
@@ -167,7 +167,7 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error 
 	}
 
 	select {
-	case t.p.publish <- &Message{m, id}:
+	case t.p.publish <- &Message{m, id, nil}:
 	case <-t.p.ctx.Done():
 		return t.p.ctx.Err()
 	}


whyrusleeping (approved):

whyrusleeping:
Thanks @vyzo !

pubsub.go
```diff
@@ -151,6 +151,7 @@ type PubSubRouter interface {
 type Message struct {
 	*pb.Message
 	ReceivedFrom peer.ID
+	VaidatorData interface{}
```
aschmahmann, line 154: @vyzo where is this field set, did you forget to add a file?

aschmahmann: Also, is this just a convenience thing since the data is already available by emitting data from the validators?

aschmahmann (dismissed):

pubsub.go
whyrusleeping: Yes, this is for validators to set, and for subscribers to consume. Basically, I'd like to avoid having to unmarshal things twice

vyzo: it is initialized to `nil` when we construct the message; it's for use from validators, they can set it to pass data to the application (eg. the unmarshalled payload to avoid unmarshalling twice).
This is a filecoin ask.

aschmahmann: @vyzo Ok, no problem with convenience functions.

Although, I'd like to point out that in less than 30 minutes you've just passed a change that causes problems if there are multiple validators for a single message since they may both clobber the `ValidateData` field. There is also no documentation for this behavior.

vyzo: Welcome to agile software engineering. It's up to the validators to coordinate, no way to enforce that.

aschmahmann: That is literally the opposite of what was agreed at https://github.com/libp2p/go-libp2p-pubsub/issues/198

vyzo: So what exactly would you propose instead of the envelope field?
Also, that's something we _could_ change when we implement #198 stuff.

vyzo: Also note that the application controls the validators; if it needs to attach overlapping validators (for messages that propagate across multiple topics), it can add coordination logic of its own in the validators.

aschmahmann: > So what exactly would you propose instead of the envelope field?

One of:
1. The validators can internally emit data (e.g. they put data in a `map[msgID]ValidatorData` where msgID can be a pointer or some other identifier
2. ValidateData can be `map[topic]interface{}` (we've already restricted to one validator per topic so this shouldn't run into any issues)

Given that this was asked for by @whyrusleeping I assume option 1 was problematic/annoying (although understanding why would be really helpful), so let's try option 2?

vyzo: you'll need a mutex too, and I don't want to bloat this thing further.

Stebalien: This completely breaks the abstraction of pubsub as a _shared service_. There's no reason we can't have validators _return_ additional data (along with an error) and have the validation subsystem collect all the extra validation data.

fxfactorial: So what's the proper usage of this field?

vyzo: It allows the validation to pass data to the application; one common usage (eg in lotus) is to pass the decoded message to avoid parsing and decoding twice.

fxfactorial: I get that and that’s my intent as well , just seems like there was an extended convo about it and wondering if any issues or assumptions need to be held when using it this way

vyzo: The issue in the convo was about multiple validators competing for the coveted field, but it's a very rare occurrence to have a message that validates in multiple topics.

PR #232, opened 2019-11-17
WithWaitUntilQueued option for topic.Publish

aarshkshah1992:
Implementation for the `WaitUntilQueued `option  as specified in #217 & enlisted in #220.

@aschmahmann @vyzo @Stebalien 

Please take a look. This is my first "meaty" PubSub PR & I'd really love to know your feedback/if there is a better way to do this.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -79,6 +79,11 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
 	}
 
 	out := rpcWithMessages(msg)
+
+	// fetch channel we can write to after we attempt to add a message to on a peer's outbound queue
+	listener, hasListener := fs.p.msgQueuedEventListeners[msgID(msg)]
+
+	// start adding the message to the outbound queue for receiver peers
 	for pid := range tosend {
 		if pid == from || pid == peer.ID(msg.GetFrom()) {
 			continue
@@ -89,12 +94,26 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			continue
 		}
 
+		var success bool
 		select {
 		case mch <- out:
+			success = true
 		default:
+			success = false
 			log.Infof("dropping message to peer %s: queue full", pid)
 			// Drop it. The peer is too slow.
 		}
+
+		if hasListener {
+			select {
+			case <-listener.ctx.Done():
+			case listener.notifChan <- &msgQueuedNotification{pid, success}:
+			}
+		}
+	}
+
+	if hasListener {
+		close(listener.notifChan)
 	}
 }
 
--- a/floodsub.go
+++ b/floodsub.go
@@ -80,7 +80,7 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
 
 	out := rpcWithMessages(msg)
 
-	// fetch channel we can write to after we attempt to add a message to on a peer's outbound queue
+	// fetch listener to send notification to after we attempt to add a message to a peer's outbound queue
 	listener, hasListener := fs.p.msgQueuedEventListeners[msgID(msg)]
 
 	// start adding the message to the outbound queue for receiver peers
--- a/topic.go
+++ b/topic.go
@@ -13,6 +13,15 @@ import (
 	pkgerrors "github.com/pkg/errors"
 )
 
+type PublishWaitUntilQueued int
+
+const (
+	NoOp PublishWaitUntilQueued = iota
+	PublishWaitUntilOneQueued
+	PublishWaitUntilAllQueued
+	PublishWaitUntilNQueued
+)
+
 // ErrTopicClosed is returned if a Topic is utilized after it has been closed
 var ErrTopicClosed = errors.New("this Topic is closed, try opening a new one")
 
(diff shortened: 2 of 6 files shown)


topic.go
```diff
@@ -184,6 +290,23 @@ func WithReadiness(ready RouterReady) PubOpt {
 	}
 }
 
+// WithWaitUntilQueued blocks the publish until the message has been added to the outbound message queue for
+// as many peers as the arg indicates
+// A value of -1 means all peers in mesh/fanout for gossipsub & all subscribed peers in floodsub
+// Please note that if nPeers is -1, the behavior is not fail fast
```
aarshkshah1992: Does a "non fail fast" behaviour make sense here ? If the caller is willing to wait for the message to be added to the queues for all peers, I figured we might as well wait till the end so we can log all failures for better debugging later.

topic_test.go
```diff
@@ -111,6 +112,79 @@ func testTopicCloseWithOpenResource(t *testing.T, openResource func(topic *Topic
...
+		"Success if we get enough notifications": {
+			1,
+			3,
+			nil,
+		},
+		"Successful if we want to wait for all peers": {
+			0,
+			-1,
+			nil,
+		},
```
aarshkshah1992, line 180: What's the best way to test for dropped messages ? As in, "Failure for the Wait for all peers" scenario.

floodsub.go
```diff
@@ -89,12 +94,26 @@ func (fs *FloodSubRouter) Publish(from peer.ID, msg *pb.Message) {
...
 		default:
+			success = false
 			log.Infof("dropping message to peer %s: queue full", pid)
 			// Drop it. The peer is too slow.
 		}
+
+		if hasListener {
+			select {
+			case <-listener.ctx.Done():
+			case listener.notifChan <- &msgQueuedNotification{pid, success}:
```
vyzo, line 110: This is way too heavy, it is going to spam with events for every message sent.
Also note that the channel could block the event loop, which is a no,no.

vyzo (changes requested):
I don't think this is the right way to approach this.

aarshkshah1992:
@vyzo  I understand your concern. Please can you give some suggestions on how to approach this ?

vyzo:
Let me think about it.

floodsub.go
aarshkshah1992: @vyzo 

I see what you mean.

We could use a buffered channel here & drop the event if the buffer is full.
But, that could lead to lost events & Publish would end up erroneously reporting that it wasn't able to fulfil the target.

vyzo: So yeah, this seems quite impossible to implement correctly. 
We can't block the event loop and we don't want to erroneously report non-delivery (or not report at all).

aarshkshah1992: @raulk  @aschmahmann  Wdyt ? Can we use a buffered channel here & do the `WaitUntilQueued ` on a "best effort" basis ?

Or, is there a smarter way to do this ?

aarshkshah1992:
Also, apologies for jumping the gun on this one and implementing it too soon. Just wanted to get my hands dirty with pubsub. Will be careful going ahead.

vyzo:
No need to apologize! Your enthusiasm is very much welcome!

floodsub.go
vyzo: we could launch a goroutine for every channel put, which would side-step the problem. But I worry about potentially unbounded number of goroutines running if the receiver is stalling.

aarshkshah1992: > We could use a buffered channel here & drop the event if the buffer is full.
> But, that could lead to lost events & Publish would end up erroneously reporting that it wasn't able to fulfil the target.

What we decide here is very important for the `WaitUntilQueuedALL ` as described [here](https://github.com/libp2p/go-libp2p-pubsub/issues/217#issuecomment-554855771)

aarshkshah1992: @vyzo Aha, here is one thing that can work:

1. For `WaitUntilNQueued`:
  
-  **Buffer size=N** & the event loop ONLY writes success cases to the channel
- `Topic.Publish` cancels the listener context as soon as the target is met

2. For `WaitiUntillAllQueued:
    This is slightly tricky as I've explained [here](https://github.com/libp2p/go-libp2p-pubsub/issues/217#issuecomment-554855771). But, we can do it like so:

- **Buffer size = len(pubsub.topics[topic])** when Publish is called
- Event loop writes both success & failure cases to the channel
- Event loop spins up a new go-routine if the buffer is full. Since we have estimated the buffer size based on the number of peers we know are interested in the topic when Publish is called, it effectively prevents spinning up too many go-routines.

Let me know what you think :)

aarshkshah1992: @vyzo @aschmahmann Aha, here is one thing that can work:

**For WaitUntilNQueued:**

-  `Buffer size = N` & the event loop ONLY writes success cases to the channel
-   Event loop drops messages if the buffer is full

**For WaitiUntillAllQueued:**
    This is _slightly_ tricky as I've explained [here](https://github.com/libp2p/go-libp2p-pubsub/issues/217#issuecomment-554855771). But, we can do it like so:

- `Buffer size  =  len(pubsub.topics[topic])` (number of peers interested in the topic) when Publish is called
- Event loop writes both success & failure cases to the channel
- Event loop spins up a new go-routine if the buffer is full. Since we have estimated the buffer size based on the number of peers we know are interested in the topic when Publish is called, it effectively prevents spinning up too many go-routines.

Let me know what you think :)

aarshkshah1992: Hey @aschmahmann @vyzo 

Please take a look at this when you can.

vyzo:
Closing as we need to revisit the semantics of this whole thing.

Issue #233, opened 2019-11-19
The Evolutionary Road to Episub


vyzo:
Gossipsub is the base pubsub protocol and intermediate step towards our endgame, which is the episub protocol.
Episub can be construed as a set of optimizations over vanilla gossipsub and will get the `meshsub/2.0.0` protocol ID, with backwards compatibility with vanilla gossipsub.
Instead of a bing-bang approach, we propose to evolve the implementation of gossipsub towards episub as a set of PRs that expand the scope of the protocol, until we have implemented all the episub features.

Here is a list of the essential features we want for the transition to episub:
- [ ] Peer eXchange on mesh prune; briefly, when a peer gets pruned from the mesh, the pruning peer should return a list of other peers where the pruned peer can connect to reestablish overlay connectivity. This will allow us to recover from pathological bootstrap situations and make the mesh self-bootstrapping. With this extension, we can bootstrap a full mesh starting from a single known peer or a star topology [`meshsub/1.1.0`].
- [ ] Mesh regrafting; briefly, when a peer receives messages only through gossip and not directly through mesh propagation, we want to regraft that peer into the mesh. This will allow us to recover from overlay pathologies because of failures and node churn and minimize the latency globally [`meshsub/1.2.0`].
- [ ] Pathological peer greylisting; briefly, when a peer is consistently dropping messages because it has a poor network connection, it should be pruned from the mesh and transition to only gossip. This is the opposite of mesh regrafting for peers and acts as a circuit-breaker [`meshsub/1.2.0`].
- [ ] Tree forming for stable sources; briefly, this is the fundamental feature of epidemic broadcast trees, which is the inspiration for episub. The idea is that stable sources can form a tree overlay on top of the mesh, where messages from the source only propagate through the tree instead of the full mesh. This will drop the global amplification factor to near 1. This is the final step in the transition, where we switch the protocol to `meshsub/2.0.0`.

cc @raulk @Stebalien @whyrusleeping @ZenGround0

raulk:
@adam-hanna this is the issue we discussed.

PR #234, opened 2019-11-21
gossipsub v1.1: prune peer exchange

vyzo:
Adds support for peer exchange on prune; see #233 
Depends on: 
- [x] https://github.com/libp2p/go-libp2p-core/pull/73
- [x] https://github.com/libp2p/go-libp2p-peerstore/pull/98 (implicit)
- [ ] https://github.com/libp2p/go-libp2p/pull/747 (implicit)

TBD:
- [x] Act on peer exchange when pruned
- [x] tests


vyzo:
cc @whyrusleeping @ZenGround0

vyzo:
Summoning @Stebalien @raulk @Kubuxu; this is ready for review.

gossipsub.go
```diff
@@ -58,24 +73,34 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
-	lastpub map[string]int64                // last publish time for fanout topics
-	gossip  map[peer.ID][]*pb.ControlIHave  // pending gossip
-	control map[peer.ID]*pb.ControlMessage  // pending control messages
+	peers   map[peer.ID]protocol.ID          // peer protocols
+	mesh    map[string]map[peer.ID]struct{}  // topic meshes
+	fanout  map[string]map[peer.ID]struct{}  // topic fanout
+	lastpub map[string]int64                 // last publish time for fanout topics
+	gossip  map[peer.ID][]*pb.ControlIHave   // pending gossip
+	control map[peer.ID]*pb.ControlMessage   // pending control messages
+	backoff map[string]map[peer.ID]time.Time // prune backoff
```
Kubuxu, line 88: gofmt

gossipsub.go
```diff
@@ -239,6 +264,90 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+		_, connected := gs.peers[p]
+		if connected {
+			continue
+		}
+
+		var srr *routing.SignedRoutingState
+		var err error
+		if pi.SignedAddrs != nil {
+			// the peer sent us a signed record; ensure that it is valid
+			srr, err = routing.UnmarshalSignedRoutingState(pi.SignedAddrs)
```
Kubuxu: Does `UnmarshalSignedRoutingState` validate the signed record?

Kubuxu (commented):
initial quick pass

gossipsub.go
```diff
@@ -239,6 +264,90 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+			if gs.p.host.Network().Connectedness(ci.p) == network.Connected {
+				continue
+			}
+
+			log.Debugf("connecting to %s", ci.p)
+			if ci.srr != nil {
+				gs.p.host.Peerstore().AddCertifiedAddrs(ci.srr, peerstore.TempAddrTTL)
+			}
+
+			ctx, cancel := context.WithTimeout(gs.p.ctx, 10*time.Second)
```
aschmahmann: 10 seconds is too short here, especially if there are no certified addresses added to the Peerstore and we need to do a DHT query.

I also used 10 seconds as the connection time in the discovery module (copy-pasted from go-libp2p-pubsub-router), but this was incorrect. It should probably be higher (e.g. a minute) and likely configurable as well.

gossipsub_test.go
```diff
@@ -923,3 +924,60 @@ func TestGossipsubTreeTopology(t *testing.T) {
...
+	for _, ps := range psubs {
+		sub, err := ps.Subscribe("test")
+		if err != nil {
+			t.Fatal(err)
+		}
+		subs = append(subs, sub)
+	}
+
+	// wait a bit for the mesh to build
+	time.Sleep(10 * time.Second)
```
aschmahmann, line 965: Non-blocking: These sleeps are really unfortunate in tests, they make them pretty error prone and/or slow.

A reasonably straightforward fix here might be to create a `getPeers` test helper function that accesses the private state of gossipsub via `eval`. Then we replace the 10 second sleep with a while loop that checks whether our target state has been met.

We can then figure out the "right" approach to this later.

gossipsub.go
```diff
@@ -239,6 +264,90 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+
+	if len(toconnect) == 0 {
+		return
+	}
+
+	// initiate connections, without blocking the event loop
+	go func() {
+		for _, ci := range toconnect {
+			select {
+			case gs.connect <- ci:
```
aschmahmann: I'm concerned about how many goroutines may end up being stuck here if the channel is backed up due to many ongoing slow DHT requests (especially if we end up looking for peers that don't exist/are not dialable).

Two approaches that may help here are:
1) Use a backoff connector object like we have in the discovery module (perhaps even move that object into pubsub so that gossipsub can share it). This will help with minimizing repeated dials to bad peers.
2) Use an infinite buffer (e.g. a map protected by a mutex). I'd rather an infinite buffer than infinite goroutines any day of the week

gossipsub.go
```diff
@@ -239,6 +264,90 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
 			gs.tracer.Prune(p, topic)
 			delete(peers, p)
 			gs.untagPeer(p, topic)
+			gs.addBackoff(p, topic)
```
aschmahmann, line 275: 👍 great!

However, we're not currently doing anything about a peer continuously chasing away other peers by sending a GRAFT. If during a GRAFT we could check if we're already over the limit and send pack a PRUNE that would largely resolve this case.

pb/rpc.proto
```diff
@@ -45,6 +45,12 @@ message ControlGraft {
 
 message ControlPrune {
 	optional string topicID = 1;
+	repeated PeerInfo peers = 2;
+}
+
+message PeerInfo {
+	optional bytes peerID = 1;
+	optional bytes signedAddrs = 2;
```
aschmahmann: Nit: really strange that we're using `bytes` instead of a predefined format we already know (a signed envelope of `RoutingStateRecord`s). Is it just too difficult with our modular libraries to do this properly?

aschmahmann (changes requested):
Looks pretty good to me, left a few comments to address.

Also, noting for posterity that this PR should resolve https://github.com/libp2p/specs/issues/215 🎉

gossipsub.go
vyzo: it is gofmt'ed! I have a save hook that does this for me.

gossipsub.go
vyzo: I think it does, let me double check.

vyzo: yes, it validates.

gossipsub_test.go
vyzo: We do it all over the place, so it doens't hurt to have one more :)
But yeah, at some point we need to figure out a good way to remove the sleeps.

gossipsub.go
vyzo: We don't want to do that because we will reject all new peers and they may not be able to form a connected mesh.
That's why it accepts the peer and resolves (with randomization) during the heartbeat.

pb/rpc.proto
vyzo: There is a reason for this: we don't want to care about the details of the routing state envelope or have to unmarshal it unnecessarily.
Also, we don't want to have a dependency on another protobuf (with import), as this inhibits other languages.

Doing it as bytes _is_ the proper way.

gossipsub.go
vyzo: A minute is probably too high, I would like this to be less than the backoff period so that it can get some new peers before attempting to regraft.
We can probably make it a little longer, with a variable controlling the length.
How about 30s? This should be sufficient, longer connection attempts will probably fail anyway.

vyzo: Also note that as we move to v1.1 the DHT requests should all disappear because we'll have signed routed records for our peers.

vyzo: turned into a variable, with default of 30s.

gossipsub.go
vyzo: Note that the DHT requests should all eventually disappear as gossipsub v1.1 gets deployed, because we'll have signed address record from all peers.

gossipsub.go
vyzo: Just to be clear, sending a PRUNE if we are over the limit is the wrong thing to do, because then the mesh can become full and fail to accept new peers.
(I considered it when designing the algorithm)

gossipsub.go
vyzo: Honestly I don't think this is an issue, and it will greatly complicate code to use a mutex/map combo.

vyzo: Also note that the channel should be drained pretty fast, and that these connection attempts are not all that frequent.

gossipsub.go
aschmahmann: I get that it used to be that way, but is that still true now that we have peer exchange?

If I send back a PRUNE only when I have >Dhi peers and I prune (Dhi-D) peers and tell them about each other won't they always be able to join the mesh?

gossipsub.go
aschmahmann: I agree they'll be drained fast if we have signed peer exchange with good NAT punchthrough. It could get clogged up though if we're doing DHT queries or bad dials.

Frequency is hopefully low, although I wonder if applications might get in trouble during startup. If we use Discover to find more than DHi peers on initialization and connect to them then if the application starts by connecting to 20 topics we may end up with many stalled goroutines.

Since it's not blocking the event loop, I'm ok with a wait + see approach. We may need to keep an eye on this and/or write out a testground test to exercise this

gossipsub.go
vyzo: They probably will be able to do that, but I'd rather reshuffle the mesh in its entirety instead of rejecting new GRAFTs.

gossipsub.go
vyzo: agreed.

gossipsub.go
vyzo: Think about a fully connected mesh where all peers are at `D_hi` -- (unlikely as it is) it won't accept new peers then, while GRAFTing and reshuffling will resolve the situation.

gossipsub.go
vyzo: Also note, many stalled goroutines is not that big of a deal -- the real issue is unbounded goroutine buildup, but I don't think this will happen.

gossipsub.go
aschmahmann: Ah, so IIUC this is really a timing problem right? Because if B is connected to C and then A tries to connect to B it's possible that when B PRUNEs A and C that A will send a GRAFT to C **before** C receives the PRUNE from B.

My concern with the current approach is that if A wants to GRAFT to B it can always keep spamming B and eventually it will get through, even though B has given A plenty of other peers to connect to. This is annoying since if a peer is "better" in some way (e.g. they are the primary publisher) then nodes might be selfish, however it's certainly not a deal-breaker and wasn't even resolvable until the Peer Exchange changes.

Since fixing this would likely require some protocol thought and is less important than the rest of the episub work, seems reasonable to push this further down the road. Would you like me to create an issue?

aschmahmann (approved):
LGTM

gossipsub.go
vyzo: It could be an attack -- I don't think it can happen naturally.

vyzo: Sure, we can create an issue to discuss further instead of it being lost in oblivion after the pr merges.

gossipsub.go
Kubuxu: Hmm, interesting. Ok.

gossipsub.go
```diff
@@ -239,6 +267,90 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+			px := prune.GetPeers()
+			if len(px) > 0 {
+				gs.pxConnect(px)
```
Stebalien, line 278: I have two concerns:

1. What if a single peer sends a _bunch_ of prunes? Could they cause us to launch a ton of goroutines (and crash)?
2. Can't a peer use this to trick us into connecting to a _ton_ of (potentially non-useful) peers? We should only try connecting to the number of peers we actually _need_ (one?).

Ideally, we'd stick these peers in a set of known peers for the topic, connecting to them as-needed whenever we're pruned or we disconnect from a peer.

gossipsub.go
```diff
@@ -441,16 +553,21 @@ func (gs *GossipSubRouter) heartbeat() {
 	tograft := make(map[peer.ID][]string)
 	toprune := make(map[peer.ID][]string)
 
+	// clean up expired backoffs
+	gs.clearBackoff()
```
Stebalien, line 575: This is probably fine, but we should monitor this.

Stebalien (changes requested):
Mostly LGTM but I want to make sure we don't introduce a DoS vector.

gossipsub.go
Stebalien: That is, walking through each of these every heartbeat _should_ be fine, but could get expensive in some edge cases.

gossipsub.go
vyzo: 1. We must be already grafted, we ignore prunes that don't correspond to a peer on our mesh. So a peer can't make us launch an arbitrary number of goroutines by sending us prunes; at most he can make us launch a single goroutine for each topic we belong to its mesh. Not much of a vector I think.

vyzo: 2. We could have a (legitimate) prune listing an arbitrary number of useless peers; we could limit the number of peers we connect to.

gossipsub.go
vyzo: We could do it every few heartbeats instead of every heartbeat.

gossipsub.go
vyzo: For 2 I added a check that limits the number of connections to at most `GossipSubPrunePeers`, so this should address the concern. 
We do want more than 1 peer in general, to expand our potential peers as much as possible.

Stebalien: Got it. SGTM.

gossipsub.go
Stebalien: Every heartbeat is probably fine, I'm just calling it out so we keep it in mind.

Stebalien (approved):

gossipsub.go
```diff
@@ -286,7 +286,12 @@ func (gs *GossipSubRouter) addBackoff(p peer.ID, topic string) {
 }
 
 func (gs *GossipSubRouter) pxConnect(peers []*pb.PeerInfo) {
+	if len(peers) > GossipSubPrunePeers {
+		peers = peers[:GossipSubPrunePeers]
```
Stebalien, line 296: nit: I prefer selecting peers randomly.

vyzo: Sure, let's do random.

vyzo: done.

Stebalien (approved):

gossipsub.go
vyzo: Ok, I slept on it and there _is_ a DoS vector: A malicious peer could send us in sequence GRAFT/PRUNE and cause us to spawn a goroutine; it could simply be sitting there sending GRAFT/PRUNE ad infinum, causing us to spawn a goroutine for each pair.
Granted, the effect is limited in how many goroutines it can fit inside the 30s window for the connection timeout, but it's still nasty.
I will rewrite the connection logic to use a limited set of pending connections and goroutine-free scheduling.

vyzo: Implemented a connection scheduler, which limits the number of max pending connections too.

gossipsub.go
```diff
@@ -57,25 +81,39 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
+	connectMx    sync.Mutex                              // mutex protecting connectPend
+	connectPend  map[peer.ID]*routing.SignedRoutingState // pending px connections
+	connectSched chan struct{}                           // px connection trigger
+	mcache       *MessageCache
+	tracer       *pubsubTracer
+}
+
+type connectInfo struct {
+	p   peer.ID
+	srr *routing.SignedRoutingState
```
Stebalien: This is really cryptic, I had to go back and look at the full patch to figure out what these were. Can we write them out?

gossipsub.go
```diff
@@ -323,16 +333,56 @@ func (gs *GossipSubRouter) pxConnect(peers []*pb.PeerInfo) {
...
+		log.Debug("ignoring peers obtained through px; too many pending connections")
+	}
+	gs.connectMx.Unlock()
+}
+
+func (gs *GossipSubRouter) connectScheduler() {
+	for {
+		gs.connectMx.Lock()
+		var ci connectInfo
+		var pend bool
```
Stebalien: `found`? `pend` doesn't mean anything to me (I know what you're saying, but it's still confusing).

Stebalien (commented):
This works but can't we just have a buffered channel (selecting with default when writing to it)?

Technically, this works _slightly_ better (de duplicates, etc.) but I'm not sure if the complexity is worth it.

vyzo:
Actually you are right; we don't need this complexity, we can just use a sufficient buffer in the connect channel.
I'll reset the last commit and do it with the channel.

Stebalien (approved):

gossipsub.go
raulk: 👍

gossipsub.go
```diff
@@ -239,6 +270,94 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+		peers = peers[:GossipSubPrunePeers]
+	}
+
+	toconnect := make([]connectInfo, 0, len(peers))
+
+	for _, pi := range peers {
+		p := peer.ID(pi.PeerID)
+
+		_, connected := gs.peers[p]
+		if connected {
```
raulk, line 305: So we'll connect to PEX peers but we'll wait until the next heartbeat to rebalance the mesh. That's why we can safely skip topic members that we're already connected to, because we'll anyway consider them in the next heartbeat (as long as we remain connected to them). I think that's correct.

gossipsub.go
raulk: A way more efficient manner would be to use a ring buffer of peer slices, with `nslots = BackoffPeriod/HeartbeatInterval`. Then you track the tick count (monotonically increasing by 1), every heartbeat increments it by 1. And when clearing the backoff, you simply do `currentTick mod nslots`, and simply clear that slot. That's pretty efficient. The data for backoff would turn into `map[string][][]peer.ID`.

gossipsub.go
```diff
@@ -33,6 +37,21 @@ var (
...
 	GossipSubFanoutTTL = 60 * time.Second
+
+	// number of peers to include in prune Peer eXchange
+	GossipSubPrunePeers = 16
+
+	// backoff time for pruned peers
+	GossipSubPruneBackoff = time.Minute
+
+	// number of active connection attempts for peers obtained through px
+	GossipSubConnectors = 16
```
raulk, line 48: Isn't this too much? 16 conn attempts at once, i.e. we'll try to connect to _all_ peers recommended by the pruning one (as this value is equal to `GossipSubPrunePeers`.

I'd prefer if connect was a buffered channel, and we'd have less concurrent goroutines. Right now it's a bit hit or miss, e.g. if we got pruned from two topics at once, the connector goroutines will be saturated with the first batch, and all peers from the second batch will be dropped.

raulk (approved):
This looks pretty great. Approving to avoid stalling any longer, but I'd really appreciate these two items being addressed before merging (see comments):

1. Backoff data structure for more efficient clearing.
2. Buffered channel for connect attempts, and less connector goroutines.

Thanks, @vyzo!

We need to enhance the specs ASAP.

raulk (commented):
We need to merge the pending PRs in go-libp2p-core and go-libp2p-peerstore before we merge this one. Those commit hashes in go.mod are nasty.

gossipsub.go
vyzo: It is a buffered channel! But yes, we can certainly lower from 16, how about 8 or 4?

vyzo:
Re: backoff data structure

I am not entirely convinced it is more efficient with the circular buffer, as we'd have to iterate through the entire ring to find if a peer is being backed off.
What we can do instead, which will improve efficiency, is to only clear backoffs once every few ticks (say every 10-15 ticks for the default 1min backoff).
How does that sound?

gossipsub.go
vyzo: yup, do you want a comment here?

vyzo:
restored yusef's changes that got lost in the rebase (with some modifications for changes in the interface)

gossipsub.go
```diff
@@ -241,6 +272,103 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+func (gs *GossipSubRouter) connector() {
+	for {
+		select {
+		case ci := <-gs.connect:
+			if gs.p.host.Network().Connectedness(ci.p) == network.Connected {
+				continue
+			}
+
+			log.Debugf("connecting to %s", ci.p)
+			cab, ok := peerstore.GetCertifiedAddrBook(gs.p.host.Peerstore())
```
raulk, line 355: Use a `sync.Once` to only set this one per goroutine?

gossipsub.go
```diff
@@ -241,6 +272,103 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+			if ok && ci.spr != nil {
+				_, err := cab.ConsumePeerRecord(ci.spr, peerstore.TempAddrTTL)
```
raulk, line 357: Note: we had to roll back the change in the peerstore that clears all addresses, and blocks subsequent uncertified address additions, once a peer record is seen. It caused downstream regressions (IPFS, and others).

For now we're just accumulating all addresses.

As a definitive solution, I'm enhancing the peerstore with "address labels" such that certified addresses will be tagged as such, and in the `host.Connect()` below, you'll be able to set dial constraints, i.e. "only dial certified addresses".

With this solution, the peerstore tracks all addresses, and it's up to the application to filter which ones it wants to actually use.

gossipsub.go
raulk: Ideally ;-)

gossipsub.go
raulk: There's a recent win here! The decisions of which peers to retain and which to eject is now performed by evaluating the peer's score! 🎉

gossipsub.go
raulk: Max inflight dials FD limit is 160 by default. Assuming each peer has an average of 3 addresses, 16 connectors could make use 30% of our FD allowance. I'd scale this down to 8 by default, but I'd add an option so the app can increase/decrease it (it may also touch the swarm limits!).

Also, we will need some fairness heuristic here. If we're pruned from two or three topics at once, the first topic will get all slots, and the other two will be queued. Instead, we might want to balance peers from topics 1, 2, and 3 for quicker healing. Some form of priority queue could work well.

gossipsub.go
```diff
@@ -241,6 +272,103 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+
+		_, connected := gs.peers[p]
+		if connected {
+			continue
+		}
+
+		var spr *record.Envelope
+		if pi.SignedPeerRecord != nil {
+			// the peer sent us a signed record; ensure that it is valid
+			envelope, r, err := record.ConsumeEnvelope(pi.SignedPeerRecord, peer.PeerRecordEnvelopeDomain)
```
raulk, line 312: We recently added https://godoc.org/github.com/libp2p/go-libp2p-core/record#ConsumeTypedEnvelope, which is much nicer on the allocator (can re-use a single `PeerRecord` instance). May want to move to that at some point.

raulk (approved):

raulk:
@vyzo 

> Re: backoff data structure
> 
> I am not entirely convinced it is more efficient with the circular buffer, as we'd have to iterate through the entire ring to find if a peer is being backed off.
> What we can do instead, which will improve efficiency, is to only clear backoffs once every few ticks (say every 10-15 ticks for the default 1min backoff).
> How does that sound?

Yeah, that sounds reasonable. We'd trade off some heap to save CPU/allocs.

If we really want to nail this, we could maintain a peer+topic index with a refcount, and the circular buffer would be in append-only log form. If we're OK losing precision, every time we add a backoff, we'd INCR the refcount, multiply the new value by the timeout, and "add a mark" on the appropriate position in the buffer (this is not a strict extension).

As entries expire, we'd DECR the refcount, and delete the entry from the index when it reaches 0.

To test whether a peer is currently backed-off, we'd simply test the index.

^^ This is all premature optimisation at this stage, but wanted to record the idea for the future.

raulk:
@vyzo have a look at my comments, and feel free to merge at your discretion.

raulk:
_(wrong button pushed above)_

I think this needs more tests. Particularly the backoff is untested, and maybe consider testing other topology dynamics.

Perhaps @dirkmc or @yusefnapora could take this on in a follow-up PR?

gossipsub.go
vyzo: Not sure if it is all that expensive, it's just a cast isn't it?

vyzo:
rebased on master

gossipsub.go
vyzo: the hardening branch reduces this to 8.

vyzo (commented):
alright, ready to merge.

PR #236, opened 2019-12-02
chore(dep): update

Stebalien:
* Update deps
* Regenerate protobufs with the latest gogo protobuf.


vyzo (commented):
it doesn't seem to update the gogo dependency, has this already been updated?

Stebalien:
Yes, apparently.

vyzo (approved):

PR #240, opened 2019-12-03
coalesce multiple messages into buffer

vyzo:
this should drain the queues faster

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -1,8 +1,8 @@
 package pubsub
 
 import (
-	"bufio"
 	"context"
+	"errors"
 	"io"
 
 	"github.com/libp2p/go-libp2p-core/helpers"
@@ -11,6 +11,7 @@ import (
 
 	ggio "github.com/gogo/protobuf/io"
 	proto "github.com/gogo/protobuf/proto"
+	pool "github.com/libp2p/go-buffer-pool"
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 
 	ms "github.com/multiformats/go-multistream"
@@ -101,18 +102,10 @@ func (p *PubSub) handlePeerEOF(ctx context.Context, s network.Stream) {
 }
 
 func (p *PubSub) handleSendingMessages(ctx context.Context, s network.Stream, outgoing <-chan *RPC) {
-	bufw := bufio.NewWriter(s)
+	bufw := newBufferedWriter(s)
 	wc := ggio.NewDelimitedWriter(bufw)
 
-	writeMsg := func(msg proto.Message) error {
-		err := wc.WriteMsg(msg)
-		if err != nil {
-			return err
-		}
-
-		return bufw.Flush()
-	}
-
+	defer bufw.Release()
 	defer helpers.FullClose(s)
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


raulk (changes requested):
I had a different impression of what we were trying to achieve here. This was my thinking:

1. An RPC arrives. If no write is in progress, we'll initiate the write.
2. A second RPC arrives while the first write is in progress. It remains queued.
3. A third RPC arrives. Instead of queueing it up, we merge it with the head of the queue.
4. We keep an eye on exceeding the pubsub message size limit. When that happens, we insert the exceeding message in the queue, and direct subsequent merges to that one.
5. When the write in progress finishes, we send the merged message in the head of the queue, and the 2nd message becomes the new head.

If we don't do it like that, I don't understand why all elements inside the protobuf are variadic to begin with.

raulk (changes requested):
The logic you're introducing here doesn't gain us much besides a simple write batching behaviour. While an IO write is blocked, elements will still continue to build up and we're back to square one. Yes, they will be released in a batch once the IO write returns, but we're assuming that will happen before the queue saturates, so we're still vulnerable to a *single* slow write (i.e. head of line blocking). I think this solution misses the forest for the trees, and the appropriate solution is to merge RPCs, truly leveraging the variadic nature of the protobufs we designed in the first place.

vyzo:
The effect is the same if we are merging RPCs; we'll have to drain the queue in the same vein.
This achieves exactly the same effect, but without the complexity of merge coalescing.

vyzo:
If you want a background goroutine doing the write, we can achieve exactly the same effect with the buffer.

vyzo:
Note that I don't want to merge the RPCs for a couple of reasons:
- we need to process the control messages and cancel out graft/prune and subscribe/unsubscribe pairs for the same topic.
- we will quickly reach the maximum size limit and stop coalescing anyway.

If the desired behaviour is to drain the queue into memory while still writing and effectively turn it into an unbounded buffer, we can do the same with the write coalescer by writing in a background goroutine.

vyzo:
So upon further discussion on slack, neither coalescing at the write level nor merging RPCs is the right solution to the problem -- we are simply shifting from one queue to the next.

The problem we are trying to address is _transient_ network hiccups, bursts, and persistently slow peers.
Briefly, we want to accommodate temporary network disruption where a peer is transiently slow to read.  We also want to accommodate bursty applications, which send many messages together.
In the same vein we also want to detect peers that are chronically slow and accumulating backlog, and act.

The current best option seems to be to have an adaptive secondary queue that has no hard-cap in number of messages, but instead compares consumption to production rate. If the consumption rate is lower than the production rate for a sustained period of time, then the peer is slow and should be throttled. We probably also want to propagate this fact to the router for further action; for instance, the gossipsub router should remove this peer from the mesh and possibly graylist it.

aschmahmann:
@vyzo this seems like a reasonable thing to do and should help a generic pubsub implementation. However, I wonder if many pubsub users would be able to more efficiently solve the problem of too many messages in the buffer by utilizing a topic-specific message merge function (and being notified of when messages are actually dropped) as I mentioned in https://github.com/libp2p/go-libp2p-pubsub/issues/217#issuecomment-560538413.

Any thoughts on why this approach, which on the surface seems less complex than an adaptive queue, would be insufficient?

vyzo:
We discussed with @raulk and we concluded that merging RPCs is a bad idea; they would have to be completely read before processing any messages, thus increasing the delay and leading to bursts.
Also, the per topic merge function doesn't solve the problem of internally dropped RPCs from router forwarding, it only applies to user published messages iiuc.

raulk:
To enrich the above, our conclusion is that we need to handle unhealthy/slow consumers. We have long talked about this, but now is the time.

Proposal: when we detect a growing backlog that (a) lasts X seconds, or takes up (b) Y bytes (both configurable, sane defaults could be 120 seconds and 2mb), we should terminate the peer's subscription. We'd drain the queue (thus deleting outstanding messages), queue up a `PRUNE` to that peer, and notify the router. Ideally, we'd enhance the PRUNE message with a reason.

aschmahmann:
As I mentioned the proposal seems like a good thing to do in the generic case of a slow consumer - at some point a slow peer is a dead peer.

For the subset of applications that use pubsub to come to agreement on some state (e.g. versioned data like IPNS, a hash chain, etc.) it's possible we can actually just **send less data** (and therefore not have to PRUNE or kill the connection) by allowing the application to merge existing RPCs.

For example, say I want to propagate the latest version of some data structure. If so, then if the old version is stuck in the queue I might be fine just dropping that message. Since the problem is a slow consumer and not a slow producer the extra burden on the producer to do the merge doesn't have to be a problem.

This approach does not only apply to user published messages. For example, in a versioned data system, like IPNS, this is not the case. If A successfully sends messages M1 and M2 to B, but B is blocked sending them to C then B could just drop M1.

raulk:
> If so, then if the old version is stuck in the queue I might be fine just dropping that message.

For this to be possible, we need to introduce "Slow peer" callbacks, cancellable messages, and callbacks for message dispatch.

raulk:
@aschmahmann mind opening issues for the above?

PR #248, opened 2019-12-16
Configurable message id function

### LINKED ISSUE #247: Optional Content addressed messages
Currently message Id's are `source_id + sequence_number`. For my particular use-case (Eth 2), it would be nice if we could optionally set this ID based on the content of message, for example: `hash(message)`. 

This would allow us to filter messages at the pubsub layer based on content rather than source peer. 

I was thinking of adding a configuration parameter to rust-libp2p, which takes a function `f: (source_id, seq_no, message) -> String`to allow the user to specify the message id, given `source_id`, `seq_no`, `message`. 

For my case, this would simply be `hash(message)`

This would be a general way to allow users to specify how messages are addressed. 

I'm open to all suggestions on this and thoughts of feasibility in the go implementation.

protolambda:
See #247: this PR is fully backwards compatible, no public API changes, except for the addition of:
- configuration method ~~`ChangeMsgIdFn`~~ `SetMsgIdFn` to `MessageCache`. The cache doesn't really need a single message id function from the start, so changing it with a method seems fine. And this way it doesn't break any existing code. Note that users of caches like this can update it once they get hold of the exact msg id function to use. E.g. gossipsub updates it when PubSub attaches the router. So it also avoids a painful early design choice of PubSub getting all options, and gossipsub not being able to adapt based on these Options, before initializing PubSub.
- introduction of `MsgIdFunction`, which has the same signature as the previous private `msgID` helper, but now it's configurable! The old `msgID` is now `DefaultMsgIdFn`.
- a PubSub `Option` created with `WithMessageIdFn` to apply your own message ID function to PubSub :)

PS: I thought this was an easy and important issue to just fix it myself, and although I have some previous experience hacking on this repo (I looked deep into its concurrency and simulation for eth2 tooling for aggregation modeling), I can definitely appreciate a review since I don't normally contribute code upstream here.

Closes #247 

cc @AgeManning

### DIFF
--- a/mcache.go
+++ b/mcache.go
@@ -28,6 +28,7 @@ func NewMessageCache(gossip, history int) *MessageCache {
 		msgs:    make(map[string]*pb.Message),
 		history: make([][]CacheEntry, history),
 		gossip:  gossip,
+		msgID:   DefaultMsgIdFn,
 	}
 }
 
@@ -35,6 +36,11 @@ type MessageCache struct {
 	msgs    map[string]*pb.Message
 	history [][]CacheEntry
 	gossip  int
+	msgID   MsgIdFunction
+}
+
+func (mc *MessageCache) ChangeMsgIdFn(msgID MsgIdFunction) {
+	mc.msgID = msgID
 }
 
 type CacheEntry struct {
@@ -43,7 +49,7 @@ type CacheEntry struct {
 }
 
 func (mc *MessageCache) Put(msg *pb.Message) {
-	mid := msgID(msg)
+	mid := mc.msgID(msg)
 	mc.msgs[mid] = msg
 	mc.history[0] = append(mc.history[0], CacheEntry{mid: mid, topics: msg.GetTopicIDs()})
 }
--- a/mcache.go
+++ b/mcache.go
@@ -39,7 +39,7 @@ type MessageCache struct {
 	msgID   MsgIdFunction
 }
 
-func (mc *MessageCache) ChangeMsgIdFn(msgID MsgIdFunction) {
+func (mc *MessageCache) SetMsgIdFn(msgID MsgIdFunction) {
 	mc.msgID = msgID
 }
 
--- a/pubsub.go
+++ b/pubsub.go
@@ -254,6 +254,10 @@ type MsgIdFunction func(pmsg *pb.Message) string
 func WithMessageIdFn(fn MsgIdFunction) Option {
 	return func(p *PubSub) error {
 		p.msgID = fn
+		// the tracer Option may already be set. Update its message ID function to make options order-independent.
+		if p.tracer != nil {
+			p.tracer.msgID = fn
+		}
 		return nil
 	}
 }
--- a/pubsub.go
+++ b/pubsub.go
@@ -117,6 +117,9 @@ type PubSub struct {
 	seenMessagesMx sync.Mutex
 	seenMessages   *timecache.TimeCache
 
+	// function used to compute the ID for a message
+	msgID MsgIdFunction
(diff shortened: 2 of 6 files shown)


protolambda:
Edit: force pushed over previous commit, for some reason Go decided I wanted the old multi addr dependency version in here. Removed that from the diff.

mcache.go
```diff
@@ -28,13 +28,19 @@ func NewMessageCache(gossip, history int) *MessageCache {
...
 }
 
 type MessageCache struct {
 	msgs    map[string]*pb.Message
 	history [][]CacheEntry
 	gossip  int
+	msgID   MsgIdFunction
+}
+
+func (mc *MessageCache) ChangeMsgIdFn(msgID MsgIdFunction) {
```
vyzo: let's call this `SetMsgIdFn`, as it is always called to set it.

vyzo (commented):
LGTM modulo a small nit regarding the naming of `ChangeMsgIdFn`.

pubsub.go
```diff
@@ -326,7 +344,7 @@ func WithDiscovery(d discovery.Discovery, opts ...DiscoverOpt) Option {
 // WithEventTracer provides a tracer for the pubsub system
 func WithEventTracer(tracer EventTracer) Option {
 	return func(p *PubSub) error {
-		p.tracer = &pubsubTracer{tracer: tracer, pid: p.host.ID()}
+		p.tracer = &pubsubTracer{tracer: tracer, pid: p.host.ID(), msgID: p.msgID}
```
vyzo, line 351: one subtlety here is that the message ID option needs to be passed before the tracer option, otherwise the two will diverge.

vyzo:
We have a small issue with the ordering of options regarding the message id function, as seen by the tracer.
I think we might have to modify the tracer inline when present in the option.

pubsub.go
```diff
@@ -240,6 +244,20 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
+// implementation of this function by configuring it with the Option from WithMessageIdFn.
+type MsgIdFunction func(pmsg *pb.Message) string
+
+// WithMessageIdFn is an option to customize the way a message ID is computed for a pubsub message.
+// The default ID function is DefaultMsgIdFn (concatenate source and seq nr.),
+// but it can be customized to e.g. the hash of the message.
+func WithMessageIdFn(fn MsgIdFunction) Option {
+	return func(p *PubSub) error {
+		p.msgID = fn
+		return nil
```
vyzo: In order to resolve the tracer id function issue, let's add this code:
```
if p.tracer != nil {
  p.tracer.msgID = fn
}
```

This will allow us to pass the options in any order.

mcache.go
protolambda: :+1: See 6bd07a7

pubsub.go
protolambda: Good catch, see 6bd07a7 for your suggested fix.

pubsub.go
protolambda: :+1: see 6bd07a7

vyzo (approved):
thank you!

raulk:
May want to make a release for @protolambda and @prestonvanloon to be able to test this change against rust.

vyzo:
Yeah, release forthcoming.

Nashatyrev:
Is it the case that Go gossip implementation lacks client validation functionality?
@vyzo may be you can comment here https://github.com/ethereum/eth2.0-specs/issues/1528#issuecomment-566540862 ?

PR #257, opened 2020-01-23
Update peer-exchange branch with PeerRecord api changes

yusefnapora:
Hey @vyzo, this PR is based on the branch for #234 and just updates it to work with the various name changes that have happened to the signed peer records PR (https://github.com/libp2p/go-libp2p-core/pull/73).

let me know if you have any questions & also feel free to comment on the -core PR if you think we need to change the API at all before merge.


gossipsub.go
```diff
@@ -3,6 +3,7 @@ package pubsub
 import (
 	"context"
 	"fmt"
+	"github.com/libp2p/go-libp2p-core/record"
```
vyzo: can you put this together with our other imports? Here it is mixed with system imports.

gossipsub.go
```diff
@@ -820,21 +828,24 @@ func (gs *GossipSubRouter) makePrune(p peer.ID, topic string) *pb.ControlPrune {
 		return p != xp
 	})
 
+	cab, peerstoreSupportsSignedAddrs := peerstore.GetCertifiedAddrBook(gs.p.host.Peerstore())
```
vyzo: can we call the variable something shorter? This is an eyesore.

gossipsub.go
```diff
@@ -304,19 +304,23 @@ func (gs *GossipSubRouter) pxConnect(peers []*pb.PeerInfo) {
...
 				log.Warningf("error unmarshalling routing record obtained through px: %s", err)
 				continue
 			}
-			if srr.PeerID != p {
-				log.Warningf("bogus routing record obtained through px: peer ID %s doesn't match expected peer %s", srr.PeerID, p)
+			rec, ok := r.(*peer.PeerRecord)
+			if !ok {
+				log.Warnf("bogus routing record obtained through px: envelope payload is not PeerRecord")
+			}
+			if rec.PeerID != p {
```
vyzo: there is a control flow issue here; if `!ok` then `rec.PeerID` is a nil pointer dereference, which could crash.

yusefnapora: thanks for catching that 👍

vyzo (approved):

Issue #209, opened 2019-10-13
Allow per-topic message size limits


whyrusleeping:
I'd like to be able to configure a maximum message size per topic. For example, filecoin transactions should never be larger than 32k, but block messages might be larger than that. 

I don't want to have to read the entire object from the other peer into memory before i drop it, I should be able to detect the message is too big, and then either kill the connection, or discard the bytes as they come in.

aschmahmann:
Seems reasonable. While not necessary it seems like we should get #189 resolved first though.

aschmahmann:
An issue that @aarshkshah1992 brought up is that the PubSub protocol currently only has one stream per peer. This means that if topic A has a 1MB max and topic B has a 1KB max that the stream potentially has to read 1MB for every RPC that comes in, since it might be a topic A message.

There are a number of possible solutions, but here are a few:
1. Have one stream per topic instead of peer
2. Modify the PubSub wire protocol so that the topic arrives first and can be deserialized before the rest of the message
3. Have the limits be configurable on a PubSub level instead of a topic level
4. Have the limit be the max of the topic sizes
5. Read the oversized message (e.g. topic size is 1k, but PubSub size is 1M and we read 1M) then if it's bigger than the topic size kill the connection and blacklist the peer

Since options 1 and 2 require protocol changes doing a combination of options 3 and 5 seems reasonable, although we may want to consider taking 1 and 2 into account for future PubSub protocol iterations.

@whyrusleeping @raulk what do you think?

PR #261, opened 2020-02-26
make wire protocol message size configurable.

raulk:
Relates to #209, but doesn't close it because this implements a configurable global maximum message size.

This PR introduces the `WithMaxMessageSize(size int)` functi onal option that can be used in the pubsub constructor. It also introduces a unit test, and detailed godocs for this feature.


raulk:
cc @vishalchangraniaxiom

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/libp2p/go-libp2p-pubsub
 
 require (
 	github.com/gogo/protobuf v1.3.1
-	github.com/hashicorp/golang-lru v0.5.4
+	github.com/hashicorp/golang-lru v0.5.4 // indirect
```
aschmahmann, line 5: Is this an artifact from your local branch or do we need to change the dependency to indirect anyway?

floodsub_test.go
```diff
@@ -1119,3 +1119,46 @@ func TestMessageSender(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	hosts := getNetHosts(t, ctx, 10)
+
+	// use a 4mb limit; default is 1mb; we'll test with a 2mb payload.
+	psubs := getPubsubs(ctx, hosts, WithMaxMessageSize(1<<22))
+
+	sparseConnect(t, hosts)
+	time.Sleep(time.Millisecond * 100)
```
aschmahmann, line 1133: These make me sad and they're all over this code base. While we could certainly improve the API (e.g. by allowing a Readiness check even without a dummy discovery mechanism) we can already remove the sleep by following the pattern in this test.

https://github.com/libp2p/go-libp2p-pubsub/blob/a1999db8934dc4320aaaa578aed05873113f1a04/topic_test.go#L122

https://github.com/libp2p/go-libp2p-pubsub/blob/a1999db8934dc4320aaaa578aed05873113f1a04/topic_test.go#L145

aschmahmann (approved):
LGTM, although I'd like to give @vyzo an opportunity to verify he's ok with this change. This seems to be a solution to a request made at https://github.com/libp2p/specs/issues/118#issuecomment-560607776

Stebalien (approved):

raulk:
Enough approvals; merging.

PR #263, opened 2020-03-02
Implementation of gossipsub v1.1 - Hardening

vyzo:
This PR implements the Hardening extensions as defined by [Gossipsub v1.1 Specification](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md)

# What is left to do
- [x] plug peer score into gossipsub algorithms
- [x] flood publishing option
- [x] adaptive gossip disemmination
- [x] (re)transmission limit for IWANT requests
- [x] fill in the peer score calculation details
- [x] Increase the code coverage with unit tests (#281 #287)
- [x] Testground testing


gossipsub.go
```diff
@@ -180,8 +210,14 @@ func (gs *GossipSubRouter) HandleRPC(rpc *RPC) {
 }
 
 func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.ControlIWant {
-	iwant := make(map[string]struct{})
+	// we ignore IHAVE gossip from any peer whose score is below the gossip threshold
+	score := gs.score.Score(p)
+	if score < gs.gossipThreshold {
```
willscott, line 357: From the protocol description, I was expecting some recording of metadata from lower-reputation peers, so that data can be learned from them if still needed as a fallback when the peers above the gossip threshold fail. Does that happen somewhere else?

vyzo: The gossip threshold is generally negative, so that we allow for some gossip disemination from low scoring peers. If the score drops below that, the peer is effectively graylisted until his negative score decays (if it does that).
So what you are asking for is exactly what is happening.

vyzo:
Summoning @yusefnapora @raulk @Stebalien @Kubuxu @whyrusleeping 

This is ready for review (modulo unit tests).

pubsub.go
```diff
@@ -356,7 +361,11 @@ func WithDiscovery(d discovery.Discovery, opts ...DiscoverOpt) Option {
 // WithEventTracer provides a tracer for the pubsub system
 func WithEventTracer(tracer EventTracer) Option {
 	return func(p *PubSub) error {
-		p.tracer = &pubsubTracer{tracer: tracer, pid: p.host.ID(), msgID: p.msgID}
+		if tracer != nil {
+			p.tracer.tracer = tracer
```
yusefnapora, line 365: I think `p.tracer` may be nil here - I get a nil-dereference when I pass a tracer into this branch, but it works if I pass nil

vyzo: oops, wrong check. it should be `if p.tracer != nil`....

fixing.

vyzo: fixed.

pubsub.go
```diff
@@ -775,6 +784,12 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 		}
 	}
 
+	// ask the router to vet the peer before commiting any processing resources
+	if !p.rt.AcceptFrom(rpc.from) {
+		log.Warningf("received message from router graylisted peer %s. Dropping RPC", rpc.from)
```
whyrusleeping, line 789: should we do something if we get a bunch of these?

gossipsub.go
```diff
@@ -384,9 +536,18 @@ func (gs *GossipSubRouter) Publish(from peer.ID, msg *pb.Message) {
 			continue
 		}
 
+		if gs.floodPublish && from == gs.p.host.ID() {
```
whyrusleeping, line 636: so this is adding in the flood publish option, flooding IFF its from us.

pubsub.go
vyzo: we could blacklist i guess, but then we'd have to count stuff.

gossipsub.go
vyzo: yes, otherwise it would be floodsub :p

score.go
```diff
@@ -0,0 +1,885 @@
...
+}
+
+// router interface
+func (ps *peerScore) Start(gs *GossipSubRouter) {
+	ps.msgID = gs.p.msgID
+	ps.host = gs.p.host
+	go ps.background(gs.p.ctx)
+}
+
+func (ps *peerScore) Score(p peer.ID) float64 {
```
whyrusleeping, line 157: As far as I can tell, this gets called for every single message received. It seems to me like it may get rather resource intensive. Let's keep this in mind and observe CPU profiles during tests

vyzo: Yes, the `Score` method gets called a lot. It should be relatively lightweight to compute, but good point about looking at CPU profiles.

score.go
```diff
@@ -0,0 +1,885 @@
...
+func (p *PeerScoreParams) validate() error {
+	for _, params := range p.Topics {
+		err := params.validate()
+		if err != nil {
+			return err
+		}
+	}
+
+	// check that we have an app specific score; the weight can be anything (but expected positive)
+	if p.AppSpecificScore == nil {
```
schomatis: Let's also check `AppSpecificWeight` here, or maybe even remove it altogether, as this is an app-controlled function that can regulate its own score values (in contrast with the other weights that modulate pubsub behavior).

gossipsub.go
```diff
@@ -238,17 +351,44 @@ func (gs *GossipSubRouter) handleIWant(p peer.ID, ctl *pb.ControlMessage) []*pb.
...
 		peers, ok := gs.mesh[topic]
 		if !ok {
+			// don't do PX when there is an unknown topic to avoid leaking our peers
+			doPX = false
+			// spam harndening: ignore GRAFTs for unknown topics
+			continue
+		}
+
+		// check the score
+		if score < 0 {
```
schomatis, line 463: In general, from a security perspective, I would prefer if all the `score` validation code would happen at a single point, similar to what is being done in `AcceptFrom`. This is not a blocker for me because I understand that it's not as straightforward as that other case, here different handlers will react slightly differently to a blacklisted (or similar) peer.

Citing this particular line as an example of the rationale behind this comment (but this logic replicates in other places). If `score < 0` then we basically won't be doing any GRAFT or PX (seems we won't be doing much besides the prune control response). This is not at all evident buried here, and it makes it harder to audit and maintain, even if it's correct (which seems to be).

Ideally handlers like these one should deal with the positive logic, what actually gets done for accepted peers, and move all security checks upstream. Again, just a suggestion because there may be other repercussions for the proposed change I'm not seeing.

gossipsub.go
```diff
@@ -70,6 +77,62 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
+// gossipThreshold is the score threshold below which gossip propagation is supressed.
+//
+// publishThreshold is the score threshold below which we shouldn't publish when using flood
+//  publishing (also applies to fanout and floodsub peers).
+//
+// graylistThreshold is the score threshold below which message processing is supressed altogether,
+//  implementing an effective graylist according to peer score.
+//
+// These thresholds should generally be negative, allowing some information to disseminate from low
+//  scoring peers.
```
schomatis: Let's add checks for these and emit a warning please, e.g., it seems that almost always we would want a graylist threshold below 0 and any value above this range will likely be a programmer error.

schomatis: From reading the spec it seems we should also check that gossip > publish > graylist (especially because this relation doesn't evident from reading the code).

gossipsub.go
```diff
@@ -103,9 +182,19 @@ func (gs *GossipSubRouter) Protocols() []protocol.ID {
 func (gs *GossipSubRouter) Attach(p *PubSub) {
 	gs.p = p
 	gs.tracer = p.tracer
+
+	// start the scoring, if any
+	if gs.score != nil {
```
schomatis: Could we invert this and turn it on by default with some off-the-shelf reasonable values and make the consumer explicitly disable it if needed?

gossipsub.go
```diff
@@ -570,22 +746,47 @@ func (gs *GossipSubRouter) heartbeat() {
...
 		if len(peers) < GossipSubDlo {
 			backoff := gs.backoff[topic]
 			ineed := GossipSubD - len(peers)
 			plst := gs.getPeers(topic, ineed, func(p peer.ID) bool {
-				// filter our current peers and peers we are backing off
+				// filter our current peers, peers we are backing off, and peers with negative score
 				_, inMesh := peers[p]
 				_, doBackoff := backoff[p]
-				return !inMesh && !doBackoff
+				score := gs.score.Score(p)
```
schomatis: Could we reuse the already computed scores in `scores` here?

gossipsub.go
```diff
@@ -630,10 +833,11 @@ func (gs *GossipSubRouter) heartbeat() {
 
 	// maintain our fanout for topics we are publishing but we have not joined
 	for topic, peers := range gs.fanout {
-		// check whether our peers are still in the topic
+		// check whether our peers are still in the topic and have a score above the publish threshold
 		for p := range peers {
 			_, ok := gs.p.topics[topic][p]
-			if !ok {
+			score := gs.score.Score(p)
```
schomatis: Same.

score.go
vyzo: That's a good suggestion, we can consider removing the app specific weight and just keep the score function.

gossipsub.go
vyzo: Yeah, I'll add a check in validate.

gossipsub.go
vyzo: I can just move the nil check inside the `Start` implementation so that it does not occlude the logic.

gossipsub.go
vyzo: no, we have scores for peers _in_ the mesh, not the other ones. But the code should short-circuit and avoid calling `Score` if `inMesh` or `doBackoff` is true.

gossipsub.go
vyzo: again here we can short-circuit.

score.go
```diff
@@ -0,0 +1,885 @@
...
+	drec.status = delivery_invalid
+	for p := range drec.peers {
+		ps.markInvalidMessageDelivery(p, msg)
+	}
+
+	// release the delivery time tracking map to free some memory early
+	drec.peers = nil
+}
+
+func (ps *peerScore) DuplicateMessage(msg *Message) {
```
yusefnapora, line 573: @vyzo I'm getting some "assignment to nil map" errors running against this branch in testground, e.g.

```
   2.0120s      ERROR << c694df9af4d2 >> panic: assignment to entry in nil map
   2.0121s      ERROR << c694df9af4d2 >>
   2.0121s      ERROR << c694df9af4d2 >> goroutine 98 [running]:
   2.0121s      ERROR << c694df9af4d2 >> github.com/libp2p/go-libp2p-pubsub.(*peerScore).DuplicateMessage(0xc000091340, 0xc00025ebd0)
   2.0122s      ERROR << c694df9af4d2 >> 	/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.2.7-0.20200311004457-925ed3894d0b/score.go:653 +0x14f
   2.0122s      ERROR << c694df9af4d2 >> github.com/libp2p/go-libp2p-pubsub.(*pubsubTracer).DuplicateMessage(0xc000091300, 0xc00025ebd0)
   2.0122s      ERROR << c694df9af4d2 >> 	/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.2.7-0.20200311004457-925ed3894d0b/trace.go:106 +0x3a1
   2.0122s      ERROR << c694df9af4d2 >> github.com/libp2p/go-libp2p-pubsub.(*validation).validate(0xc000389290, 0x0, 0x0, 0x0, 0xc0000a8f90, 0x26, 0xc00025ebd0)
   2.0123s      ERROR << c694df9af4d2 >> 	/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.2.7-0.20200311004457-925ed3894d0b/validation.go:207 +0x6ad
   2.0133s      ERROR << c694df9af4d2 >> github.com/libp2p/go-libp2p-pubsub.(*validation).validateWorker(0xc000389290)
   2.0133s      ERROR << c694df9af4d2 >> 	/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.2.7-0.20200311004457-925ed3894d0b/validation.go:184 +0x7c
   2.0133s      ERROR << c694df9af4d2 >> created by github.com/libp2p/go-libp2p-pubsub.(*validation).Start
   2.0134s      ERROR << c694df9af4d2 >> 	/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.2.7-0.20200311004457-925ed3894d0b/validation.go:98 +0x70
   2.3247s INCOMPLETE << c694df9af4d2 >>
```

This must be because `drec.peers` was released in an earlier `RejectMessage` call. Can you add a nil check here for `drec.peers` and exit early if it's already been rejected? Or do we actually want to keep tracking duplicate deliveries after we've rejected the message?

vyzo: hrm, we do want to keep track of rejected duplicates, but I am not sure this is the case here.
Are you generating any rejected messages?

vyzo: It's in a valid delivery.

vyzo: I think we might have some throttling going on which can also set the peers map to nil.

vyzo: ie, it is possible to start validation and also throttle validation of a duplicate because it can't push to the validation workers.

vyzo: yeah, that's what I think is happening: the validation queue throttles before we enter validation, which rejects the message (with a message ID of unknown validity) and sets the peer map to nil.
In the meantime there is another validation happening that completes successfully, and boom.
Anyhow, a nasty little bug.

vyzo: it should be fixed in https://github.com/libp2p/go-libp2p-pubsub/pull/263/commits/5f1b070d5922dc076572cbd03010766a6f8b89cf

vyzo:
also merged in #268 as this fixes a tracing bug.

score.go
Kubuxu: We also should watch lock contention there.

gossipsub.go
vyzo: I short-circuited the score calculation to only occur if `!inMesh && !doBackoff`.

gossipsub.go
vyzo: done.

gossipsub.go
vyzo: I am not sure there we can come up with reasonable off-the-shelf defaults, let me think about it.

vyzo: for now I have moved this check inside `Start` so that it doesn't occlude the logic.

gossipsub.go
vyzo: done.

gossipsub.go
vyzo: it's a matter of some code duplication and convenience, but we can certainly check upfront here.

vyzo: ok, I tried it and the code gets convoluted and requires a goto. Not a very good tradeoff.

gossipsub.go
vyzo: So I have given it some more thought and I think we can make a gossipsub v1.1 constructor function that can fill in default parameters given the expected message rate and the application score function.

vyzo:
Add a `WithPeerScoreDebug` option to enable periodic inspection/dumping of the peer scores.

gossipsub.go
schomatis: @vyzo Why would you need a goto? Could you push that code in another branch please? I think I may have not expressed this correctly and we may thinking of different refactorings.

vyzo:
Some improvements of the backoff handling:
- we now extend the backoff when pruning because of backoff or a negative score; we also check the exact expiration.
- backoff is cleaned up every 15 heartbeat ticks to amortize the cost of iterating over the maps.

gossipsub.go
```diff
@@ -238,17 +368,49 @@ func (gs *GossipSubRouter) handleIWant(p peer.ID, ctl *pb.ControlMessage) []*pb.
...
+			doPX = false
+			// spam harndening: ignore GRAFTs for unknown topics
+			continue
+		}
+
+		// check the score
+		if score < 0 {
+			// we don't GRAFT peers with negative score
+			log.Debugf("GRAFT: ignoring peer %s with negative score [score = %f, topic = %s]", p, score, topic)
+			// we do send them PRUNE however, because it's a matter of protocol correctness
```
Stebalien, line 466: FYI, this does leak information and immediately tells the attacker that they failed and should try again.

gossipsub.go
Stebalien: Do messages ever expire? If not, can't an attacker:

1. Collect a set of messages from us.
2. Wait for us to "forget" we sent the messages.
3. Feed us our own messages back.

Causing us to repeatedly flood our own messages out to the network (assuming they've collected enough of them)?

Stebalien: We should drop our own messages if we receive them from the network (although I'm still concerned about the general replay issue).

gossipsub.go
```diff
@@ -434,12 +619,20 @@ func (gs *GossipSubRouter) Join(topic string) {
 
 	gmap, ok = gs.fanout[topic]
 	if ok {
+		// these peers have a score above the publish threshold, which may be negative
+		// so drop the ones with a negative score
+		for p := range gmap {
+			if gs.score.Score(p) < 0 {
+				delete(gmap, p)
+			}
+		}
```
Stebalien, line 712: Why do this _here_? To avoid having to eagerly maintain `gs.fanout`?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// P6: IP-colocation factor.
+	// The parameter has an associated counter which counts the number of peers with the same IP.
+	// If the number of peers in the same IP exceeds IPColocationFactorThreshold, then the value
+	// is the square of the difference, ie (PeersInSameIP - IPColocationThreshold)^2.
+	// If the number of peers in the same IP is less than the threshold, then the value is 0.
+	// The weight of the parameter MUST be negative, unless you want to disable for testing.
+	// Note: In order to simulate many IPs in a managable manner when testing, you can set the weight to 0
+	//       thus disabling the IP colocation penalty.
+	IPColocationFactorWeight    float64
+	IPColocationFactorThreshold int
```
Stebalien: Note: this doesn't work well for IPv6 at all so I'm not sure how useful it is in general. I assume an attacker would just disable IPv4 and pick IPv6 addresses out of their ISP assigned address range at will. ISPs are supposed to give out 64bit ranges.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	for _, c := range conns {
+		remote := c.RemoteMultiaddr()
+
+		ip4, err := remote.ValueForProtocol(ma.P_IP4)
+		if err == nil {
+			res = append(res, ip4)
+			continue
+		}
+
+		ip6, err := remote.ValueForProtocol(ma.P_IP6)
```
Stebalien: Note: this function is dangerous as it'll pick out the first IP address, regardless of what comes first. I can't think of a case where it'll fail at the moment, but it's best to avoid it and call something like `manet.ToIP`.

score.go
```diff
@@ -0,0 +1,952 @@
...
+			ps.markDuplicateMessageDelivery(p, msg, time.Time{})
+		}
+	}
+}
+
+func (ps *peerScore) RejectMessage(msg *Message, reason string) {
+	ps.Lock()
+	defer ps.Unlock()
+
+	// TODO: the reasons should become named strings; good enough for now.
```
Stebalien: Why not just fix it now?

score.go
```diff
@@ -0,0 +1,952 @@
...
+		// it has a valid message ID; all we can do is ignore it.
+		return
+	}
+
+	drec := ps.deliveries.getRecord(ps.msgID(msg.Message))
+
+	if reason == "validation throttled" {
+		// if we reject with "validation throttled" we don't penalize the peer(s) that forward it
+		// because we don't know if it was valid.
+		drec.status = delivery_throttled
```
Stebalien: nit: non-idiomatic go variable name.

gossipsub.go
vyzo: They will be backed off and if they try repeatedly immediately after they will never get in because they will keep extending their backoff period.
Otherwise they will have to wait at least 1min to try again.

I think we should still emit the `PRUNE` message here, because it's a matter of protocol correctness.

Consider an honest node that happens to have a negative score (eg because of staked participation and them not having put in the stake yet, or because it was in the mesh and got pruned because it was not delivering enough messages). If we don't emit a PRUNE here they will think they are in the mesh while they are not and fail to build a mesh.

gossipsub.go
vyzo: That's a good suggestion, we should never re-broadcast our own messages from the network.

gossipsub.go
vyzo: yes indeed.

score.go
vyzo: We can use `/64`s for IPv6 addresses, that's an easy fix.

score.go
vyzo: ok, can do.

score.go
vyzo: sure, I can fix it now.

score.go
vyzo: it's an unexported constant.

score.go
vyzo: In general it is _very_ useful as it makes it very hard to launch a sybil attack using just a few IP addresses (which is something that has been observed in the filecoin testnet).

vyzo: I think we need to track both the IP and the /64 in the case of IPv6.
We can also add a whitelist for eg datacenters, so that those /64 don't get penalized.

gossipsub.go
vyzo: So the simplest ways to implement is to check if `ReceivedFrom` also matches our own host ID.
If it doesn't and it claims to be from us, we can drop it.

gossipsub.go
```diff
@@ -70,6 +77,74 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
+//
+// publishThreshold is the score threshold below which we shouldn't publish when using flood
+//  publishing (also applies to fanout and floodsub peers).
+//
+// graylistThreshold is the score threshold below which message processing is supressed altogether,
+//  implementing an effective graylist according to peer score.
+//
+// These thresholds should generally be negative, allowing some information to disseminate from low
+//  scoring peers.
+func WithPeerScore(params *PeerScoreParams, gossipThreshold, publishThreshold, graylistThreshold float64) Option {
```
raulk: Can we encapsulate the thresholds within a struct too? That would make things more extensible in the future.

BTW – I like the separation between params that drive peer score calculation, and params that drive compensatory actions.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	meshMessageDeliveriesActive bool
+
+	// sticky mesh rate failure penalty counter
+	meshFailurePenalty float64
+
+	// invalid message counter
+	invalidMessageDeliveries float64
+}
+
+type peerScore struct {
```
raulk, line 58: Godocs here would be much appreciated 🙏 Describe the role of this struct, when/where it's used, and it's locking requirements (contention?), would be great.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	defer gcDeliveryRecords.Stop()
+
+	var inspectScores <-chan time.Time
+	if ps.inspect != nil {
+		ticker := time.NewTicker(ps.inspectPeriod)
+		defer ticker.Stop()
+		inspectScores = ticker.C
+	} else {
+		// never fires
+		inspectScores = make(chan time.Time)
```
raulk: nil would work as well.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	defer refreshScores.Stop()
+
+	refreshIPs := time.NewTicker(time.Minute)
+	defer refreshIPs.Stop()
+
+	gcDeliveryRecords := time.NewTicker(time.Minute)
+	defer gcDeliveryRecords.Stop()
+
+	var inspectScores <-chan time.Time
+	if ps.inspect != nil {
```
raulk, line 267: I would probably be useful to `defer` an inspection, so that we dump the scores one last time when the background process ends.

score.go
```diff
@@ -0,0 +1,952 @@
...
+}
+
+type TopicScoreParams struct {
+	// The weight of the topic.
+	TopicWeight float64
+
+	// P1: time in the mesh
+	// This is the time the peer has ben grafted in the mesh.
+	// The value of of the parameter is the time/TimeInMeshQuantum, capped by TimeInMeshCap
+	// The weight of the parameter MUST be positive.
```
raulk: or zero.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	TimeInMeshWeight  float64
+	TimeInMeshQuantum time.Duration
+	TimeInMeshCap     float64
+
+	// P2: first message deliveries
+	// This is the number of message deliveries in the topic.
+	// The value of the parameter is a counter, decaying with FirstMessageDeliveriesDecay, and capped
+	// by FirstMessageDeliveriesCap.
+	// The weight of the parameter MUST be positive.
```
raulk: or zero.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// P3b: sticky mesh propagation failures
+	// This is a sticky penalty that applies when a peer gets pruned from the mesh with an active
+	// mesh message delivery penalty.
+	// The weight of the parameter MUST be negative (or zero if you want to disable it)
+	MeshFailurePenaltyWeight, MeshFailurePenaltyDecay float64
+
+	// P4: invalid messages
+	// This is the number of invalid messages in the topic.
+	// The value of the parameter is a counter, decaying with InvalidMessageDeliveriesDecay.
+	// The weight of the parameter MUST be negative.
```
raulk: or zero.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// time to remember counters for a disconnected peer.
+	RetainScore time.Duration
+}
+
+type TopicScoreParams struct {
+	// The weight of the topic.
+	TopicWeight float64
+
+	// P1: time in the mesh
+	// This is the time the peer has ben grafted in the mesh.
```
raulk: been.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// The value of of the parameter is the time/TimeInMeshQuantum, capped by TimeInMeshCap
+	// The weight of the parameter MUST be positive.
+	TimeInMeshWeight  float64
+	TimeInMeshQuantum time.Duration
+	TimeInMeshCap     float64
+
+	// P2: first message deliveries
+	// This is the number of message deliveries in the topic.
```
raulk: Can we specify why the "first" adjective?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// The value of the parameter is a counter, decaying with FirstMessageDeliveriesDecay, and capped
```
raulk: FirstMessageDeliveriesDecay seems like a decay factor. Can we specify what the decay function is?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	FirstMessageDeliveriesWeight, FirstMessageDeliveriesDecay float64
+	FirstMessageDeliveriesCap                                 float64
+
+	// P3: mesh message deliveries
+	// This is the number of message deliveries in the mesh, within the MeshMessageDeliveriesWindow of
+	// message validation; deliveries during validation also count and are retroactively applied
+	// when validation succeeds.
+	// This window accounts for the minimum time before a hostile mesh peer trying to game the score
+	// could replay back a valid message we just sent them.
+	// It effectively tracks first and near-first deliveries, ie a message seen from a mesh peer
```
raulk: Expand on "near-first"?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	// counter value below which it is considered 0.
+	DecayToZero float64
+
+	// time to remember counters for a disconnected peer.
+	RetainScore time.Duration
+}
+
+type TopicScoreParams struct {
+	// The weight of the topic.
+	TopicWeight float64
```
raulk: Does the sum of weights have to be 1?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+type PeerScoreParams struct {
```
raulk: Can we move the two Params structs to a `score_params.go` file?

score.go
```diff
@@ -0,0 +1,952 @@
...
+	delivery_throttled        // we can't tell if it is valid because validation throttled
+)
+
+type PeerScoreInspectFn func(map[peer.ID]float64)
+
+// WithPeerScoreDebug is a gossipsub router option that enables peer score debugging.
+// When this option is enabled, the supplied function will be invoked periodically to allow
+// the application to inspec or dump the scores for connected peers.
+// This option must be passed _after_ the WithPeerScore option.
+func WithPeerScoreDebug(inspect PeerScoreInspectFn, period time.Duration) Option {
```
raulk: nit: homogeneous naming. `WithPeerScoreInspect`? Or `PeerScoreDumpFn`/`WithPeerScoreDump`?

score.go
```diff
@@ -0,0 +1,952 @@
...
+		msgID:      DefaultMsgIdFn,
+	}
+}
+
+// peer score parameter validation
+func (p *PeerScoreParams) validate() error {
+	for _, params := range p.Topics {
+		err := params.validate()
+		if err != nil {
+			return err
```
raulk: Can we wrap this error to annotate with the topic name that failed validation?

vyzo:
rebased and regargeted on master as #234 has merged.

score.go
```diff
@@ -0,0 +1,952 @@
...
+		return 0
+	}
+
+	ps.Lock()
+	defer ps.Unlock()
+
+	return ps.score(p)
+}
+
+func (ps *peerScore) score(p peer.ID) float64 {
```
raulk, line 168: Eventually, we might find that we'd benefit from caching the score, and recalculating it in batch every N ticks (resolution), under a global lock. Recalculating the score on access may be heavy and may introduce minor skews that may create subtle unfairness across peers based on access time.

score.go
```diff
@@ -0,0 +1,952 @@
...
+		// the topic parameters
+		topicParams, ok := ps.params.Topics[topic]
+		if !ok {
+			// we are not scoring this topic
+			continue
+		}
+
+		// P1: time in Mesh
+		if tstats.inMesh {
+			p1 := float64(tstats.meshTime / topicParams.TimeInMeshQuantum)
```
raulk, line 190: We should globally cap the sum total as well. Else, an attacker could boost their score artificially by meshing with us on many topics, and staying on those for a long time.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	p5 := ps.params.AppSpecificScore(p)
+	score += p5 * ps.params.AppSpecificWeight
+
+	// P6: IP collocation factor
+	for _, ip := range pstats.ips {
+		peersInIP := len(ps.peerIPs[ip])
+		if peersInIP > ps.params.IPColocationFactorThreshold {
+			surpluss := float64(peersInIP - ps.params.IPColocationFactorThreshold)
+			p6 := surpluss * surpluss
+			score += p6 * ps.params.IPColocationFactorWeight
```
raulk, line 248: Comment: this weight is negative -- it is a penalty score.

score.go
```diff
@@ -0,0 +1,952 @@
...
+	for p, pstats := range ps.peerStats {
+		if !pstats.connected {
+			// has the retention period expired?
+			if now.After(pstats.expire) {
+				// yes, throw it away (but clean up the IP tracking first)
+				ps.removeIPs(p, pstats.ips)
+				delete(ps.peerStats, p)
+			}
+
+			// we don't decay retained scores, as the peer is not active.
```
raulk, line 327: Can a peer disconnect and connect seconds before the retention period elapses, in order to keep their score "alive" perpetually? We may want to introduce additional measures here, e.g.:

1. Tallying periods of connectedness/disconnectedness across interruptions. When the total disconnectedness time surprasses a threshold, we forget the peer's score.
2. Tracking absolute number of disconnections and penalising for interruptions, relative to interruptions from other peers => rationale: if we have a high interruption rate from all peers, we might be the ones that are unstable and don't want to penalise anybody because we'll be shooting ourselves in the foot.

raulk:
First batch of comments; I'll continue my review later!

raulk (commented):
First batch of comments; I'll continue my review later!

score.go
vyzo: No, there is no requirement to normalize the weight, so it is up to the application.

score.go
vyzo: We already have an implicit global cap: each individual topic score is capped, so the global cap is the weight sum of the caps.

score.go
vyzo: We could, but that just adds extra complexity here. Can we punt on this until we have a clear idea that it is a problem and how to resolve it?

gossipsub.go
Stebalien: Do we have any general protection against re-sending the same valid messages over and over in a cycle?

gossipsub.go
vyzo: Sure, we can do that.

score.go
vyzo: we can do that.

score.go
Stebalien: I wonder if a better approach would be to simply say that scores "bleed" between peers with similar IP addresses. That would allow us to say things like "everyone in this IP block appears to be bad, let's reduce their collective score".

Of course, that could be an attack vector...

score.go
Stebalien: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

gossipsub.go
vyzo: There is the timecache, which marks it for 2 minutes (configuration variable).
Outside of the timecach window it's up to the application to detect stale/replayed messages at validation time.

score.go
vyzo: the negative comment exists in the parameters struct, it's redundant here.

score.go
vyzo: I thought about it, but I think it's premature to do it now.

score.go
vyzo: Sure.

score.go
vyzo: Huh? Decay is multiplicative for all counters.

score.go
vyzo: ok.

score.go
vyzo: ok...

score.go
mcdee: +1 to removing `AppSpecificWeight`, I don't see it adding anything.

score.go
raulk: I'm asking for a code comment, not API docs. I'll add it myself. (We'll all be maintaining this code in the future.)

score.go
raulk: Then let's clarify that! ;-) These *Decay fields are factors that could be applied linearly, polynomially, exponentially, the reader doesn't know with

score.go
raulk: Complexity how? Counting connections and disconnections is simple; aren't we already subscribing to connectedness events, anyway?

Anyway, I think it's trivial to game the scoring system right now, and it shouldn't be this simple. cc @yiannisbot @daviddias

score.go
raulk: I was thinking of a scenario where an attacker could enumerate all topics we're interested in, and GRAFT with us on all of them, whereas the average honest peer only GRAFTs with us on a subset of topics. That attacker could accumulate a higher score by the mere fact that they're behaving well on more topics than the rest. But I guess that's how the algorithm is designed to work? Also, the probability of us accepting all of their `GRAFT`s would be low.

score.go
vyzo: That's a heavy statement. How would you game the score by connecting and disconnecting?

vyzo: we can make it so that only negative scores are retained if you are so concerned about this.

raulk: > How would you game the score by connecting and disconnecting?

1. Attacker gets GRAFT'ed onto a topic.
2. Attacker behaves well, accumulates positive score.
3. Attacker disconnects. Their positive score is retained for `RetainScore` time.
4. Just before `RetainScore` elapses, the attacker connects again. This refreshes their expiry time.
5. They wait a little, so that a score recalculation kicks in. This line will actually award them points for the time they've been disconnected: https://github.com/libp2p/go-libp2p-pubsub/blob/feat/hardening/score.go#L517 (EDIT: or maybe not? I don't know if we reset P1 on an interruption).
6. They disconnect.
7. Lather, rinse, repeat. You keep building up your P1 score without even being connected.

vyzo: Ok, that's a valid concern. The simple (and correct) solution is to only retain negative scores, so that the attacker can't build up score by using this method.

vyzo: P1 is only valid while they are in the mesh, it is not retained btw.

vyzo: So your attack is not plausible against P1, this simply disappears when they leave the mesh.
It is however a valid concern about building score by publishing messages with the first message deliveries counter.

raulk: @vyzo 

> P1 is only valid while they are in the mesh, it is not retained btw.

👍

> The simple (and correct) solution is to only retain negative scores, so that the attacker can't build up score by using this method.

I think we need something more sophisticated, though. I wouldn't like a peer to lose all their positive karma due to a 1-sec hiccup. That's why I suggested some form of failure counting and tolerance, or live/dead time tallying, after which, yes, we'd reset any positive scores.

vyzo: Following sync discussion, it's hard to build logic around disconnections because we don't know the reason of the disconnection -- it could just be the connection manager kicking in and us killing the connection.

gossipsub.go
Stebalien: Ah, got it. That's probably as good as we can get for now.

score.go
vyzo: done.

gossipsub.go
vyzo: done.

vyzo:
Made PX emission optional (to be enabled in bootstrappers) and added a gating threshold (>= 0) for accepting PX in order to address @lanzafame's gossiping wolves attack.

score.go
vyzo: On further thought it would make sense to cap the topic score contribution; I'll add an (optional) total topic score cap parameter.

score.go
vyzo: done.

score.go
vyzo: done.

score.go
vyzo: renamed to `WithPeerScoreInspect`.

score.go
vyzo: done.

score.go
vyzo: fixed.

score.go
vyzo: fixed.

score.go
vyzo: fixed.

score.go
vyzo: Implemented an (optional) TopicScoreCap parameter that produces the desired effect.

score.go
vyzo: I implemented negative score retention; take another look.
When we have a better idea how to retain positive scores, we can add the relevant logic.

gossipsub.go
vyzo: I have added the `ReceivedFrom` check; if a message has source as ourselves and was not received from ourselves, it is ignored and not forwarded.

score.go
vyzo: done.

score.go
vyzo: Implemented the `/64` and whitelisting for IPv6.

score.go
vyzo: done.

score.go
vyzo: sigh... done.

vyzo:
Merged support for direct peering agreements from #273.

raulk:
Taking another look now.

gossipsub.go
```diff
@@ -208,17 +371,34 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
...
+		return nil
+	}
+
 	ihave := make(map[string]*pb.Message)
 	for _, iwant := range ctl.GetIwant() {
 		for _, mid := range iwant.GetMessageIDs() {
-			msg, ok := gs.mcache.Get(mid)
-			if ok {
-				ihave[mid] = msg
+			msg, count, ok := gs.mcache.GetForPeer(mid, p)
```
raulk, line 406: This creates an entry in the mcache for each message ID the peer wants. Can't a peer just spoof random message IDs, and send us huge IWANTs up to the max message size, to make us OOM?

raulk: One way to prevent this is to track the IHAVEs we've sent, and interesect the incoming IWANTs against them. If the peer is asking for random stuff, we need to decrement their score or blacklist them entirely.

vyzo: It will only create an entry _iff_ we have the message in mcache, and it will last for just a few seconds (until we delete the message from the cache).
I don't think it's a plausible attack vector.

vyzo:
Merged #279 which implements opportunistic grafting.

vyzo:
the tests are now green on travis (check https://travis-ci.com/github/libp2p/go-libp2p-pubsub), but somehow the integration with github seems broken and has failed to report it.

vyzo:
ah nm, heisenbug -- it just got reported.

daviddias (approved):
We've discussed this offline and agree that the next step is:

- Merge this PR to master to support integrations in other projects
- Hold on releasing it as a new version for now.

People on board @vyzo @raulk @Stebalien

raulk (approved):

raulk:
Can we open an issue to track unit tests? Testground tests are showing a lot of improvement, but we really need regression tests that will run on CI.

cc @daviddias

vyzo:
Agreed -- @dirkmc is working on extending the coverage with unit tests.

vyzo:
Opened issue #281 for tracking unit test coverage.

Stebalien (dismissed):
Predicated on a _plan_ for landing tests. A plan that we stick to.

vyzo:
@Stebalien we have extensively tested on TestGround[TM]. 
Unit tests can only do so much, we can't test emergent behaviour and the reaction to various attacks with them. But we are working on adding basic unit tests that can catch regressions.

daviddias:
With regards to coverage

image

Seems that gossipsub is well covered above 70% and this is before @dirkmc latest PR to add more tests

It would be great to get to a minimum of 70% across the board. @vyzo is this something you can work on? It will help immensely future contributors that want to make changes without breaking the codebase.

vyzo:
@david -- we are working with @dirkmc to ensure good coverage for the v1.1 changes.
As for the rest, agreed that it would be nice to have. I will work on it time permitting, but might take a little while.

vyzo:
Rebased on master for merging.

daviddias:
YEaasasssss!!! Phenomenal work everyone!!!

bitcard:
mark

Stebalien:
@bitcard please stop commenting "mark" on random issues. If you want to bookmark something, save the link somewhere. Everyone subscribed to these issues gets an email whenever you do this.

PR #265, opened 2020-03-04
remove topic descriptors

### LINKED ISSUE #264: disable encryption
what do I need to do to unencrypt the message，I look at the source code, there should be no place to set，or I missed it
the source code is as follows：

```
message TopicDescriptor {
	optional string name = 1;
	optional AuthOpts auth = 2;
	optional EncOpts enc = 3;

	message AuthOpts {
		optional AuthMode mode = 1;
		repeated bytes keys = 2; // root keys to trust

		enum AuthMode {
			NONE = 0; // no authentication, anyone can publish
			KEY = 1; // only messages signed by keys in the topic descriptor are accepted
			WOT = 2; // web of trust, certificates can allow publisher set to grow
		}
	}

	message EncOpts {
		optional EncMode mode = 1;
		repeated bytes keyHashes = 2; // the hashes of the shared keys used (salted)

		enum EncMode {
			NONE = 0; // no encryption, anyone can read
			SHAREDKEY = 1; // messages are encrypted with shared key
			WOT = 2; // web of trust, certificates can allow publisher set to grow
		}
	}
}


func (p *PubSub) Subscribe(topic string, opts ...SubOpt) (*Subscription, error) {
	td := pb.TopicDescriptor{Name: &topic}

	return p.SubscribeByTopicDescriptor(&td, opts...)
}

func (p *PubSub) SubscribeByTopicDescriptor(td *pb.TopicDescriptor, opts ...SubOpt) (*Subscription, error) {
	if td.GetAuth().GetMode() != pb.TopicDescriptor_AuthOpts_NONE {
		return nil, fmt.Errorf("auth mode not yet supported")
	}
	if td.GetEnc().GetMode() != pb.TopicDescriptor_EncOpts_NONE {
		return nil, fmt.Errorf("encryption mode not yet supported")
	}
	topic, _, err := p.tryJoin(td.GetName())
	if err != nil {
		return nil, err
	}

	return topic.Subscribe(opts...)
}
```

Stebalien:
These have never actually been used and it doesn't look like anyone is actually calling these functions in practice.

fixes #264

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -105,6 +105,7 @@ github.com/libp2p/go-eventbus v0.1.0 h1:mlawomSAjjkk97QnYiEmHsLu7E136+2oCWSHRUvM
 github.com/libp2p/go-eventbus v0.1.0/go.mod h1:vROgu5cs5T7cv7POWlWxBaVLxfSegC5UGQf8A2eEmx4=
 github.com/libp2p/go-flow-metrics v0.0.1/go.mod h1:Iv1GH0sG8DtYN3SVJ2eG221wMiNpZxBdp967ls1g+k8=
 github.com/libp2p/go-flow-metrics v0.0.3 h1:8tAs/hSdNvUiLgtlSy3mxwxWP4I9y/jlkPFT7epKdeM=
+github.com/libp2p/go-flow-metrics v0.0.3 h1:8tAs/hSdNvUiLgtlSy3mxwxWP4I9y/jlkPFT7epKdeM=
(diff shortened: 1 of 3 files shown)


vyzo (approved):
fine by me.

vyzo:
Let's not rush to merge this though, I am tired of resolving conflicts because of changes in the protobufs in the open prs.

Stebalien:
This is why we merge things...

PR #273, opened 2020-03-31
gossipsub v1.1: Support direct peering agreements

vyzo:
This adds support for peers with direct peering agreements, as requested from above.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -146,6 +146,30 @@ func WithPeerExchange(doPX bool) Option {
 	}
 }
 
+// WithDirectPeers is a gossipsub router option that specifies peers with direct
+// peering agreements. These peers are connected outside of the mesh, with all (valid)
+// message unconditionally forwarded to them. The router will maintain open connections
+// to these peers. Note that the peering agreement should be reciprocal with direct peers
+// symmetrically configured at both ends.
+func WithDirectPeers(pis []peer.AddrInfo) Option {
+	return func(ps *PubSub) error {
+		gs, ok := ps.rt.(*GossipSubRouter)
+		if !ok {
+			return fmt.Errorf("pubsub router is not gossipsub")
+		}
+
+		direct := make(map[peer.ID]struct{})
+		for _, pi := range pis {
+			direct[pi.ID] = struct{}{}
+			ps.host.Peerstore().AddAddrs(pi.ID, pi.Addrs, peerstore.PermanentAddrTTL)
+		}
+
+		gs.direct = direct
+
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -392,6 +433,17 @@ func (gs *GossipSubRouter) handleGraft(p peer.ID, ctl *pb.ControlMessage) []*pb.
 			continue
 		}
 
+		// we don't GRAFT to/from direct peers; complain loudly if this happens
+		_, direct := gs.direct[p]
+		if direct {
```
raulk, line 438: So as you note below, this could happen in a non-reciprocal link. Do we really want to prune the peer altogether, i.e. refuse to give it any service entirely? I think printing the WARN is enough.

gossipsub.go
```diff
@@ -146,6 +146,30 @@ func WithPeerExchange(doPX bool) Option {
...
+		for _, pi := range pis {
+			direct[pi.ID] = struct{}{}
+			ps.host.Peerstore().AddAddrs(pi.ID, pi.Addrs, peerstore.PermanentAddrTTL)
+		}
+
+		gs.direct = direct
+
+		return nil
+	}
+}
```
raulk, line 171: The kinds of users using this option are likely to form a federated/hierarchical network topology. In the case of Filecoin, there are probably two types of links:

* intra-miner: miners will probably operate N network peers for redundancy. They will want to tag each of their peers as direct peers. Effectively this will create a flooding behaviour _inside_ their network.
* inter-miner: peering miners will not want to point to _all_ network peers of their counterparts; that would defeat the purpose of gossipsub and would end up devolving into an inefficient flooding network. Approaches:
  * each miner nominates a subset of all their peers as their "externally interfacing peers", and other peers add only those peers in their peering agreements. This might be too static.
  * introduce "peering groups" -- each group lists all the peers of the counterpart, with a _target active peer count_, e.g. 2 or 3. We would take care of peering with at most that many peers from that group.

raulk (commented):
Skimmed over this. LGTM, but there are some considerations re: the topology that could result from this.

gossipsub.go
vyzo: Yeah, I think so too.

gossipsub.go
vyzo: These are interesting considerations. For now, @jbenet just wants to have direct peering agreement support without much further complexity.

vyzo: We can revisit when the need for peering groups arises.

PR #278, opened 2020-04-07
fix: fix a busy loop while offline

### LINKED ISSUE #7049: IPFS hungs up in a busy loop (no announce filter all IPs)
#### Version information:
0.5.0-dev-84724a360

I got a client which only have private IPs (not worth announcing), so I filter them out (with no announce).

I noticed that IPFS also announces `::1` and `127.0.0.1`. When those are also added to the NoAnnounce filter, IPFS ends up with no IPs left to announce on startup.

While this should normally not impact network operations to find a proper relay (because of NAT), IPFS ends up in a busy loop complaining about the missing IP addresses.

Stebalien:
If advertising fails, we should wait and retry instead of busy looping.

fixes https://github.com/ipfs/go-ipfs/issues/7049

(and be extra careful about stopping timers, just because)

### DIFF
--- a/discovery.go
+++ b/discovery.go
@@ -21,6 +21,9 @@ var (
 	DiscoveryPollInterval = 1 * time.Second
 )
 
+// interval at which to retry advertisements when they fail.
+const discoveryAdvertiseRetryInterval = 2 * time.Minute
+
 type DiscoverOpt func(*discoverOptions) error
 
 type discoverOptions struct {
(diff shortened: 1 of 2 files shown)


Stebalien:
CI is broken because CI is broken.

ribasushi (approved):
LGTM

willscott (approved):

PR #279, opened 2020-04-10
Opportunistic grafting

vyzo:
Opportunistic grafting attempts to slowly improve underperforming meshes.

It works as follows: we check the median score of peers in the mesh; if this score is below the opportunisticGraftThreshold, we graft a peer at random with score over the median.
 
The intention is to (slowly) improve an underperforming mesh by introducing good scoring peers that may have been gossiping at us. This allows us to get out of sticky situations where we are stuck with poor peers and also recover from churn of good peers.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -951,7 +951,7 @@ func (gs *GossipSubRouter) heartbeat() {
 			medianIndex := len(peers) / 2
 			medianScore := scores[plst[medianIndex]]
 
-			// if the media score is below the threshold, select a better peer (if any) and GRAFT
+			// if the median score is below the threshold, select a better peer (if any) and GRAFT
 			if medianScore < gs.opportunisticGraftThreshold {
 				backoff := gs.backoff[topic]
 				plst = gs.getPeers(topic, 1, func(p peer.ID) bool {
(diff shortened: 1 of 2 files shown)


gossipsub.go
```diff
@@ -212,6 +221,9 @@ type GossipSubRouter struct {
 	// threshold for peer score before we graylist the peer and silently ignore its RPCs
 	graylistThreshold float64
 
+	// threshold for media peer score before triggering opportunistic grafting
```
raulk:
```suggestion
	// threshold for median peer score before triggering opportunistic grafting
```

gossipsub.go
```diff
@@ -909,6 +925,50 @@ func (gs *GossipSubRouter) heartbeat() {
...
+
+			// now compute the median peer score in the mesh
+			plst := peerMapToList(peers)
+			sort.Slice(plst, func(i, j int) bool {
+				return scores[plst[i]] < scores[plst[j]]
+			})
+			medianIndex := len(peers) / 2
+			medianScore := scores[plst[medianIndex]]
+
+			// if the media score is below the threshold, select a better peer (if any) and GRAFT
```
raulk:
```suggestion
			// if the median score is below the threshold, select a better peer (if any) and GRAFT
```

raulk (approved):
LGTM. Let’s take it for a spin on Testground!

vyzo:
Yup! That's precisely the plan!

PR #280, opened 2020-04-11
disable flaky tests

vyzo:
Clean up for #263 
@aschmahmann I have discovered that a couple of tests you wrote hung locally not just in travis.

### DIFF
--- a/discovery_test.go
+++ b/discovery_test.go
@@ -219,6 +219,7 @@ func TestSimpleDiscovery(t *testing.T) {
 }
 
 func TestGossipSubDiscoveryAfterBootstrap(t *testing.T) {
+	t.Skip("flaky test disabled")
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -369,6 +369,7 @@ func TestGossipsubGossip(t *testing.T) {
 }
 
 func TestGossipsubGossipPiggyback(t *testing.T) {
+	t.Skip("test no longer relevant; gossip propagation has become eager")
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	hosts := getNetHosts(t, ctx, 20)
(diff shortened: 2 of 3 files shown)


aschmahmann:
@vyzo `TestGossipSubDiscoveryAfterBootstrap` hangs in the `feat/hardening` branch and not in `master` (I tried numerous times on each). However I think it's likely this test is flaky. but that it can be difficult to distinguish from some of the other flakiness in the gossipsub tests that were there the last time I checked. If this test is causing you problems then just disable it and let's get it fixed.

Note: You should upgrade your libp2p dependencies (e.g. simultaneous connect fixes from https://github.com/libp2p/go-libp2p-swarm) since that would help get to the root of the problems in the tests and figure out whether the flakiness stems from the tests, pubsub, or some lower level libp2p component.

`TestSubscriptionManyNotifications` does occasionally hang, what's interesting is that the test hangs in the `connectAll` function which indicates that either the test somehow exhausts resources when running 33 pubsub instances or that there's a bug somewhere in the stack. Skipping the test and filing an issue if it's not fixed before merging this branch into master is fine by me.

vyzo:
Yeah, we have an issue for fixing flaky tests, they will be revisited.

vyzo:
What has changed in the hardening branch is that we now have backoff if we prune a peer -- is it possible that the test relied on regrafting immediately after a prune to establish a mesh?

aschmahmann:
Maybe that's related, in my brief debugging I noticed that there were peers that were "added" but hadn't made it into the mesh yet. The number of peers should be low enough for no pruning to happen, but maybe there was a disconnect. The GossipDlo was also increased which increased the number of nodes run in the test.

I found that adding the simultaneous connect patch helped, but the tests were still taking a long time (e.g. 2 minutes, which is not coincidentally the dial timeout for discovery based dials, makes me wonder if there's weird dial behavior/resource contention going on).

vyzo:
Also note that discovery itself is not broken, we tested in the testground and it seemed to work fine.

PR #283, opened 2020-04-14
Hardening unit tests

dirkmc:
See https://github.com/libp2p/go-libp2p-pubsub/issues/281

Tests:
- [x] IWANT spam
- [x] IHAVE spam
- [X] GRAFT spam
- [x] Invalid message spam

PR #284, opened 2020-04-20
Improve spam protections

vyzo:
Currently we return a PRUNE for every GRAFT we are rejecting, even if the GRAFTs are coming too fast, in flood style. Similarly, we are asking for every (unknown) message advertised in an IHAVE even if these are too many, and potentially a flood.

This patch adds rudimentary protections for both, closing the (single node) flood holes:
- We limit the number of message ids we emit in an IHAVE and accept in an IWANT per heartbeat to `GossipSubMaxIHaveLength`; default is 5000 message ids, which would take some 200KB.
  We also only accept up to `GossipSubMaxIHaveMessages` IHAVE messages per heartbeat from each peer.
- We stop emitting PRUNES if a GRAFT is coming too fast since the last backoff, controlled by the `GossipSubGraftFloodThreshold` parameter. We also add an extended backoff penalty to avoid GRAFTING ourselves on this peer for a while.

Also merges in #283 which tests the spam protections.


gossipsub.go
```diff
@@ -70,20 +70,37 @@ var (
...
+	// then there is no PRUNE response emitted. This protects against GRAFT floods and should be
+	// less than GossipSubPruneBackoff.
+	GossipSubGraftFloodThreshold = 10 * time.Second
+
+	// Maximum number of messages to include in an IHAVE message. Also controls the maximum
+	// number of IHAVE ids we will accept and request with IWANT from a peer within a heartbeat,
+	// to protect from IHAVE floods. You should adjust this value from the default if your
+	// system is pushing more than 5000 messages in GossipSubHistoryGossip heartbeats; with the
+	// defaults this is 1666 messages/s.
+	GossipSubMaxIHaveLength = 5000
```
raulk, line 87: Related question: do we track inflight IWANTs? My rationale is that if we are poorly connected to the mesh and we recurrently only hear about a subset of messages via gossip, we might end up asking many peers for the same message at once, unless we deduplicate inflight requests.

If we don't, we'd be selfishly causing more traffic than necessary. However, I acknowledge this is tricky logic to implement, because it needs to factor in many concerns:

* What if the peer we ask never sends back the message?
* Do we queue up other candidates we know we can fall back on?
* What is the grace period? And how does it relate to window sliding? (by the time we fall back to the next candidate, the message might have slid away from the cache)
* Memory footprint of this new state to maintain.

gossipsub.go
```diff
@@ -70,20 +70,37 @@ var (
...
-		connect: make(chan connectInfo, GossipSubMaxPendingConnections),
-		mcache:  NewMessageCache(GossipSubHistoryGossip, GossipSubHistoryLength),
+		peers:    make(map[peer.ID]protocol.ID),
+		mesh:     make(map[string]map[peer.ID]struct{}),
+		fanout:   make(map[string]map[peer.ID]struct{}),
+		lastpub:  make(map[string]int64),
+		gossip:   make(map[peer.ID][]*pb.ControlIHave),
+		control:  make(map[peer.ID]*pb.ControlMessage),
+		backoff:  make(map[string]map[peer.ID]time.Time),
+		peerhave: make(map[peer.ID]int),
```
raulk, line 103: note for reviewers: `peerhave` is what was added here. GitHub diffs don't make it easy to hunt down the change :D

gossipsub.go
```diff
@@ -359,6 +378,18 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
 		return nil
 	}
 
+	// IHAVE flood protection
+	if gs.peerhave[p] > GossipSubMaxIHaveMessages {
```
raulk, line 386: Can we capture the value of `gs.peerhave[p]` and output it on the log? It's useful for debugging purposes.

gossipsub.go
```diff
@@ -359,6 +378,18 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
 		return nil
 	}
 
+	// IHAVE flood protection
```
raulk, line 384: Sorry for the dumb question, but since we're taking no locks here, can you remind me if this is performed under a monothreaded event loop?

gossipsub.go
```diff
@@ -359,6 +378,18 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
...
+	if gs.peerhave[p] > GossipSubMaxIHaveMessages {
+		log.Debugf("IHAVE: peer %s has advertised too many times within this heartbeat interval; ignoring", p)
+		return nil
+	}
+	gs.peerhave[p]++
+
+	if gs.iasked[p] >= GossipSubMaxIHaveLength {
```
raulk, line 391: Same as above (log the current value).

gossipsub.go
```diff
@@ -359,6 +378,18 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
 		return nil
 	}
 
+	// IHAVE flood protection
+	if gs.peerhave[p] > GossipSubMaxIHaveMessages {
+		log.Debugf("IHAVE: peer %s has advertised too many times within this heartbeat interval; ignoring", p)
+		return nil
+	}
+	gs.peerhave[p]++
```
raulk: Can we increment the value of `gs.peerhave[p]` anyway? That way we can record how many times a peer actually sent us IHAVEs, even if we discarded them.

gossipsub.go
```diff
@@ -459,6 +499,23 @@ func (gs *GossipSubRouter) handleGraft(p peer.ID, ctl *pb.ControlMessage) []*pb.
...
 		}
 
+		// make sure we are not backing off that peer
+		expire, backoff := gs.backoff[topic][p]
+		if backoff && now.Before(expire) {
+			log.Debugf("GRAFT: ignoring backed off peer %s", p)
+			// refresh the backoff
+			gs.addBackoff(p, topic)
+			// check the flood cutoff -- is the GRAFT coming too fast?
+			floodCutoff := expire.Add(GossipSubGraftFloodThreshold - GossipSubPruneBackoff)
```
raulk, line 510: Another (possibly more correct) way of implementing this could be sending back the desired backoff time to a peer at the time of PRUNE, and penalising peers who do not honour it?

gossipsub.go
```diff
@@ -1111,6 +1173,15 @@ func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}
 		return
 	}
 
+	// shuffle to emit in random order
+	shuffleStrings(mids)
+
+	// if we are emitting more than GossipSubMaxIHaveLength mids, truncate the list
+	if len(mids) > GossipSubMaxIHaveLength {
+		log.Debugf("too many messages for gossip; truncating IHAVE list (%d messages)", len(mids))
+		mids = mids[:GossipSubMaxIHaveLength]
```
raulk: Would it make more sense to shuffle by peer? Or maybe shuffle once, then create as many partitions as peers we are gossiping to? I'm thinking we want to get the word out there to _someone_ for as many messages as we can, instead of picking a set and cloning the gossip for all peers?

raulk (commented):
Few comments.

gossipsub.go
vyzo: sure!

gossipsub.go
vyzo: No, we are not tracking in flight IWANTs. 
I agree that we do want to track them, and it will make it much easier to respond to distributed IHAVE floods as well.
I'll open an issue so that we can tackle this, as it is rather complex and I want to think a bit about it.

gossipsub.go
vyzo: yes, it's monothreaded in the event loop.

gossipsub.go
vyzo: sure.

gossipsub.go
vyzo: sure, good point.

gossipsub.go
vyzo: Yeah, it would be nice of us to send the backoff period and it makes things more robust for peers with different values of it.
It will take a protocol change however, so let me consider a bit and do it in a follow up PR.

gossipsub.go
vyzo: This is a good point, and we have enough redundancy in the system that even if we truncate the list we'll get multiple IHAVEs from different peers and eventually all the messages.
Only downside is that we need to copy the slice.

gossipsub.go
vyzo: done.

gossipsub.go
vyzo: done.

gossipsub.go
vyzo: done.

gossipsub.go
vyzo: done.

dirkmc (approved):
LGTM 👍 

In Bitswap we try to "keep peers busy" when serving blocks, meaning we send messages to whichever peer has the least amount of data in its send queue. It may make sense to follow an analogous approach here, whereby when deciding how many messages to send to each peer we prioritize them by score, and then round-robin

gossipsub.go
vyzo: follow up in #285.

gossipsub.go
vyzo: follow up in #286.

Issue #287, opened 2020-04-20
Improve test coverage


dirkmc:
This issue tracks improvements in test coverage for
- [x] score.go (started at 32%)
- [x] score_params.go (started at 12%)
- [x] gossipsub.go (started at 63%)
- [x] trace.go (started at 58%)
- [x] tracer.go (started at 0%)
- [x] pubsub.go (started at 73%)
- [x] validation.go (started at 57%)
- [x] randomsub.go (started at 0%)

vyzo:
We are a little over 82% now, which was the target (80%) for the big test writing push.
We have achieved satisfactory coverage, so I will close the issue.

Great job fellas!

PR #288, opened 2020-04-20
Score tests

dirkmc:
Part of https://github.com/libp2p/go-libp2p-pubsub/issues/287
- [x] Time in mesh
- [x] First message deliveries
- [x] Mesh message deliveries
- [x] Mesh failure penalty
- [x] Invalid message deliveries
- [x] App specific score
- [x] IP colocation factor

### DIFF
--- a/score_test.go
+++ b/score_test.go
@@ -1,0 +1,86 @@
+package pubsub
+
+import (
+	"testing"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+func TestScoreTimeInMesh(t *testing.T) {
+	// Create parameters with reasonable default values
+	mytopic := "mytopic"
+	params := &PeerScoreParams{
+		AppSpecificScore: func(peer.ID) float64 { return 0 },
+		Topics:           make(map[string]*TopicScoreParams),
+	}
+	topicScoreParams := &TopicScoreParams{
+		TopicWeight:       0.5,
+		TimeInMeshWeight:  1,
+		TimeInMeshQuantum: time.Millisecond,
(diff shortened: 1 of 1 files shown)


score_test.go
```diff
@@ -0,0 +1,134 @@
...
+	params := &PeerScoreParams{
+		AppSpecificScore: func(peer.ID) float64 { return 0 },
+		Topics:           make(map[string]*TopicScoreParams),
+	}
+	topicScoreParams := &TopicScoreParams{
+		TopicWeight:       1,
+		FirstMessageDeliveriesWeight:  1,
+		FirstMessageDeliveriesDecay: 1.0, // test without decay for now
+		FirstMessageDeliveriesCap: 2000,
+		TimeInMeshQuantum: time.Second, // bug? not setting this causes a div by zero
```
yusefnapora: @vzyo - I hit a panic when this is unset, since we divide by it when calculating the time in mesh score.

vyzo: ouch -- we should fix the validation to catch this and return an error.
once we do that let's also update the validation test to keep the score_params coverage to 100% :)

yusefnapora:
reading through, I just realized I'm using the `withinVariance` helper in places where I don't really need to... we should be able to do exact comparisons for the scores that aren't based on time.

vyzo:
good point.

vyzo:
failure in `TimeInMesh`:
```
=== RUN   TestScoreTimeInMesh

    TestScoreTimeInMesh: score_test.go:47: Score: 35.000000. Expected 20.000000 ± 10.000000

```

vyzo (approved):
looks great!

PR #289, opened 2020-04-21
deflake validator overload test

vyzo:
see #202 
This test regularly fails on travis because of the way it was written -- the goroutines would become available, resulting in a successful validation even though a validation failure was expected.
This is an attempt to fix; also moves the validation tests to their own file.


daviddias (approved):
LGTM (for other viewers, check the changes by commit so that it is obvious what changed)

PR #295, opened 2020-04-22
more test coverage for gossipsub

vyzo:
Adds more tests to increase coverage, as the title suggests.
See #287


gossipsub.go
```diff
@@ -287,6 +287,8 @@ func (gs *GossipSubRouter) Attach(p *PubSub) {
 	// connect to direct peers
 	if len(gs.direct) > 0 {
 		go func() {
+			// add a small delay to make this unit-testable
+			time.Sleep(time.Second)
```
yusefnapora: Should we make this a variable, so that we only add the delay when testing?

yusefnapora (approved):
This looks great to me

gossipsub.go
vyzo: Maybe so, but a 1s delay to connect can't hurt the real use case :)

vyzo: alright, I'll turn it into a variable -- not sure we gain much by it though, I'll leave it to its default value of 1s.

vyzo: done (will push shortly).

dirkmc (approved):

.travis.yml
```diff
@@ -9,7 +9,6 @@ go:
 env:
   matrix:
     - GOTFLAGS="-race"
-    - GOTFLAGS="-race -tags=openssl"
```
Stebalien: "fails with no output produced" is a reason to track down the bug and fix it. Let's not just remove a test.

vyzo: I have added a `travis_wait` and increased the test timeout, and things seem to be smooth now.
I can restore.

vyzo: I suspect the reason is that travis takes a very long time to build the tests for running, which is exacerbated with openssl.

vyzo: restored.

PR #299, opened 2020-04-22
add test with signed peer records

yusefnapora:
@vyzo as suggested, I just copied the star topology test and tweaked it a bit. It manually creates a signed peer record for each host and adds it to all the other hosts' peer stores.

I'm not sure how to test the error cases (e.g. failing to marshal the signed peer record), but at least this gets rid of the big red block in the coverage.

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/libp2p/go-libp2p-core/record"
 	"io"
 	"math/rand"
 	"testing"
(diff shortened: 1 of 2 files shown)


gossipsub_test.go
```diff
@@ -995,6 +996,83 @@ func TestGossipsubStarTopology(t *testing.T) {
...
+	defer cancel()
+
+	hosts := getNetHosts(t, ctx, 20)
+	psubs := getGossipsubs(ctx, hosts, WithPeerExchange(true))
+
+	// manually create signed peer records for each host and add them to the
+	// peerstores of the other hosts.
+	// this is necessary until the identify service is updated to exchange
+	// signed records
+	for i := range hosts {
```
vyzo: I think we should do this after the connect step and only add the peer records to the connected hosts.
This will allow us to test the actual exchange of records without having anything in the peer store and no routing to fall back on.

vyzo: Actually we don't have to add them anywhere other than the center of the star now that I think about it.

vyzo (commented):
Looking good, just one minor but important detail.

gossipsub_test.go
vyzo: done.

vyzo (approved):

PR #301, opened 2020-04-23
make go test produce output

vyzo:
Trying to debug the "no output received" error we occasionally see, even with the 20m wait/timeout.

This adds `-x` to the `go test` invocation so that we get verbose output to
- debug the situation; I suspect the build gets stuck.
- make travis not give up during the build and before the tests have even started.


vyzo:
The builds are in `travis_wait` invocation yet there is no output from `go test`.
Honestly I have no idea what is happening, it seems like an internal travis scheduling issue perhaps.

vyzo:
@stebalien any other ideas? This is a very annoying problem.

vyzo:
So the observed behaviour was that all verbose output came together after 15min or so, which was not the intended effect.
Still searching for a clean solution to this problem -- all I can think is increasing the travis wait further from 20m (the default in the command) to even 30 min.
This incidentally is what the travis docs recommend to address this issue.

PR #304, opened 2020-04-23
sqrt behaviour for randomsub plus tests

vyzo:
See #287.

### DIFF
--- a/randomsub.go
+++ b/randomsub.go
@@ -18,19 +18,21 @@ var (
 )
 
 // NewRandomSub returns a new PubSub object using RandomSubRouter as the router.
-func NewRandomSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
+func NewRandomSub(ctx context.Context, h host.Host, size int, opts ...Option) (*PubSub, error) {
 	rt := &RandomSubRouter{
+		size:  size,
 		peers: make(map[peer.ID]protocol.ID),
 	}
 	return NewPubSub(ctx, h, rt, opts...)
 }
 
 // RandomSubRouter is a router that implements a random propagation strategy.
-// For each message, it selects the square root of peers, with a min of RandomSubD,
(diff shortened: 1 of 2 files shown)


yusefnapora (approved):
looks good 👍

dirkmc (approved):

randomsub.go
```diff
@@ -119,6 +123,14 @@ func (rs *RandomSubRouter) Publish(msg *Message) {
...
+		sqrt := int(math.Ceil(math.Sqrt(float64(rs.size))))
+		if sqrt > target {
+			target = sqrt
+		}
+		if target > len(rspeers) {
+			target = len(rspeers)
+		}
 		xpeers := peerMapToList(rspeers)
 		shufflePeers(xpeers)
 		xpeers = xpeers[:RandomSubD]
```
emberian, line 136: It goes through all the work of calculating target, but it's not actually used - should it replace the use of `RandomSubD` here?

randomsub.go
```diff
@@ -119,6 +123,14 @@ func (rs *RandomSubRouter) Publish(msg *Message) {
 	}
 
 	if len(rspeers) > RandomSubD {
+		target := RandomSubD
+		sqrt := int(math.Ceil(math.Sqrt(float64(rs.size))))
```
emberian, line 127: This assumes the size of the network is fixed for the lifetime of the router - what's the suggested use of randomsub in the case where the network size is varying?

randomsub.go
vyzo: ooops, that's a bug! thanks for spotting.

randomsub.go
vyzo: It is more of a hint actually. Note that we don't recommend using randomsub in real networks, it was done more as an experiment.

randomsub.go
vyzo: Care for a patch?

PR #309, opened 2020-04-24
More coverage for score.go and a bug fix

vyzo:
I decided that I didn't like those red blocks in the coverage of score.go and look at that, I discovered a bug.


yusefnapora (approved):
good catch!

PR #310, opened 2020-04-24
Add explicit control of the protocol list supported by the gossipsub router

vyzo:
Some networks may want to disable floodsub compatibility, as it could put a strain on highly connected peer resources.


yusefnapora:
Is any of the gossipsub v1.1 behavior changed if we don't advertise the new protocol ID?

vyzo:
PX is affected, as it won't try to do PX when the peer is gsv1.0. Other than that, not really!

vyzo:
Closing in favor of a more general solution for #408.

PR #312, opened 2020-04-27
only count connections with pubsub streams in ip colocation factor

vyzo:
otherwise we might be vulnerable to active MITM attacks on secio.


vyzo:
I am going to add a test to satisfy the requirements of keeping coverage.

yusefnapora (approved):
I'm sorry @vyzo, for some reason I thought I'd already reviewed this. It seems good to me 👍

PR #313, opened 2020-04-28
Implement Extended Validators

vyzo:
See https://github.com/libp2p/specs/pull/265 and https://github.com/ethereum/eth2.0-specs/issues/1761

### DIFF
--- a/validation.go
+++ b/validation.go
@@ -384,6 +384,8 @@ func (v *validation) validateTopic(vals []*topicVal, src peer.ID, msg *Message) 
 			result = ValidationReject
 			break
 		case ValidationIgnore:
+			// throttled validation has the same effect, but takes precedence over Ignore as it is not
+			// known whether the throttled validator would have signaled rejection.
 			if result != validationThrottled {
 				result = ValidationIgnore
 			}
--- a/validation.go
+++ b/validation.go
@@ -15,8 +15,30 @@ const (
 	defaultValidateThrottle    = 8192
 )
 
-// Validator is a function that validates a message.
+// Validator is a function that validates a message with a binary decision: accept or reject.
 type Validator func(context.Context, peer.ID, *Message) bool
+
+// ValidatorEx is an extended validation function that validates a message with an enumerated decision
+type ValidatorEx func(context.Context, peer.ID, *Message) ValidationResult
+
+// ValidationResult represents the decision of an extended validator
+type ValidationResult int
+
+const (
+	// ValidationAccept is a validation decision that indicates a valid message that should be accepted and
+	// delivered to the application and forwarded to the network.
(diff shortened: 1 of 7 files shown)


vyzo:
A note should be made in the behaviour of multi-topic validation decisions:
- If any validator rejects, the message is rejected
- if any validator ignores or throttles (but none rejects), the message is ignored
- the message is accepted if and only if all validators accept.

validation.go
```diff
@@ -109,9 +131,36 @@ func (v *validation) AddValidator(req *addValReq) {
...
+	case Validator:
+		validator = makeValidatorEx(v)
+
+	case func(ctx context.Context, p peer.ID, msg *Message) ValidationResult:
+		validator = ValidatorEx(v)
+	case ValidatorEx:
+		validator = v
+
+	default:
+		req.resp <- fmt.Errorf("Unknown validator type for topic %s; must be an instance of Validator or ValiatorEx", topic)
```
yusefnapora:
```suggestion
		req.resp <- fmt.Errorf("Unknown validator type for topic %s; must be an instance of Validator or ValidatorEx", topic)
```

validation.go
```diff
@@ -292,55 +372,63 @@ loop:
...
-		valid := <-rch
-		if !valid {
-			return false
+		switch <-rch {
+		case ValidationAccept:
+		case ValidationReject:
+			result = ValidationReject
+			break
+		case ValidationIgnore:
+			if result != validationThrottled {
```
yusefnapora, line 389: @vyzo in the PR comment you said "if any validator ignores or throttles (but none rejects), the message is ignored", but this looks like if a message is both ignored and throttled, the final result will be `validationThrottled` instead of `validationIgnored`. 

That seems fine, I just wanted to mention in case the difference is meaningful. Also, maybe it's worth dropping a code comment here to explain the behavior?

vyzo: yes, this is intentional; throttling supersedes ignoring as we don't know if the throttled validator would have rejected.
I can surely add a comment here nonetheless.

yusefnapora:
@vyzo the travis failure has this message:

```
    TestGossipsubScoreValidatorEx: gossipsub_test.go:1434: expected negative score for peer2, but got 0.000000
```

Is it possible travis is so slow that the score is decaying to zero, even with decay at `0.9999`?

vyzo:
Yeah I noticed, I am making a commit to add a delay after subscription and before publishing the messages.

validation.go
vyzo: done.

vyzo:
I will add a couple more tests to ensure coverage in the patch itself, it has a couple of red blocks.

yusefnapora (approved):

vyzo:
rebased on master.

PR #314, opened 2020-04-29
add utility function for computing score parameter decays

vyzo:
First baby step towards easy default score parameter computations is to have the decay factor become time instead of a float.


yusefnapora (approved):
ship it!

Issue #292, opened 2020-04-21
Question about pubsub message relaying


lukasz-zimnoch:
Looking at `handleIncomingRPC` method defined in `pubsub.go` we can see a part which forwards received messages if our `pubsub` instance has an existing subscription to the given message topic:

```
(...)

for _, pmsg := range rpc.GetPublish() {
	if !p.subscribedToMsg(pmsg) {
		log.Warning("received message we didn't subscribe to. Dropping.")
		continue
	}

	msg := &Message{pmsg, rpc.from, nil}
	p.pushMsg(msg)
}

p.rt.HandleRPC(rpc)
```

If our pubsub doesn't have this subscription, we are just ignoring the current message. Also, `HandleRPC` is a no-op in floodsub implementation so we don't do anything with the `rpc` later. 

The question: is there a particular reason why `pubsub` ignores messages they are not subscribed to, instead of just passing them to other peers and gain on the delivery rate?

vyzo:
There is no reason to pass messages if we are not subscribing, that's a DoS waiting to happen.
We are considering adding a relay-only option to pubsub that would enable relaying without subscription.

pdyraga:
> We are considering adding a relay-only option to pubsub that would enable relaying without subscription.

Is there any PR or issue for that one where we could learn more and potentially contribute?

Unless I misunderstood, I think this is what @lukasz-zimnoch is asking about. If there is a peer that is not subscribed to the given topic and this peer is the only point currently connecting two other peers subscribed to that topic, we could have a relay-only somewhere in pubsub allowing to pass-thru messages.

vyzo:
Yes, issue #28 has been open for a while.

vyzo:
Yes, issue #28 has been open for a while.

lukasz-zimnoch:
@vyzo I see there was already a contribution https://github.com/libp2p/go-libp2p-pubsub/pull/229 but has been closed. We are interested in contributing here but we need to know more. Why #229 didn't pass and what is a must for a complete solution here? Could you provide those details?

pdyraga:
@lukasz-zimnoch @vyzo What if we close this issue as it kinda duplicates #28 and we'll continue the discussion there?

vyzo:
I wanted to avoid distractions while the gossipsub v1.1 work was in progress; we can certainly reopen that pr.

vyzo:
[...]

PR #317, opened 2020-04-30
Topic relay

lukasz-zimnoch:
Refs #28 #292 

This pull request ships an implementation of the topic `Relay` operation. That opens the possibility to take part in voluntary topic dissemination by all interested nodes without the need for subscribing.  

**Implementation details**

This implementation tries to follow @vyzo [comment](https://github.com/libp2p/go-libp2p-pubsub/issues/292#issuecomment-619058811) thus it uses a relay reference counter and try to not conflate with subscriptions. However, regarding peer discovery, we probably must follow the same logic as subscriptions which is reflected in the implementation. Also, relays canceling logic has probably a place for improvements so I'd appreciate suggestions on how I can do it better to satisfy all possible use cases.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -607,8 +607,8 @@ func (p *PubSub) handleRemoveTopic(req *rmTopicReq) {
 }
 
 // handleRemoveSubscription removes Subscription sub from bookeeping.
-// If this was the last Subscription for a given topic, it will also announce
-// that this node is not subscribing to this topic anymore.
+// If this was the last subscription and no more relays exist for a given topic,
+// it will also announce that this node is not subscribing to this topic anymore.
 // Only called from processLoop.
 func (p *PubSub) handleRemoveSubscription(sub *Subscription) {
 	subs := p.mySubs[sub.topic]
@@ -634,8 +634,8 @@ func (p *PubSub) handleRemoveSubscription(sub *Subscription) {
 }
 
 // handleAddSubscription adds a Subscription for a particular topic. If it is
-// the first Subscription for the topic, it will announce that this node
-// subscribes to the topic.
+// the first subscription and no relays exist so far for the topic, it will
+// announce that this node subscribes to the topic.
 // Only called from processLoop.
 func (p *PubSub) handleAddSubscription(req *addSubReq) {
 	sub := req.sub
@@ -661,8 +661,8 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
 }
 
 // handleAddRelay adds a relay for a particular topic. If it is
-// the first relay for the topic, it will announce that this node
-// relays for the topic.
+// the first relay and no subscriptions exist so far for the topic , it will
+// announce that this node relays for the topic.
 // Only called from processLoop.
 func (p *PubSub) handleAddRelay(req *addRelayReq) {
 	topic := req.topic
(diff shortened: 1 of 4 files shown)


pubsub.go
```diff
@@ -638,6 +660,54 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
...
+	}
+
+	p.myRelays[topic]--
+
+	if p.myRelays[topic] == 0 {
+		delete(p.myRelays, topic)
+	}
+
+	// stop announcing only if there are no more relays and subs
+	if p.myRelays[topic] == 0 && len(p.mySubs[topic]) == 0 {
```
vyzo: It is probably cleaner to merge the second conditional with the body of the first:
```
If p.myRelays[topic] == 0 {
  delete(p.myRelays, topic)

  if len(p.mySubs[topic]) == 0 {
   ...
  }
}
```

pubsub.go
```diff
@@ -638,6 +660,54 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
...
+		p.rt.Join(topic)
+	}
+
+	p.myRelays[topic]++
+
+	req.resp <- func() {
+		select {
+		case p.rmRelay <- topic:
+		case <-p.ctx.Done():
+		}
```
vyzo: It is possible to invoke the cancellation function more than once, and cancel unrelated relaying operations.
I think it is best to keep a variable in the closure (a boolean would do) that gets poisoned if it is invoked once, and avoids the notification in subsequent calls.

vyzo: We would also want a test for this behaviour.

topic_test.go
```diff
@@ -510,6 +527,134 @@ func TestSubscriptionNotificationSubUnSub(t *testing.T) {
...
+	relay3Cancel, err := topics[0].Relay()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(time.Millisecond * 100)
+
+	if pubsubs[0].myRelays[topic] != 3 {
+		t.Fatal("incorrect number of relays")
+	}
```
vyzo: This results in a race in the test (see the travis failure),
If you need to examine the internal state of pubsub, you can use the eval channel:
```
pubsubs[0].eval <- func() {
  ...
}
```

topic_test.go
```diff
@@ -510,6 +527,134 @@ func TestSubscriptionNotificationSubUnSub(t *testing.T) {
...
+
+	relay1Cancel()
+	relay2Cancel()
+	relay3Cancel()
+
+	time.Sleep(time.Millisecond * 100)
+
+	if pubsubs[0].myRelays[topic] != 0 {
+		t.Fatal("incorrect number of relays")
+	}
```
vyzo: same here.

pubsub.go
```diff
@@ -638,6 +660,54 @@ func (p *PubSub) handleAddSubscription(req *addSubReq) {
...
+	topic := req.topic
+
+	// announce we want this topic if neither relays nor subs exist so far
+	if p.myRelays[topic] == 0 && len(p.mySubs[topic]) == 0 {
+		p.disc.Advertise(topic)
+		p.announce(topic, true)
+		p.rt.Join(topic)
+	}
+
+	p.myRelays[topic]++
```
vyzo: We need to increment the counter _before_ trying to announce; the check should then be for the counter would be against 1.

vyzo: Note that it is not necessary a problem for the code as it is written right now, but it's subtlety we would like to avoid.

vyzo (changes requested):
Thank you! This overall looks good, just a few minor issues that should be addressed.

vyzo:
Do you mind rebasing on master? The branch is currently out of date, which would prevent us from merging.

lukasz-zimnoch:
Thanks for review @vyzo! All comments addressed.

vyzo (approved):
thank you!

Issue #286, opened 2020-04-20
Notify pruned peers of the backoff period


vyzo:
When we prune peers we have a backoff period, during which we won't re-GRAFT the pruned peer.
Furthermore, with #284 we have added GRAFT flood protections so that peers who send us GRAFTs too fast get penalized for quite a while.
From all this, it has become quite obvious that we should send the backoff period back to the pruned peer so that it knows when it's safe for it to re-attempt a GRAFT.
This will protect us from bugs that might arise because of a mismatch in the backoff period between peers, and also more clearly mark backoff violating peers for penalties, simplifying the logic somewhat.

The only downside is that it will take a (small, non-breaking) wire protocol change.

dirkmc:
Currently IIUC we extend the backoff each time a peer tries to GRAFT too early.
We may want to consider not extending the backoff period if a peer asks to GRAFT too soon (unless they do it too frequently, which we are now checking for)

vyzo:
I think we do want the behaviour of extending the backoff period when pruning for two reasons:
- we make sure we don't try to GRAFT while the other side is still backing us off
- we stop GRAFT floods dead in their tracks; basically a selfish or malicious peer trying to GRAFT its way in won't have any luck getting in.

Issue #285, opened 2020-04-20
Track in-flight IWANT requests


vyzo:
This came up in #284.

It is desirable to track in-flight IWANT requests, so that:
- we don't repeatedly ask multiple peers for the same message when we are poorly connected to the mesh
- we can track spammers who send us bogus message IDs and never follow up on the IWANT request in order to execute a distributed IHAVE flood attack.

Complexities, as enumerated by @raulk:
- What if the peer we ask never sends back the message?
- Do we queue up other candidates we know we can fall back on?
- What is the grace period? And how does it relate to window sliding? (by the time we fall back to the next candidate, the message might have slid away from the cache)
- Memory footprint of this new state to maintain.

whyrusleeping:
This kinda just feels like bitswap

vyzo:
yeah, but in real-time.

vyzo:
Maybe one day bitswap can switch to gossipsub for supporting block discovery instead of using the dht.
It would be much more efficient and it will avoid taxing the provider records system.
One idea would be to find the root CID through the dht, using provider records, and then use a pubsub topic for handling block discovery.
cc @Stebalien

vyzo:
cc @dirkmc as well.

dirkmc:
In Bitswap nodes remember which blocks their peers want for the lifetime of the connection. That probably wouldn't be necessary for gossipsub so it should be possible to have lower memory requirements.

aschmahmann:
> One idea would be to find the root CID through the dht, using provider records, and then use a pubsub topic for handling block discovery.

This was proposed in January, it's a pretty reasonable idea. However, there's some tricky work to be done in figuring out if/how we'd want IPFS to switch from a random access pattern to one that's slightly different (e.g. if you only provide/find the root then people can't string arbitrary blocks together into a new file and get all the providers of those blocks to help out, but it seriously helps on resource usage). libp2p/go-libp2p-kad-dht#559 is the first step along the path.

vyzo:
So a solution for tracking IWANTs for non-follow up purposes without using too much memory, is to random sample the set of IWANTs and track only a small subset of them. 
This will allow us to respond to a distributed IHAVE flood attack without blowing the memory.
It will not help with duplicate IWANTs, but that's less of a problem and less pressing to solve.

PR #318, opened 2020-05-01
P7: Behaviour pattern penalties

vyzo:
This adds a new score parameter, P7, which is a quadratic counter of behavioural pattern penalties.
In contrast to other parameters, which are implicitly updated through tracing, this counter is explicitly incremented by the router.
It is intended to capture and penalize behavioural patterns in #285 and #286, such as not following up on `IWANT` requests from `IHAVE` advertisements or not respecting the prune backoff, but it can also be used for other behaviours in the future.

cc @raulk

### DIFF
--- a/score.go
+++ b/score.go
@@ -26,6 +26,9 @@ type peerStats struct {
 
 	// IP tracking; store as string for easy processing
 	ips []string
+
+	// behavioural pattern penalties (applied by the router)
+	behaviourPenalty float64
 }
 
 type topicStats struct {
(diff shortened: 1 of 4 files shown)


yusefnapora (approved):
This seems good, as long as we document (here and in the spec) which behaviors are penalized when we add the actual penalties in future PRs.

vyzo:
Yes of course! I will update the spec once #285 and #286 are ready.

vyzo:
Rebased on master for merging.

PR #319, opened 2020-05-04
add defensive checks for potentially duplicate traces in score

vyzo:
It is possible that a message reappears after it has slided out of the message cache but before it has been gc'ed from the delivery record cache, which can result in a panic if the delivery status changes from throttled/ignored/rejected to valid.

This should stop the panic and also give us a trace to debug the situation.


yusefnapora (approved):
Looks good 👍

PR #320, opened 2020-05-05
Fragment large RPCs into multiple messages

### LINKED ISSUE #307: Implement RPC fragmentation for packed messages
We pack messages in a single RPC in response to IWANT requests.
This may pack too many or big(ish) messages and exceed the max message size limit, which would break the stream to the requester.
We need to fragment large packings of messages into multiple RPCs.
In fact we should probably fragment even it exceeds just a threshold, so that we minimize latency for the message pack.

yusefnapora:
This adds a `fragmentRPC` function that's used by gossipsub's `sendRPC` method to avoid writing RPC messages that are larger than the maximum message size, which can happen if a peer requests a lot of large messages via gossip.

It can fail if an individual message is larger than the maximum, but I haven't written a test for that code path yet.

closes #307

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -939,10 +939,14 @@ func fragmentRPC(rpc *RPC, limit int) ([]*RPC, error) {
 
 	// outRPC returns the current RPC message if it will fit sizeToAdd more bytes
 	// otherwise, it will create a new RPC message and add it to the list.
-	// if withCtl is true, the new RPC message will have a non-nil empty Control message.
+	// if withCtl is true, the returned message will have a non-nil empty Control message.
 	outRPC := func(sizeToAdd int, withCtl bool) *RPC {
 		current := rpcs[len(rpcs)-1]
-		if current.Size()+sizeToAdd < limit {
+		// check if we can fit the new data, plus an extra byte for the protobuf field tag
+		if current.Size()+sizeToAdd+1 < limit {
+			if withCtl && current.Control == nil {
+				current.Control = &pb.ControlMessage{}
+			}
 			return current
 		}
 		var ctl *pb.ControlMessage
@@ -982,9 +986,6 @@ func fragmentRPC(rpc *RPC, limit int) ([]*RPC, error) {
 	}
 
 	// we need to split up the control messages into multiple RPCs
-	// add a blank rpc message to the end of the list, then use outRPC to get or create
-	// RPC messages to fit each control message
-	rpcs = append(rpcs, &RPC{RPC: pb.RPC{Control: &pb.ControlMessage{}}, from: rpc.from})
 	for _, graft := range ctl.Graft {
 		out := outRPC(graft.Size(), true)
 		out.Control.Graft = append(out.Control.Graft, graft)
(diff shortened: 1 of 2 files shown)


vyzo:
test failure with data race!

vyzo:
And don't worry too much about failing if an individual message is larger than the maximum, at best we can warn.

gossipsub.go
```diff
@@ -899,18 +899,91 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
+			// push control messages that need to be retried
+			ctl := rpc.GetControl()
+			if ctl != nil {
+				gs.pushControl(p, ctl)
+			}
+		}
+	}
+}
+
+func fragmentRPC(rpc *RPC, limit int) ([]*RPC, error) {
```
vyzo, line 939: I think we should separate the control messages in their own RPC if we are over size -- are we counting those? we could overflow because of them.

gossipsub.go
```diff
@@ -899,18 +899,91 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
+
+	c := (rpc.Size() / limit) + 1
+	rpcs := make([]*RPC, 1, c)
+	rpcs[0] = &RPC{RPC: pb.RPC{Control: &pb.ControlMessage{}}, from: rpc.from}
+	outRPC := func(sizeToAdd int) *RPC {
+		current := rpcs[len(rpcs)-1]
+		if current.Size()+sizeToAdd < limit {
+			return current
+		}
+		next := &RPC{RPC: pb.RPC{Control: &pb.ControlMessage{}}, from: rpc.from}
```
vyzo: we shouldn't need a control message here.

gossipsub.go
```diff
@@ -899,18 +899,109 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
+	rpcs := make([]*RPC, 1, c)
+	rpcs[0] = &RPC{RPC: pb.RPC{}, from: rpc.from}
+
+	// outRPC returns the current RPC message if it will fit sizeToAdd more bytes
+	// otherwise, it will create a new RPC message and add it to the list.
+	// if withCtl is true, the new RPC message will have a non-nil empty Control message.
+	outRPC := func(sizeToAdd int, withCtl bool) *RPC {
+		current := rpcs[len(rpcs)-1]
+		if current.Size()+sizeToAdd < limit {
+			return current
```
vyzo: I think we need one extra byte for the protobuf tag.

gossipsub_test.go
```diff
@@ -1717,3 +1717,145 @@ func (ps *mockPeerScoreInspector) score(p peer.ID) float64 {
...
+	w := ggio.NewDelimitedWriter(os)
+	truth := true
+	topic := "test"
+	err = w.WriteMsg(&pb.RPC{Subscriptions: []*pb.RPC_SubOpts{&pb.RPC_SubOpts{Subscribe: &truth, Topicid: &topic}}})
+
+	remotePidBytes, err := s.Conn().RemotePeer().MarshalBinary()
+	if err != nil {
+		panic(err)
+	}
+	toPrune := []*pb.PeerInfo{{PeerID: remotePidBytes}}
```
vyzo: are we sending it back its own self as a peer? we can just send nothing.

vyzo (commented):
looks good, just a couple of comments.

gossipsub.go
```diff
@@ -899,18 +899,149 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
 		return
 	}
 
-	select {
-	case mch <- out:
-		gs.tracer.SendRPC(out, p)
-	default:
-		log.Infof("dropping message to peer %s: queue full", p)
+	// fragment into multiple RPCs if necessary
+	outRPCs, err := fragmentRPC(out, gs.p.maxMessageSize)
```
vyzo, line 909: I wonder if we can avoid allocating the array for the common case (fast path) of not needing fragmentation.
Perhaps move the size limit check outside `fragmentRPC`?

gossipsub.go
```diff
@@ -899,18 +899,149 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
+			if ctl != nil {
+				gs.pushControl(p, ctl)
+			}
+		}
+	}
+}
+
+func fragmentRPC(rpc *RPC, limit int) ([]*RPC, error) {
+	if rpc.Size() < limit {
+		return []*RPC{rpc}, nil
```
vyzo, line 941: I am talking about this slice allocation here.

gossipsub_test.go
```diff
@@ -1717,3 +1717,305 @@ func (ps *mockPeerScoreInspector) score(p peer.ID) float64 {
...
+type iwantEverything struct {
+	h                host.Host
+	lk               sync.Mutex
+	rpcsWithMessages int
+	msgsReceived     int
+	ihavesReceived   int
+}
+
+func (iwe *iwantEverything) handleStream(s network.Stream) {
+	os, err := iwe.h.NewStream(context.Background(), s.Conn().RemotePeer(), GossipSubID_v10)
```
vyzo, line 1792: we should probably add a `defer s.Close()` here (not that it matters much, but let's be nice)

vyzo (commented):
This is looking great, and has great coverage.

Only (minor) issue remaining is avoiding the allocation of the RPC array in the fast path, let's not put undue pressure on the garbage collector!

gossipsub.go
yusefnapora: That's a good idea 👍

gossipsub.go
```diff
@@ -899,18 +899,158 @@ func (gs *GossipSubRouter) sendRPC(p peer.ID, out *RPC) {
...
 	}
+
+	// If we're below the max message size, go ahead and send
+	if out.Size() < gs.p.maxMessageSize {
+		select {
+		case mch <- out:
+			gs.tracer.SendRPC(out, p)
+		default:
+			dropRPC(out, "dropping message to peer %s: queue full", p)
+		}
```
vyzo: let's make a function for this, call it `GossipSubRouter.doSendRPC`, as the code is duplicated below.

vyzo (commented):
Great! Let's make a `doSendRPC` function for the duplicate code and it's ready to merge!

vyzo:
Mind rebasing on master as well so that we can merge?

vyzo (approved):

PR #323, opened 2020-05-06
Better handling of PRUNE backoff

### LINKED ISSUE #286: Notify pruned peers of the backoff period

vyzo:
Changes in the backoff logic:
- Extend the PRUNE protobuf message to include the backoff (in seconds)
- Notify the peer of the backoff period when pruning.
- Respect the peer specified backoff period when pruned.
- If a peer doesn't respect the backoff, it is penalized through P7; if the regrafting comes before the flood threshold, it is penalized twice.
- Remove the ad-hoc BackoffPenalty/sinkholing in favour of P7 penalties and graylisting.

Closes #286 

cc @raulk @daviddias


yusefnapora (approved):
This looks good - I think it would be nice to mention the backoff penalty in the godoc comment in `score_params.go` where we define P7. Something like

```
	// This parameter has an associated counter which tracks misbehaviour as detected by the
	// router. The router currently applies penalties for the following behaviors: 
  // - attempting to re-graft before the prune backoff time has elapsed
```

And we can add to the list when we start penalizing for scam IWANTs (#285).

vyzo:
Sure, will do!

PR #324, opened 2020-05-07
Probabilistically track in-flight IWANT requests and apply penalty for broken promises

### LINKED ISSUE #285: Track in-flight IWANT requests

vyzo:
Closes #285 

cc @raulk @daviddias

### DIFF
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -1,0 +1,59 @@
+package pubsub
+
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+// gossipTracer is an internal tracer that tracks IWANT requests in order to penalize
+// peers who don't follow up on IWANT requests after an IHAVE advertisement.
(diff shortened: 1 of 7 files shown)


vyzo:
Added some tests to explicitly test the gossip tracer and behavioural penalties.

yusefnapora (approved):
Looks good. 

I like the change from a single `scoreTracer` to using a slice of `internalTracer`s - that's nice and flexible and could maybe be useful in unit tests to capture delivery events, etc.

PR #326, opened 2020-05-08
fix: duplicate topic label name

alanshaw:
Also note that there are multiple references to testground in here - testground, sidecar etc. I assume not intentional?


vyzo:
Yeah, it's not intentional at all -- raul probably copied this everywhere.
Can you clean it up a bit?

Stebalien:
Closing as this is a year old.

PR #328, opened 2020-05-11
decaying connmgr tags for message delivery

### LINKED ISSUE #306: Implement decaying connection manager tags for first time publishers
This will ensure that recent publishers and gossipers remain connected even at the face of a connection flood and subsequent connection manager action.

Requirements:
- [ ] https://github.com/libp2p/go-libp2p-core/pull/104
- [ ] https://github.com/libp2p/go-libp2p-connmgr/pull/61

yusefnapora:
This adds decaying connection manager tags that give a boost to peers that deliver messages. closes #306 

I was hoping to finish and fix the test over the weekend, but didn't get a chance. It seems to basically work, but some of the honest peers aren't connected at the end of the test. 

I'll sort it out in the morning, just wanted to get the branch up in case @vyzo wants to look through before then.


gossipsub.go
```diff
@@ -239,6 +249,7 @@ type GossipSubRouter struct {
 	mesh     map[string]map[peer.ID]struct{}  // topic meshes
 	fanout   map[string]map[peer.ID]struct{}  // topic fanout
 	lastpub  map[string]int64                 // last publish time for fanout topics
+	tags     map[string]connmgr.DecayingTag   // decaying connection manager tags for each topic
```
vyzo: shouldn't this live inside the `tagTracer`?

gossipsub.go
```diff
@@ -866,12 +869,12 @@ func (gs *GossipSubRouter) Leave(topic string) {
 	gs.tracer.Leave(topic)
 
 	delete(gs.mesh, topic)
+	delete(gs.tags, topic)
```
vyzo: this should probably go inside the `tagTracer`

tag_tracer.go
```diff
@@ -0,0 +1,193 @@
...
+	// GossipSubConnTagBumpMessageDelivery is the amount to add to the connection manager
+	// tag that tracks message deliveries. Each time a peer is the first to deliver a
+	// message within a topic, we "bump" a tag by this amount, up to a maximum
+	// of GossipSubConnTagMessageDeliveryCap.
+	// Note that the delivery tags decay over time, decreasing by GossipSubConnTagDecayAmount
+	// at every GossipSubConnTagDecayInterval.
+	GossipSubConnTagBumpMessageDelivery = 1
+
+	// GossipSubConnTagDecayInterval is the decay interval for decaying connection manager tags.
+	GossipSubConnTagDecayInterval = time.Minute
```
vyzo: let's make this a little longer if it is to decrement by integer leaps.

vyzo: may be `10*time.Minute`?

tag_tracer.go
```diff
@@ -0,0 +1,193 @@
...
+	// bump up to max of GossipSubConnTagMessageDeliveryCap
+	bumpFn := func(value connmgr.DecayingValue, delta int) (after int) {
+		val := value.Value + delta
+		if val > GossipSubConnTagMessageDeliveryCap {
+			return GossipSubConnTagMessageDeliveryCap
+		}
+		return val
+	}
+
+	return t.decayer.RegisterDecayingTag(name, GossipSubConnTagDecayInterval, decayFn, bumpFn)
```
vyzo: I think we need a nil check (probably at the beginning of the function) for this, we might panic if the connmgr doesn't support decaying tags.

yusefnapora:
@vyzo the tag tracer is still missing "near first" message deliveries. Can you think of a good way to do this without duplicating the delivery record tracking from the score tracer?

vyzo:
I think we can get away very simply by starting tracking in `Validate` and stopping all tracking when the message is `Delivered` or `Rejected`. The `Duplicate` should only fire while still validating.
So basically we can have a lightweight version of what we have in the score tracer.

yusefnapora:
I squashed a few things and marked this as "ready for review", but we still need to wait until the dependency PRs are merged.

tag_tracer.go
```diff
@@ -0,0 +1,298 @@
...
+	drec.status = deliveryValid
+	drec.validated = time.Now()
+
+	for p := range drec.peers {
+		// this check is to make sure a peer can't send us a message twice and get a double count
+		// if it is a first delivery.
+		if p != msg.ReceivedFrom {
+			nearFirstPeers = append(nearFirstPeers, p)
+		}
+	}
```
vyzo: We might want to nil the `drec.peers` map before we return the near first peers, so as to avoid holding onto memory unnecessarily. It shouldn't be used again after the delivery status has been marked as `deliveryValid`.

vyzo (commented):
this looks good, just a comment.

Note: we'll want to `go mod tidy` this when it's time to merge.

tag_tracer.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// defensive check that this is the first rejection trace -- delivery status should be unknown
+	if drec.status != deliveryUnknown {
+		log.Warnf("unexpected rejection trace: message from %s was first seen %s ago and has delivery status %d", msg.ReceivedFrom, time.Now().Sub(drec.firstSeen), drec.status)
+		return
+	}
+
+	drec.status = deliveryInvalid
+	drec.peers = nil
+}
```
vyzo: actually, there are some cases where the action (marking as invalid) is incorrect; if the rejection is because of invalid or missing signature or validation queue full we shouldn't clear the delivery record. This is because the rejection happens without entering the validation pipeline and we can end up rejecting a message that is still validating.

So let's add a check here for the reason and short-circuit if it is one of the above; see also the handling in score.go

vyzo: Might even make sense to check explicitly for one of validation throttled, validation ignored, validation rejected as these are the only ones that indicate a message that was in the validation queue.
Then we can set the delivery status to a known value (not necessarily invalid).

tag_tracer.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	// bump up to max of GossipSubConnTagMessageDeliveryCap
+	bumpFn := func(value connmgr.DecayingValue, delta int) (after int) {
+		val := value.Value + delta
+		if val > GossipSubConnTagMessageDeliveryCap {
+			return GossipSubConnTagMessageDeliveryCap
+		}
+		return val
+	}
```
raulk: I'm moving the [preset decay and bump functions](https://github.com/libp2p/go-libp2p-connmgr/pull/61/files#diff-7167ce1adea9dc2edd55d7253ca2d1f5) from `go-libp2p-connmgr` into `go-libp2p-core` so that you can reuse them here without introducing an additional module dependency.

What you've implemented here is basically the `FixedDecay` decay function, and the `SumBounded` bump function.

raulk: Done. Mind adopting the preset functions here? https://github.com/libp2p/go-libp2p-core/pull/104/commits/d3277047d6ca1eae46b03c4816ca00e696b114e8

tag_tracer_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+import (
+	"fmt"
+	"github.com/benbjohnson/clock"
+	connmgr "github.com/libp2p/go-libp2p-connmgr"
+	connmgri "github.com/libp2p/go-libp2p-core/connmgr"
+	"github.com/libp2p/go-libp2p-core/peer"
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"testing"
+	"time"
```
raulk: import order

gossipsub_connmgr_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"context"
+	"github.com/benbjohnson/clock"
+	"github.com/libp2p/go-libp2p-core/host"
+	swarmt "github.com/libp2p/go-libp2p-swarm/testing"
+	"testing"
+	"time"
+
+	bhost "github.com/libp2p/go-libp2p-blankhost"
+	connmgr "github.com/libp2p/go-libp2p-connmgr"
+	"github.com/libp2p/go-libp2p-core/peer"
```
raulk: nit: import order.

tag_tracer.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func newTagTracer(cmgr connmgr.ConnManager) *tagTracer {
+	decayer, ok := connmgr.SupportsDecay(cmgr)
+	if !ok {
+		log.Warnf("connection manager does not support decaying tags, delivery tags will not be applied")
+	}
+	return &tagTracer{
+		cmgr:      cmgr,
+		msgID:     DefaultMsgIdFn,
```
raulk, line 75: Do we always want to use the default? shouldn't we use the message ID function that the user has set?

tag_tracer.go
```diff
@@ -0,0 +1,265 @@
...
+		log.Warnf("unable to create decaying delivery tag: %s", err)
+		return
+	}
+	t.decaying[topic] = tag
+}
+
+func (t *tagTracer) removeDeliveryTag(topic string) {
+	t.Lock()
+	defer t.Unlock()
+	delete(t.decaying, topic)
```
raulk: Shouldn't we have a function for erasing a decaying tag explicitly?

tag_tracer.go
yusefnapora: We set it to the router's `msgID` func in the `Start` method. Although now that I removed the background process we don't really need a `Start` method and could just put it in the constructor instead.

vyzo: The actual MsgID function is not known when it is constructed, it might be changed by the options.
So keep the Start method :)

yusefnapora: vyzo reminded me that we should keep it in `Start`, since the message id is set in an option that might be applied after the `tagTracer` is created.

tag_tracer.go
vyzo: We don't actually use that though. When would we want to remove a tag? Perhaps when we have a rejection with validationFailed?

tag_tracer.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (t *tagTracer) DuplicateMessage(msg *Message) {
+	t.Lock()
+	defer t.Unlock()
+
+	id := t.msgID(msg.Message)
+	peers, ok := t.nearFirst[id]
+	if !ok {
+		// the peers map should have been created in ValidateMessage
```
vyzo: that comment is not technically accurate and is confusing; the duplicate might arrive after we have Delivered/Rejected the message.

tag_tracer.go
vyzo: Actually there is a case we do want to delete the tag when we remove it: imagine Joining, Leaving, and then Joining again a topic; that would lead to an error when we try to recreate the tag.
So yeah, you are right, we have to delete it.

raulk: https://github.com/libp2p/go-libp2p-core/pull/151 and https://github.com/libp2p/go-libp2p-connmgr/pull/72 are merged.

Will wait til you integrate those features here by importing master. Will make a release once you give me the thumbs up.

yusefnapora:
hmm, I added another 5 minutes to the travis timeout, but we're still hitting it. I can't imagine the tests in this branch take an extra 5m to run... I wonder if I could be deadlocking somewhere?

yusefnapora:
okay, I finally wised up and figured out how to enable the race detector by default when I run tests in IntelliJ. With it enable, we hit the max goroutine limit in `TestGossipsubConnTagMessageDeliveries`. But reducing the number of peers in the test seems to make it happy.

vyzo:
we should be getting output in that case.

yusefnapora:
hmm, that's true. It did also take a few minutes on my laptop before it failed, so it might actually be taking more than 5m on travis

yusefnapora:
I got the end-to-end test passing by lowering the number of test peers again, but now the unit test is failing, and it's a little confusing.

```
=== RUN   TestTagTracerDeliveryTags
    TestTagTracerDeliveryTags: tag_tracer_test.go:138: expected delivery tag value = 10, got 11
    TestTagTracerDeliveryTags: tag_tracer_test.go:145: expected delivery tag value = 0, got 1
```

This is just after adding 50 minutes to the fake clock, and checking that the tag value has decreased.

On my laptop, it gets the expected value, but it always seems to be off by one on Travis. The reason I'm confused is that the result should only depend on the fake clock we pass into the decayer, not the real clock. I tried sleeping the real clock for a second after bumping the fake one, to give the Decayer's processing routine time to run, but it still fails.

I'm going to try increasing the sleep time to five seconds, but I'll probably be asleep myself by the time it's done running. Anyway, Vyzo, perhaps you could take a look if it's still failing.

.travis.yml
```diff
@@ -18,7 +18,7 @@ cache:
 
 script:
   - go build ./...
-  - travis_wait 30 go test -timeout 30m -v -race -coverprofile=coverage.txt -covermode=atomic ./...
+  - travis_wait 35 go test -timeout 35m -v -race -coverprofile=coverage.txt -covermode=atomic ./...
```
vyzo: let's restore this to 30m, it doesn't seem to change anything.

go.mod
```diff
@@ -1,17 +1,24 @@
 module github.com/libp2p/go-libp2p-pubsub
 
 require (
+	github.com/benbjohnson/clock v1.0.1
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v1.0.4
-	github.com/libp2p/go-libp2p-blankhost v0.1.4
-	github.com/libp2p/go-libp2p-core v0.5.3
+	github.com/ipfs/go-log/v2 v2.0.7 // indirect
```
vyzo: let's not forget to `go mod tidy`.

vyzo (commented):
Hard to tell what's happening with the tests on travis, code looks good to me.

tag_tracer.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+// nearFirstPeers returns the peers who delivered the message while it was still validating
+func (t *tagTracer) nearFirstPeers(msg *Message) []peer.ID {
+	t.Lock()
+	defer t.Unlock()
+	peersMap, ok := t.nearFirst[t.msgID(msg.Message)]
+	if !ok {
+		return []peer.ID{}
```
vyzo: returning just nil will do.

vyzo (approved):
at last, green checkmaks!

PR #333, opened 2020-05-13
fix break bug in validation

fxfactorial:
break needs a loop label or to return.


vyzo (changes requested):
Indeed this is a (minor) bug, please use a break with label instead of returning.

fxfactorial:
@vyzo done - got another one , same thing

vyzo (approved):
thank you!

PR #339, opened 2020-05-16
Implement outbound peer quotas and limit overgraft

vyzo:
Introduces a new overlay parameter, `D_out`, which is used to control the number of outbound connected peers we should keep in the mesh.

During mesh maintenance:
- when pruning because of over subscription, we constrain the survivor set to include D_out outbound peers.
- when we have less than `D_out` peers in the mesh, we graft some outbound peers.

In addition, when receiving a graft over the `D_hi` limit, the graft is rejected unless it comes from an outbound peer.

cc @daviddias @raulk

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -356,10 +356,21 @@ func (gs *GossipSubRouter) AddPeer(p peer.ID, proto protocol.ID) {
 	// track the connection direction
 	outbound := false
 	conns := gs.p.host.Network().ConnsToPeer(p)
+loop:
 	for _, c := range conns {
 		if c.Stat().Direction == network.DirOutbound {
-			outbound = true
-			break
+			// only count the connection if it has a pubsub stream
+			for _, s := range c.GetStreams() {
+				switch s.Protocol() {
+				case FloodSubID:
+					fallthrough
+				case GossipSubID_v10:
+					fallthrough
+				case GossipSubID_v11:
+					outbound = true
+					break loop
+				}
+			}
 		}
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -348,6 +352,17 @@ func (gs *GossipSubRouter) AddPeer(p peer.ID, proto protocol.ID) {
 	if direct {
 		gs.p.host.ConnManager().TagPeer(p, "pubsub:direct", 1000)
 	}
+
+	// track the connection direction
+	outbound := false
+	conns := gs.p.host.Network().ConnsToPeer(p)
+	for _, c := range conns {
```
yusefnapora, line 360: This is perhaps overly paranoid, but is it possible to only count the connection used for the gossipsub stream? Otherwise there's an outside chance someone could trick us into open some other libp2p conn to them and get counted as outbound. Not sure how, exactly, although if we're running the autonat service they could request a dial-back.

vyzo: I guess we could, but it sounds too paranoid?

vyzo: we can however.

vyzo: Also, autonat dialback is not an issue -- we don't use the same host for the dial back.

vyzo: done.

yusefnapora (approved):
looks good. thanks for humoring my paranoia 😄

vyzo:
rebased on master for merging.

PR #340, opened 2020-05-16
fix panic in timecache-backed blacklist

### LINKED ISSUE #338: remove panic inducing depencency
I use: 

```go
blklist, _ := libp2p_pubsub.NewTimeCachedBlacklist(15 * time.Minute)
libp2p_pubsub.WithBlacklist(blklist),
pubsub, err := libp2p_pubsub.NewGossipSub(ctx, p2pHost, options...)
```

and use the `pubsub.BlacklistPeer(peer)` but the upstream dep on `whyrusleeping` imo is bad because banning peers more than once, the same peer, causes a panic on add, so to use it right, I'd need to hold the ref to the blacklist, call contained before calling add, kinda too much of a hassle to accomdate this API. 

```
func (tc *TimeCache) Add(s string) {
	_, ok := tc.M[s]
	if ok {
		panic("putting the same entry twice not supported")
	}
```

from timecache.go 

Please implement blacklist without a panic causing lib

fxfactorial:
closes #338

### DIFF
--- a/blacklist.go
+++ b/blacklist.go
@@ -10,7 +10,7 @@ import (
 
 // Blacklist is an interface for peer blacklisting.
 type Blacklist interface {
-	Add(peer.ID)
+	Add(peer.ID) bool
 	Contains(peer.ID) bool
 }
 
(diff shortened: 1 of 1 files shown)


vyzo (approved):

fxfactorial:
ping - possible to have merged/released today - would like to use in my code

vyzo:
done - thank you!

fxfactorial:
Cool , thank you , I love libp2p

PR #344, opened 2020-05-25
Fix bug in oversubscription pruning D_out constraint

vyzo:
If some outbound peers were already part of the selection and near the end of it, they could slide off when rotating in outbound peers from outside the selection.
This fixes the issue, by doing a first pass and rotating to the front outbound peers already part of the selection.


yusefnapora (approved):
nice catch 👍

PR #345, opened 2020-05-27
feat: add direct connect ticks option

alanshaw:
In [drand](https://github.com/drand/drand) we have a gossipsub relay to allow users to subscribe to getting random values over pubsub. We want to support pure gossip relays who relay from a relay. For this we need direct peering agreements and want to mitigate the possibility of "missing" randomness messages by ensuring the direct connect ticks period is less than the period between randomness messages.

This PR simply adds a new functional option allowing us to set the direct connect ticks value without modifying the global variable.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -305,6 +305,21 @@ func WithDirectPeers(pis []peer.AddrInfo) Option {
 	}
 }
 
+// WithDirectConnectTicks is a gossipsub router option that sets the number of
+// heartbeat ticks between attempting to reconnect direct peers that are not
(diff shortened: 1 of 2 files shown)


vyzo:
There is already a variable that you can set: `GossipSubDirectConnectTicks`.
Do you really need a functional option?

vyzo (approved):
This is all fine, if you can't set the global variable.

vyzo:
Mind rebasing on master?

alanshaw:
@vyzo done :)

PR #348, opened 2020-06-03
Protect direct and mesh peers in the connection manager

vyzo:
Almost all the services are now starting to protect essential peers, so we should do the same.
This adds protection for direct and mesh peers.

~An unfortunate side effect of this change is that we have to (temporarily I hope) disable the tests for the tags, as the connection manager doesn't offer an interface to query protected peers.~
edit: resolved! the tests are back.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ require (
 	github.com/benbjohnson/clock v1.0.2
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v1.0.4
+	github.com/libp2p/go-eventbus v0.2.1 // indirect
 	github.com/libp2p/go-libp2p-blankhost v0.1.6
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
 	github.com/libp2p/go-libp2p-core v0.5.7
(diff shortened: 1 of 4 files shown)


vyzo:
cc @raulk

vyzo:
Implementation of protection status queries in https://github.com/libp2p/go-libp2p-core/pull/158 and https://github.com/libp2p/go-libp2p-connmgr/pull/76

Stebalien:
What are the limits on mesh peers? That is, can I keep grafting to keep getting my connection protected?

vyzo:
We only accept outbound connections for grafting when over the `Dhi` limit, and that fixes itself within the next heartbeat.
So at best excess peers would get protection for up to 1s.

vyzo:
Note that once pruned you can't get regrafted unless the backoff elapses, which is 1minute by default.

vyzo:
reinstated the tests, using the new connection manager interface.

yusefnapora (approved):
looks good - nice job getting the test issue sorted

Stebalien (approved):

PR #353, opened 2020-06-15
only do PX on leave if PX was enabled in the node

### LINKED ISSUE #349: Question: is pubsub discovery enabled by default (or some parts of it)
I know it shouldn't have been enabled as the only option enabled is:
```GO
optsPS := []pubsub.Option{
	pubsub.WithMessageSigning(withMessageSigning),
}
```
The question was raised as I saw in one of my tests employing 8 hosts with attached pubsubs instances that where connected manually: x connects with y and z, z with a and b and so on. After about 14 seconds in my test, I saw that the connections changed (some connections appeared between peers that were not connected manually). I have commented out any pubsub usage (including publishing on topics, initialization and so on) and I have observed that my connections did not changed (remained as initially done).

We are using github.com/libp2p/go-libp2p-pubsub v0.3.1
Running in debug mode, I saw that the `disc` instance is instantiated with `&discover{}` and the call to Start returns on the first if. I can not figure out where are the extra connections whenever I turn on the pubsub instance. The thing reproduce even if there are no messages broadcast on the network. However extra connections do not appear if any of the pubsub instances did not joined a topic.

vyzo:
Closes #349

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1005,7 +1005,7 @@ func (gs *GossipSubRouter) sendGraft(p peer.ID, topic string) {
 }
 
 func (gs *GossipSubRouter) sendPrune(p peer.ID, topic string) {
-	prune := []*pb.ControlPrune{gs.makePrune(p, topic, true)}
+	prune := []*pb.ControlPrune{gs.makePrune(p, topic, gs.doPX)}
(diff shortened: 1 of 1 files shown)


aschmahmann (approved):
Seems reasonable to me.

PR #354, opened 2020-06-18
Fragmentation for the hello packet

### LINKED ISSUE #402: Handling large number of topic subscriptions
Currently if a node subscribes to enough topics such that the RPC that is used to send topic subscription announcements exceeds the maximum message size (defaults to 1MiB) then we'll silently fail and stop announcing any of our topics to newly connected peers.

We should more gracefully error, or inform the user, when they've exceeded the maximum number of topics they can subscribe to and additionally it'd be nice to have a mechanism in place to allow for a larger number of subscriptions.

- [ ] (Beginner) Error when exceeding maximum topic limit
  - Easiest: Just check if the (hello packet) message size is too large and log an error
  - More correct: When subscribing to a new topic check if the new topic would force the hello packet to exceed the maximum size and if so return an error instead of allowing the topic to be subscribed to.
- [ ] (Expert) Allow sending more topics than fit in a single message

requilence:
We have the node with a lot of pubsub subscriptions and observing this problem:
- The node sends helloPacket to all new peers
- This packet exceeds the maximum buffer size on clients
- Connection resets
- Repeats from the beginning

As a result, pubsub generates 20Mbit outbound 😬

This PR uses fragmentRPC introduced in #320 to split the helloPacket payload

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -131,12 +131,19 @@ func (p *PubSub) handleSendingMessages(ctx context.Context, s network.Stream, ou
 			if !ok {
 				return
 			}
-
-			err := writeMsg(&rpc.RPC)
+			rpcFragments, err := fragmentRPC(rpc, p.maxMessageSize)
 			if err != nil {
 				s.Reset()
-				log.Infof("writing message to %s: %s", s.Conn().RemotePeer(), err)
+				log.Infof("unable to fragment RPC: %s", err)
 				return
+			}
+			for _, rpcFragment := range rpcFragments {
+				err := writeMsg(rpcFragment)
+				if err != nil {
+					s.Reset()
+					log.Infof("writing message to %s: %s", s.Conn().RemotePeer(), err)
+					return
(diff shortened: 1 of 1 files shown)


jacobheun:
@requilence can you look into the failing tests?

@vyzo can you give a quick review of this

vyzo (commented):
there is some complexity with the subscription retransmissions; let me think about this a bit.

vyzo (changes requested):
the correct way to implement this is to retry all failed transmissions (perhaps the queue got full because there were too many of them), until the hello packet is completely sent (with all fragmented messages) or an error occurs at which point the stream has to be aborted.

BigLep:
@vyzo : will close and link to the appropriate issue.

BigLep:
@vyzo : did you create the appropriate issue so we can close this?

vyzo:
@BigLep We already have #402.

PR #356, opened 2020-07-06
Define a public error for sub.Next() on a cancelled subscription

prestonvanloon:
We find this error happens often in Prysm with our frequently changing pubsub topic subscriptions. We would like to ignore this error by referencing the publicly defined ErrSubscriptionCancelled.

Context: https://github.com/prysmaticlabs/prysm/issues/6449

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -31,7 +31,7 @@ var (
 
 	// ErrSubscriptionCancelled may be returned when a subscription Next() is called after the
 	// subscription has been cancelled.
-	ErrSubscriptionCancelled = errors.New("subscription cancelled by calling sub.Cancel()")
+	ErrSubscriptionCancelled = errors.New("subscription cancelled")
 )
 
 var log = logging.Logger("pubsub")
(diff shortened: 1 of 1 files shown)


pubsub.go
```diff
@@ -28,6 +28,10 @@ const DefaultMaxMessageSize = 1 << 20
 
 var (
 	TimeCacheDuration = 120 * time.Second
+
+	// ErrSubscriptionCancelled may be returned when a subscription Next() is called after the
+	// subscription has been cancelled.
+	ErrSubscriptionCancelled = errors.New("subscription cancelled by calling sub.Cancel()")
```
vyzo: let's call the error message just "subscription cancelled"; no need to perpetuate a bad error message.

vyzo (commented):
LGTM, let's just improve on the error message.

pubsub.go
prestonvanloon:
```suggestion
	ErrSubscriptionCancelled = errors.New("subscription cancelled")
```

vyzo (approved):
thank you!

PR #360, opened 2020-07-22
Extended peer score inspection

vyzo:
aids debugging negative scores

### DIFF
--- a/score.go
+++ b/score.go
@@ -78,6 +78,7 @@ type peerScore struct {
 
 	// debugging inspection
 	inspect       PeerScoreInspectFn
+	inspectEx     ExtendedPeerScoreInspectFn
 	inspectPeriod time.Duration
 }
 
@@ -114,12 +115,33 @@ const (
 )
 
 type PeerScoreInspectFn func(map[peer.ID]float64)
+type ExtendedPeerScoreInspectFn func(map[peer.ID]*PeerScoreSnapshot)
+
+type PeerScoreSnapshot struct {
+	Score              float64
+	Topics             map[string]*TopicScoreSnapshot
+	AppSpecificScore   float64
+	IPColocationFactor int
+	BehaviourPenalty   float64
+}
+
+type TopicScoreSnapshot struct {
+	TimeInMesh               time.Duration
+	FirstMessageDeliveries   float64
+	MeshMessageDeliveries    float64
+	InvalidMessageDeliveries float64
+}
 
 // WithPeerScoreInspect is a gossipsub router option that enables peer score debugging.
 // When this option is enabled, the supplied function will be invoked periodically to allow
-// the application to inspec or dump the scores for connected peers.
+// the application to inspect or dump the scores for connected peers.
+// The supplied function can have one of two signatures:
+//  - PeerScoreInspectFn, which takes a map of peer IDs to score.
+//  - ExtendedPeerScoreInspectFn, which takes a map of peer IDs to
+//    PeerScoreSnapshots and allows inspection of individual score
+//    components for debugging peer scoring.
 // This option must be passed _after_ the WithPeerScore option.
-func WithPeerScoreInspect(inspect PeerScoreInspectFn, period time.Duration) Option {
+func WithPeerScoreInspect(inspect interface{}, period time.Duration) Option {
 	return func(ps *PubSub) error {
 		gs, ok := ps.rt.(*GossipSubRouter)
 		if !ok {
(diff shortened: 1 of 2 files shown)


score.go
```diff
@@ -130,7 +152,19 @@ func WithPeerScoreInspect(inspect PeerScoreInspectFn, period time.Duration) Opti
 			return fmt.Errorf("peer scoring is not enabled")
 		}
 
-		gs.score.inspect = inspect
+		switch i := inspect.(type) {
+		case PeerScoreInspectFn:
+			gs.score.inspect = i
+		case func(map[peer.ID]float64):
```
raulk: Alternatively you could change these to type aliases, then you wouldn't need the explicit type-casting.

score.go
```diff
@@ -130,7 +152,19 @@ func WithPeerScoreInspect(inspect PeerScoreInspectFn, period time.Duration) Opti
...
+			gs.score.inspect = PeerScoreInspectFn(i)
+		case ExtendedPeerScoreInspectFn:
+			gs.score.inspectEx = i
+		case func(map[peer.ID]*PeerScoreSnapshot):
+			gs.score.inspectEx = ExtendedPeerScoreInspectFn(i)
```
raulk: Same as above.

score.go
```diff
@@ -320,6 +354,14 @@ func (ps *peerScore) background(ctx context.Context) {
 
 // inspectScores dumps all tracked scores into the inspect function.
 func (ps *peerScore) inspectScores() {
+	if ps.inspect != nil {
```
raulk, line 369: The `WithPeerScoreInspect` option could be called twice, once with a simple inspector, once with an extended inspector. That wouldn't error and would lead to setting both inspectors, yet at runtime we would only call the simple one.

Either we should error on multiple `WithPeerScoreInspect` calls, or we should remove the `else` in this `if` block and turn it into a separate `if`, in order to call both inspectors.

score.go
```diff
@@ -334,6 +376,45 @@ func (ps *peerScore) inspectScores() {
...
+		if len(pstats.topics) > 0 {
+			pss.Topics = make(map[string]*TopicScoreSnapshot, len(pstats.topics))
+			for t, ts := range pstats.topics {
+				pss.Topics[t] = &TopicScoreSnapshot{
+					FirstMessageDeliveries:   ts.firstMessageDeliveries,
+					MeshMessageDeliveries:    ts.meshMessageDeliveries,
+					InvalidMessageDeliveries: ts.invalidMessageDeliveries,
+				}
+				if ts.inMesh {
+					pss.Topics[t].TimeInMesh = ts.meshTime
```
raulk: nit: instantiate the `TopicScoreSnapshot` as a local var, populate it, then set it _once_ in the map. Maybe the compiler optimises this double access, but I'm not sure, and this will be called quite frequently.

score.go
```diff
@@ -334,6 +376,45 @@ func (ps *peerScore) inspectScores() {
...
+		for _, ip := range pstats.ips {
+			_, whitelisted := ps.params.IPColocationFactorWhitelist[ip]
+			if whitelisted {
+				continue
+			}
+			peersInIP := len(ps.peerIPs[ip])
+			if peersInIP > ps.params.IPColocationFactorThreshold {
+				surpluss := peersInIP - ps.params.IPColocationFactorThreshold
+				pss.IPColocationFactor += surpluss
+			}
```
raulk: Isn't this the code to _actually_ calculate the score? If so, we should encapsulate it in a function, so it's not duplicated and there's no risk of falling out of sync if we ever update the logic.

raulk (changes requested):
Few minor bugs and suggestions.

score.go
vyzo: done, made them type aliases.

score.go
vyzo: done.

score.go
vyzo: I think allowing multiple invocations of `WithPeerScoreInspect` might be a mistake because there are conflicts with the time period, let's disallow.
Regardless, I will remove the else as it is confusing.

vyzo: done.

score.go
vyzo: sure, will do.

vyzo: done.

score.go
vyzo: yeah, I can extract to a function.

vyzo: done.

raulk (approved):

Kubuxu (approved):

Issue #361, opened 2020-07-24
go-ipfs gossipsub produces invalid protobuf messages


rklaehn:
I am trying a rust-gossipsub based program to work with go-ipfs. I was having some strange intermittent problems.

After 3 days I nailed it down to this: go-ipfs under some circumstances produces protobuf messages that do not conform to the protobuf format used in go-ipfs.

The gossipsub protobuf has message ids as strings since a long time. See e.g. https://github.com/libp2p/go-libp2p-pubsub/blob/b565e5939a8407ba7824d59e9cb28b9b453032ac/pb/rpc.proto#L33

But I am getting messages that are overall valid protobuf messages, but where the messageid is not a valid utf8 byte string.

```
$ echo 1ac9010ac6010a142f6b6c6b2f70726f642f323032302d30312d3039122a1220d024d0c98917394729c109ddf8d94c98d36ff3899b7f3f6d148b8e6baf084dc316236b1c66ea8a8e122a12202fe480d71af77f741f0a3abf054e30c5d414a8b61890d612c075f0a82a3e682e1610f7ab52a56bb2122a12200b04920eca11fe080538fd2e165d7e9547b90ee42879e36a267f37dbaa0beee01623bbb58618d1f8122a12208a68f5e8b6a57ef6515d1a9deebb060766edf4dd71f925a37eafa2fadc125ac616235bff420e07ec  | xxd -r -p | protoc --decode_raw
3 {
  1 {
    1: "/klk/prod/2020-01-09"
    2: "\022 \320$\320\311\211\0279G)\301\t\335\370\331L\230\323o\363\211\233\177?m\024\213\216k\257\010M\303\026#k\034f\352\212\216"
```

Here are the messages from an instrumented version of the rust gossipsub:
```
Jul 24 14:26:05.599  WARN libp2p_gossipsub::protocol: message id not an utf8 string: 12202fe480d71af77f741f0a3abf054e30c5d414a8b61890d612c075f0a82a3e682e1610f7ab52a57069    
Jul 24 14:26:05.599  WARN libp2p_gossipsub::protocol: message id not an utf8 string: 1220d024d0c98917394729c109ddf8d94c98d36ff3899b7f3f6d148b8e6baf084dc316236b1c66ea8e82    
[...]

PR #363, opened 2020-07-24
Add warning about messageIDs

rklaehn:
See https://github.com/libp2p/specs/pull/285 and https://github.com/libp2p/go-libp2p-pubsub/issues/361

The problem is that many people implementing a protocol will not look at the specs, but grab the proto definition from the most mature impl, which is go-libp2p. But most languages, including rust, produce protobuf that rejects non-utf8 strings, leading to issues like https://github.com/libp2p/rust-libp2p/issues/1671 .

Ideally, in the long term this should be changed to bytes here as well. Bytes is just the better choice for an opaque id, where you might want to use compact random data and/or hashes.

### DIFF
--- a/pb/rpc.proto
+++ b/pb/rpc.proto
@@ -32,10 +32,12 @@ message ControlMessage {
 
 message ControlIHave {
 	optional string topicID = 1;
+	// implementors from other languages should use bytes here - go protobuf emits invalid utf8 strings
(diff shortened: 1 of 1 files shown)


vyzo (approved):
Yes, we should eventually fix it here as well.

daviddias (approved):

PR #365, opened 2020-07-27
Refactor Gossipsub Parameters To Make Them More Configurable

### LINKED ISSUE #364: Gossipsub Parameters are Not Easily Configurable
Currently all the gossipsub specific parameters are package globals which are initialized at startup.Ex: 
```
	GossipSubD = 6
	GossipSubDlo = 5
	GossipSubDhi = 12
	GossipSubDscore = 4
	GossipSubDout = 2
        .
        .
        .
``` 
However for our usecase we need the ability to modify a few parameters, before creating the gossipsub router.  The current method would be to set the package variable to the value we want and the instantiate the router. However this method seems potentially brittle as it could easily conflict if we instantiate another router with different parameters. Currently there is no way to provide specific gossipsub parameters for one router instance.

We could instead set all the parameters using a `GossipSubParams` config and just call the values from there instead of referring to the global parameters. That way each instance of the gossip-sub router refers to the parameters in the config instead of the globally set parameters.

nisdas:
Resolves #364 

- [x] Add a special gossipsub parameter config to be able to instantiate the router with.
- [x] Add a pubsub option that allows a custom config to be set when creating the pubsub object.
- [x] Fix references so that only the param object in the gossipsub router struct is accessed instead of the
default global parameter.
- [x] Add a unit test to ensure this feature works as expected.

cc @vyzo @raulk


gossipsub.go
```diff
@@ -28,41 +28,72 @@ const (
...
+
+// GossipSubParams defines all the gossipsub specific parameters.
+type GossipSubParams struct {
 	// overlay parameters.
 
 	// GossipSubD sets the optimal degree for a GossipSub topic mesh. For example, if GossipSubD == 6,
 	// each peer will want to have about six peers in their mesh for each topic they're subscribed to.
 	// GossipSubD should be set somewhere between GossipSubDlo and GossipSubDhi.
-	GossipSubD = 6
+	GossipSubD int
```
vyzo: let's drop the GossipSub prefix in these variables, it's redundant.

gossipsub.go
```diff
@@ -176,29 +208,93 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
-		directConnectTicks:      GossipSubDirectConnectTicks,
-		opportunisticGraftTicks: GossipSubOpportunisticGraftTicks,
+		directConnectTicks:      params.GossipSubDirectConnectTicks,
+		opportunisticGraftTicks: params.GossipSubOpportunisticGraftTicks,
 
-		fanoutTTL: GossipSubFanoutTTL,
+		fanoutTTL: params.GossipSubFanoutTTL,
 
 		tagTracer: newTagTracer(h.ConnManager()),
+		paramsCfg: params,
```
vyzo: let's just call this params.

gossipsub.go
```diff
@@ -176,29 +208,93 @@ func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, er
...
-		Dhi:    GossipSubDhi,
-		Dscore: GossipSubDscore,
-		Dout:   GossipSubDout,
-		Dlazy:  GossipSubDlazy,
+		D:      params.GossipSubD,
+		Dlo:    params.GossipSubDlo,
+		Dhi:    params.GossipSubDhi,
+		Dscore: params.GossipSubDscore,
+		Dout:   params.GossipSubDout,
+		Dlazy:  params.GossipSubDlazy,
```
vyzo: let's just use the params struct directly, this is getting messy.

vyzo: Also, let's make the params object a plain struct, not a pointer so that there is no extra indirection.

vyzo (changes requested):
Thank you!

A couple of notes on how to make things better:
1) Naming: drop the GossipSub prefix in the params struct, it's redundant
2) Grouping: let's use the params struct directly instead of copying values right and left, it's confusing
3) Indirection: let's keep the params struct as a struct (not a pointer) to avoid the extra memory indirection.

jacobheun:
@nisdas can you take a look at @vyzo's feedback so we can move this forward?

nisdas:
Hey @jacobheun , yeap I will try to get on this as soon as possible. I am a bit swamped right now but I will address the feedback as soon as I can.

nisdas:
@vyzo @jacobheun 
Sorry for the delay, but this is ready for another look

protolambda:
I am interested in this configuration feature as well. Can we move this forward? Do you need any help?

protolambda:
Is anyone looking at merging this feature? Can I help?

nisdas:
@protolambda I will resync with master and fix conflicts by today

BigLep:
@nisdas: what's the status here?  Will you be able to fix the tests?

nisdas:
@BigLep I didn't get another review for this, so I havent been keeping it updated. However all tests should pass, the previous failure seemed unrelated to the PR(1 test run passed and the other exited abruptly with no test error). Its passing now, after resyncing the branch.

aschmahmann:
Do you mind rebasing on master? CI should be passing now.

nisdas:
Alright done @aschmahmann , I have resolved the previous conflicts

BigLep:
@Stebalien is working on getting CI updates to run.

Stebalien:
This didn't need a rebase. Trying to kick CI... and failing, apparently.

Stebalien:
Let's wait for #415. That should fix CI.

Stebalien:
I'm going to re-open this as a new PR to get CI to run.

Stebalien:
Merged. Thanks all for your patience.

PR #369, opened 2020-07-31
trace topics in Deliver/Reject/Duplicate events

vyzo:
Necessary to filter only specific topics in tracers and simplify processing logic in tools.
Also downgrades the trace buffer overflow message to Debug as it spams lotus.


whyrusleeping (approved):
LGTM

raulk (approved):

PR #370, opened 2020-08-03
Add Topic() method to Topic struct.

### LINKED ISSUE #358: pubsub.Topic should satisfy fmt.Stringer
Due to the requirement of maintaining a single `*pubsub.Topic` handle per pubsub topic, it is often helpful to identify the topic to which a given handle belongs.

Common solutions include:

- `map[string]*topic.Topic`
- Defining a local topic handle that embeds `*topic.Topic`, and exposes the topic string via a `String() string` method

Seeing as how `*topic.Topic` already contains an unexported `topic` field, it would be both trivial and extremely helpful to add the following method:

```go
func (t *Topic) String() string {
    return t.topic
}
```

I'm happy to submit a PR if there's any interest.

lthibault:
Fixes #358 and part of #198.

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -26,6 +26,11 @@ type Topic struct {
 	closed bool
 }
 
+// String returns the topic associated with t
+func (t *Topic) String() string {
+	return t.topic
+}
+
 // EventHandler creates a handle for topic specific events
 // Multiple event handlers may be created and will operate independently of each other
 func (t *Topic) EventHandler(opts ...TopicEventHandlerOpt) (*TopicEventHandler, error) {


aschmahmann:
@vyzo you good with merging this?

vyzo:
We want a `String()` method I think, but I am not going to block it.

lthibault:
@vyzo I initially used `Topic` because it is used by [`Subscription`](https://github.com/libp2p/go-libp2p-pubsub/blob/55f4ad6eb98b9e617e46641e7078944781abb54c/subscription.go#L18), but I actually agree that `String` is better.

Shall I make the change?

vyzo:
yes please!

lthibault:
@vyzo Done!

vyzo (approved):
thank you!

PR #374, opened 2020-08-10
add slack time to prune backoff clearance

### LINKED ISSUE #368: Gossipsub backoff latencies may result in punishment
Consider the following scenario:

1. Peer `A` sends a `PRUNE` to peer `B` with a backoff time of 60 seconds. It stores the 60 second backoff time in its internal data structure.
2. Peer `B` receives the `PRUNE` message after a short delay of lets say one second and stores the 60 second backoff in its internal data structure.
3. Exactly 60 seconds after peer `A` sent the `PRUNE` it might decide to `GRAFT` to peer `B` again and sends a `GRAFT`.
4. This time the message gets received and processed by peer `B` in less than a second and therefore for peer `B` the backoff is still active and it declines the `GRAFT` and punishes peer `A`.

To circumvent such punishments peer `A` should consider some slack time before it sends a `GRAFT` to peer B. So I propose adding some slack times to backoffs after sending a `PRUNE`. Note that this slack time should only be considered for sending `GRAFT`s and not for checking if an incoming `GRAFT` is allowed (in this case we consider the real backoff time without the additional slack).

vyzo:
Fixes #368.

Implements https://github.com/libp2p/specs/pull/289.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -118,6 +118,10 @@ var (
 	// before attempting to re-graft.
 	GossipSubPruneBackoff = time.Minute
 
+	// GossipSubPruneBackoffSlack is the slack time after expiry of a backoff and before
+	// attempting a GRAFT.
+	GossipSubPruneBackoffSlack = time.Second
+
 	// GossipSubConnectors controls the number of active connection attempts for peers obtained through PX.
 	GossipSubConnectors = 8
 
@@ -1497,7 +1501,7 @@ func (gs *GossipSubRouter) clearBackoff() {
 		return
 	}
 
-	now := time.Now()
+	now := time.Now().Add(GossipSubPruneBackoffSlack)
 	for topic, backoff := range gs.backoff {
 		for p, expire := range backoff {
 			if expire.Before(now) {
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -118,10 +118,6 @@ var (
 	// before attempting to re-graft.
 	GossipSubPruneBackoff = time.Minute
 
-	// GossipSubPruneBackoffSlack is the slack time after expiry of a backoff and before
-	// attempting a GRAFT.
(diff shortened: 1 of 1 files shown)


raulk (commented):
Hm, not sure adding a new parameter is the right way to go. The slack we're modelling for is the heartbeat interval, no?

My understanding is that slack we need to model for is loosely `slack = heartbeat(B) - latency(B->A)`, which we could simplify as `heartbeat` only.

I'm OK keeping the separate var if you really want it, although I think it pollutes things. But I'd say let's initialize it to the _effective_ heartbeat value; that is, taking into account any parameter changes the user may have made.

vyzo:
Well no, it's a little more complex (a real edge case).
We already have some slack in that we only clear backoff every 15 ticks or so.
The problem arises if we happen to remove the backoff at exactly the 60s mark, in which case it becomes eligible for graft - but the other side may be still in backoff because of variable latency.
It's very unlikely that this could happen, but it is still a bug.
The patch simply adds some extra slack to model the variable latency so that we don't hit the issue in the edge case.

In terms of mechanics, yeah we can add a heartbeat (or two) instead of having an extra variable; I am fine with that.

vyzo:
Changed it to use 2*heartbeat for slack.

raulk:
BTW -- took the liberty to submit a spec change PR, as this should be part of the agreed-upon state machine: https://github.com/libp2p/specs/pull/289

vyzo:
Yeah, I saw -- good thinking, this is a tricky edge case that needs to be specified.

vyzo:
gah BUG alert: the Add shouldn't go into the now variable, but rather in the backoff itself; fixing.

vyzo:
ok, now it should be good.

raulk (approved):

PR #375, opened 2020-08-10
don't add direct peers to fanout

### LINKED ISSUE #371: Gossipsub direct peers get added into fanout and from there into mesh on subscription
Consider the following scenario:

1. `P1` and `P2` get initiated as direct peers of each other
2. `P1` subscribes to a topic `"topic"`
3. `P2` publishes a message in topic `"topic"`, this will create a fanout, but it gets not checked if the peers are direct peers (see https://github.com/libp2p/go-libp2p-pubsub/blob/2b5243c72f0d1a13a49d346e0f8391bbb0076f8d/gossipsub.go#L905), therefore `P1` will be part of the fanout of `P2`.
4. `P2` subscribes to the topic `"topic"` and and the peers from the fanout will be added to the mesh which will result in a `GRAFT` to peer `P1`.

The spec states:
> It is an error to GRAFT on an explicit peer, [...]

Therefore I assume this is a bug and either direct peers shouldn't get added to the fanout (a direct peer in a fanout doesn't make sense) or they should get filtered out before the fanout gets added to the mesh on subscription.

vyzo:
Fixes #371.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -903,7 +903,8 @@ func (gs *GossipSubRouter) Publish(msg *Message) {
 			if !ok || len(gmap) == 0 {
 				// we don't have any, pick some with score above the publish threshold
 				peers := gs.getPeers(topic, gs.D, func(p peer.ID) bool {
-					return gs.score.Score(p) >= gs.publishThreshold
+					_, direct := gs.direct[p]
+					return !direct && gs.score.Score(p) >= gs.publishThreshold
 				})
 
 				if len(peers) > 0 {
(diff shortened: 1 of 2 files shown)


raulk (commented):
Do we have a test that tests the edge case outlined in #371? If not, we should add one to this PR before merging to catch potential regressions.

vyzo:
we don't have a test, but we can certainly add one.

vyzo:
rebased on master and added test.

raulk (approved):

PR #376, opened 2020-08-12
Add api for dynamically setting topic score parameters

### LINKED ISSUE #346: Feature request for gossipsub v1.1: delayed adding of topic score parameters
Wondering if this is feasible to be implemented. 
At the current time, we have to know from the start the topic list in order to use the peer scoring feature. This will bring more complexity in a system that will have a topic list that changes in time.
Thanks

vyzo:
Closes #346 

Adds a `SetScoreParams` to Topic so that we can dynamically set topic score parameters.

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -1781,6 +1781,57 @@ func TestGossipsubPeerScoreInspect(t *testing.T) {
 	}
 }
 
+func TestGossipsubPeerScoreResetTopicParams(t *testing.T) {
+	// this test exercises the code path sof peer score inspection
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	hosts := getNetHosts(t, ctx, 1)
+
+	ps := getGossipsub(ctx, hosts[0],
+		WithPeerScore(
+			&PeerScoreParams{
+				Topics: map[string]*TopicScoreParams{
+					"test": &TopicScoreParams{
+						TopicWeight:                    1,
+						TimeInMeshQuantum:              time.Second,
+						FirstMessageDeliveriesWeight:   1,
+						FirstMessageDeliveriesDecay:    0.999,
+						FirstMessageDeliveriesCap:      100,
+						InvalidMessageDeliveriesWeight: -1,
+						InvalidMessageDeliveriesDecay:  0.9999,
+					},
+				},
+				AppSpecificScore: func(peer.ID) float64 { return 0 },
+				DecayInterval:    time.Second,
+				DecayToZero:      0.01,
+			},
+			&PeerScoreThresholds{
+				GossipThreshold:   -1,
+				PublishThreshold:  -10,
+				GraylistThreshold: -1000,
+			}))
+
+	topic, err := ps.Join("test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = topic.SetScoreParams(
+		&TopicScoreParams{
+			TopicWeight:                    1,
+			TimeInMeshQuantum:              time.Second,
+			FirstMessageDeliveriesWeight:   1,
+			FirstMessageDeliveriesDecay:    0.999,
+			FirstMessageDeliveriesCap:      200,
+			InvalidMessageDeliveriesWeight: -1,
+			InvalidMessageDeliveriesDecay:  0.9999,
+		})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
 type mockPeerScoreInspector struct {
 	mx     sync.Mutex
 	scores map[peer.ID]float64
--- a/score.go
+++ b/score.go
@@ -182,9 +182,11 @@ func newPeerScore(params *PeerScoreParams) *peerScore {
 	}
 }
 
-// update interface
+// SetTopicScoreParams sets new score parameters for a topic.
+// If the topic previously had parameters and the parameters are lowering delivery caps,
+// then the score counters are recapped appropriately.
+// Note: assumes that the topic score parameters have already been validated
 func (ps *peerScore) SetTopicScoreParams(topic string, p *TopicScoreParams) error {
-	// Note: assumes that the topic score parameters have already been validated
 	ps.Lock()
 	defer ps.Unlock()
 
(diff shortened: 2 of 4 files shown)


topic.go
```diff
@@ -31,6 +31,46 @@ func (t *Topic) String() string {
 	return t.topic
 }
 
+// SetScoreParams sets the topic score parameters if the pubsub router supports peer
+// scoring
+func (t *Topic) SetScoreParams(p *TopicScoreParams) error {
```
raulk, line 36: Can `SetScoreParameters` be called at any time during runtime to reset the topic's score parameters?

* If yes, how does this interact with any existing cached scores for peers in that topic? Should we be resetting them? Or at least letting the user decide through a boolean param?
* If no, we should remove that gotcha by erroring if the user attempts to change the parameters _after_ the first publish or subscribe operation.

vyzo: Well, as it is it can be called at any time; but we can change it to check whether parameters already exist for the score and if that's the case we can error out.
This sounds like more reasonable behaviour.

vyzo: Added check for already existing parameters and error if that's the case.

raulk: ~If they can only be set a topic construction time~ if they can only be set once, why can't we make it a `TopicOpt`?

vyzo: we could, but there is the issue of Join/Leave/Join pattern.

vyzo: ~Also, for the sake of argument, there is currently no such thing as a TopicOpt; we'd have to add it and definite semantics for it.~

topic.go
```diff
@@ -31,6 +31,45 @@ func (t *Topic) String() string {
...
+	}
+
+	err := p.validate()
+	if err != nil {
+		return fmt.Errorf("invalid topic score parameters: %w", err)
+	}
+
+	result := make(chan error, 1)
+	select {
+	case t.p.eval <- func() {
```
raulk: let's not inline this, it's hideously difficult to read.

vyzo:
reworked the logic (and squashed a bit) so that we can reset topic score parameters on demand.

topic.go
```diff
@@ -31,6 +31,45 @@ func (t *Topic) String() string {
...
+	defer t.mux.Unlock()
+
+	if t.closed {
+		return ErrTopicClosed
+	}
+
+	err := p.validate()
+	if err != nil {
+		return fmt.Errorf("invalid topic score parameters: %w", err)
+	}
```
raulk: Could do outside the lock.

topic.go
```diff
@@ -31,6 +31,45 @@ func (t *Topic) String() string {
...
+			result <- fmt.Errorf("pubsub router is not gossipsub")
+			return
+		}
+
+		if gs.score == nil {
+			result <- fmt.Errorf("peer scoring is not enabled in router")
+			return
+		}
+
+		result <- gs.score.SetTopicScoreParams(t.topic, p)
```
raulk: can we please divide this in two lines, assigning to `err`?

score.go
```diff
@@ -182,6 +182,50 @@ func newPeerScore(params *PeerScoreParams) *peerScore {
 	}
 }
 
+// update interface
```
raulk: Is this a TODO?

gossipsub_test.go
```diff
@@ -1781,6 +1781,57 @@ func TestGossipsubPeerScoreInspect(t *testing.T) {
...
+
+	topic, err := ps.Join("test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = topic.SetScoreParams(
+		&TopicScoreParams{
+			TopicWeight:                    1,
+			TimeInMeshQuantum:              time.Second,
```
raulk, line 1823: Is there any rebucketing or timer adjustment we need todo if the resolution changes?

score_test.go
```diff
@@ -842,6 +842,103 @@ func TestScoreRetention(t *testing.T) {
 	}
 }
 
+func TestScoreResetTopicParams(t *testing.T) {
```
raulk: This test only seems to be checking recapping; maybe rename the test appropriately? With the current name, it seems like it's doing a different thing.

gossipsub_test.go
```diff
@@ -1781,6 +1781,57 @@ func TestGossipsubPeerScoreInspect(t *testing.T) {
...
+				PublishThreshold:  -10,
+				GraylistThreshold: -1000,
+			}))
+
+	topic, err := ps.Join("test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = topic.SetScoreParams(
```
raulk, line 1820: Do we not want a test that verifies that a peer that wasn't greylisted is effectively graylisted after stricter parameters are set, or something like that?

gossipsub_test.go
vyzo: no, that's dynamically computed when the score is computed, we only keep track of raw time in the stats.

topic.go
vyzo: ok

topic.go
vyzo: sure.

topic.go
vyzo: sure, why not.

score.go
vyzo: no, it's a very terse description; I'll make it better.

score_test.go
vyzo: ok

gossipsub_test.go
vyzo: That's overkill I think, we have very good coverage for this diff and this ought to work just fine.

vyzo: We can still do it in the score unit test, I think doing it in a full gossipsub test would be quite complicated for the value we derive.

topic.go
vyzo: done

topic.go
vyzo: done

topic.go
vyzo: done

score_test.go
vyzo: done

gossipsub_test.go
vyzo: added a test that explicitly verifies that the score changes as expected, that should cover us.

raulk (approved):

PR #379, opened 2020-08-21
Consider promises fullfilled as soon as validation begins

vyzo:
So that excessively long validation times don't break promises.

### DIFF
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -91,8 +91,7 @@ func (gt *gossipTracer) GetBrokenPromises() map[peer.ID]int {
 
 var _ internalTracer = (*gossipTracer)(nil)
 
-func (gt *gossipTracer) DeliverMessage(msg *Message) {
-	// someone delivered a message, stop tracking promises for it
+func (gt *gossipTracer) fullfillPromise(msg *Message) {
 	mid := gt.msgID(msg.Message)
 
 	gt.Lock()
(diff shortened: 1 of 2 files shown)


daviddias (approved):
The code doesn't read the best as in what was the message flow but checked in with Vyzo to understand the details better. 

Note to self (and others): What was happening that validation was taking too much time, which meant that the promised was not being fulfilled early enough, leading the delay to be considered a penalty.

This was impactful in scenarios were the validator was extensive or when the resources available to the machine were not enough (CPU 100%) causing the validation to take a long time/

gossip_tracer.go
```diff
@@ -113,12 +117,14 @@ func (gt *gossipTracer) RejectMessage(msg *Message, reason string) {
 		return
 	}
 
-	mid := gt.msgID(msg.Message)
-
-	gt.Lock()
-	defer gt.Unlock()
+	gt.fullfillPromise(msg)
```
raulk, line 120: nit: typo -- should be `fulfill`

gossip_tracer.go
```diff
@@ -113,12 +117,14 @@ func (gt *gossipTracer) RejectMessage(msg *Message, reason string) {
...
+}
 
-	delete(gt.promises, mid)
+func (gt *gossipTracer) ValidateMessage(msg *Message) {
+	// we consider the promise fullfilled as soon as the message begins validation
+	// if it was a case of signature issue it would have been rejected immediately
+	// without triggering the Validate trace
+	gt.fullfillPromise(msg)
```
raulk, line 127: Interesting optimisation. I think I've paged out this info, but how do we react if a peer sends us a message we didn't request? I guess there's no way of recognising that situation?

vyzo: that's ok, it won't be in the promise map.

gossip_tracer.go
vyzo: oops! I guess we'll live with it for a little while :)

vyzo: I'll fix it before we cut a release.

gossip_tracer.go
vyzo: and yeah, there is no way of recognizing that situation; it will still be validated and caught by the score tracer if it is invalid, so it's not a problem.

Also, it's more of a bugfix than an optimization :) I have observed validation time spikes of over 4.5s in my lotus node, which in an overnight run resulted in half the peers having negative scores because of behaviour penalties!

PR #384, opened 2020-09-02
Random Early Drop for overloaded validation queue

vyzo:
Introduces a new peer gater for gossipsub, which allows us to do random early drops when we are experiencing elevated validation throttling, because of elevated load or a spam attack targetting the validation queue.

### DIFF
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -14,8 +14,14 @@ import (
 // The tracking of promises is probabilistic to avoid using too much memory.
 type gossipTracer struct {
 	sync.Mutex
-	msgID        MsgIdFunction
-	promises     map[string]map[peer.ID]time.Time
+
+	msgID MsgIdFunction
+
+	// promises for messages by message ID; for each message tracked, we track the promise
+	// expiration time for each peer.
+	promises map[string]map[peer.ID]time.Time
+	// promises for each peer; for each peer, we track the promised message IDs.
+	// this index allows us to quickly void promises when a peer is throttled.
 	peerPromises map[peer.ID]map[string]struct{}
 }
 
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -70,6 +70,7 @@ func (gt *gossipTracer) GetBrokenPromises() map[peer.ID]int {
 	var res map[peer.ID]int
 	now := time.Now()
 
+	// find broken promises from peers
 	for mid, peers := range gt.promises {
 		for p, expire := range peers {
 			if expire.Before(now) {
@@ -91,7 +92,7 @@ func (gt *gossipTracer) GetBrokenPromises() map[peer.ID]int {
 
 var _ internalTracer = (*gossipTracer)(nil)
 
-func (gt *gossipTracer) fullfillPromise(msg *Message) {
+func (gt *gossipTracer) fulfillPromise(msg *Message) {
 	mid := gt.msgID(msg.Message)
 
 	gt.Lock()
@@ -101,12 +102,12 @@ func (gt *gossipTracer) fullfillPromise(msg *Message) {
 }
 
 func (gt *gossipTracer) DeliverMessage(msg *Message) {
-	// someone delivered a message, fullfill promises for it
-	gt.fullfillPromise(msg)
+	// someone delivered a message, fulfill promises for it
+	gt.fulfillPromise(msg)
 }
 
 func (gt *gossipTracer) RejectMessage(msg *Message, reason string) {
-	// A message got rejected, so we can fullfill promises and let the score penalty apply
+	// A message got rejected, so we can fulfill promises and let the score penalty apply
 	// from invalid message delivery.
 	// We do take exception and apply promise penalty regardless in the following cases, where
 	// the peer delivered an obviously invalid message.
@@ -117,14 +118,14 @@ func (gt *gossipTracer) RejectMessage(msg *Message, reason string) {
 		return
 	}
 
-	gt.fullfillPromise(msg)
+	gt.fulfillPromise(msg)
 }
 
 func (gt *gossipTracer) ValidateMessage(msg *Message) {
-	// we consider the promise fullfilled as soon as the message begins validation
+	// we consider the promise fulfilled as soon as the message begins validation
 	// if it was a case of signature issue it would have been rejected immediately
 	// without triggering the Validate trace
-	gt.fullfillPromise(msg)
+	gt.fulfillPromise(msg)
 }
 
 func (gt *gossipTracer) AddPeer(p peer.ID, proto protocol.ID) {}
@@ -134,3 +135,19 @@ func (gt *gossipTracer) Leave(topic string)                   {}
 func (gt *gossipTracer) Graft(p peer.ID, topic string)        {}
 func (gt *gossipTracer) Prune(p peer.ID, topic string)        {}
 func (gt *gossipTracer) DuplicateMessage(msg *Message)        {}
+
+func (gt *gossipTracer) ThrottlePeer(p peer.ID) {
+	gt.Lock()
+	defer gt.Unlock()
+
+	// remove promises for peers that have been throttled so that they are not unfairly penalized
+	for mid, peers := range gt.promises {
+		_, hasPromise := peers[p]
+		if hasPromise {
+			delete(peers, p)
+			if len(peers) == 0 {
+				delete(gt.promises, mid)
+			}
+		}
+	}
+}
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -14,14 +14,16 @@ import (
 // The tracking of promises is probabilistic to avoid using too much memory.
 type gossipTracer struct {
 	sync.Mutex
-	msgID    MsgIdFunction
-	promises map[string]map[peer.ID]time.Time
+	msgID        MsgIdFunction
+	promises     map[string]map[peer.ID]time.Time
+	peerPromises map[peer.ID]map[string]struct{}
 }
 
 func newGossipTracer() *gossipTracer {
 	return &gossipTracer{
-		msgID:    DefaultMsgIdFn,
-		promises: make(map[string]map[peer.ID]time.Time),
+		msgID:        DefaultMsgIdFn,
+		promises:     make(map[string]map[peer.ID]time.Time),
+		peerPromises: make(map[peer.ID]map[string]struct{}),
 	}
 }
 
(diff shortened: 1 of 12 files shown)


peer_gater.go
```diff
@@ -0,0 +1,252 @@
...
+
+type peerGater struct {
+	sync.Mutex
+
+	threshold, decay   float64
+	validate, throttle float64
+
+	lastThrottle time.Time
+
+	stats map[peer.ID]*peerGaterStats
```
Stebalien: This looks like a potential OOM vector, with a 6 hour timeout.

Stebalien: Also, maybe we want to taint the IP as well? It's pretty easy to roll a new peer ID.

Stebalien: For filecoin, we may _also_ want to be able to use an arbitrary "keyer" function (e.g., a function that takes the peer ID and returns a miner ID).

peer_gater.go
```diff
@@ -0,0 +1,252 @@
...
+		select {
+		case <-tick.C:
+			pg.decayStats()
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (pg *peerGater) decayStats() {
```
Stebalien, line 219: uber nit for future reference: pass the time in here from tick to avoid querying the time twice.

peer_gater.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	pg.throttle *= pg.throttle
+	if pg.throttle < DefaultDecayToZero {
+		pg.throttle = 0
+	}
+
+	now := time.Now()
+	for p, st := range pg.stats {
+		if st.connected {
```
Stebalien: Shouldn't we decay even _if_ we're not connected?

peer_gater.go
```diff
@@ -0,0 +1,252 @@
...
+	st, ok := pg.stats[p]
+	if !ok {
+		st = &peerGaterStats{}
+		pg.stats[p] = st
+	}
+	return st
+}
+
+func (pg *peerGater) AcceptFrom(p peer.ID) AcceptStatus {
+	if pg == nil {
```
Stebalien, line 317: :heart:

Stebalien (approved):

peer_gater.go
vyzo: tainting the IP is a really good idea I think; I'll add it.

Re: OOM -- I don't know, the memory usage is very low, just a few bytes.

peer_gater.go
vyzo: I keep them intact so that a poorly performing spammer can't erase his stats by disconnecting.

Stebalien: Hm. I'm a bit concerned that we'll keep disconnecting because they're useless, and we won't give them a chance to get better. But I get why and it makes sense. Otherwise, a spammer would just rotate through peer IDs.

vyzo: that's a fair point, I'll make a small tweak to the probability threshold (basically add 1 to both the numerator and denominator) so that a peer always has a chance and can survive some early hiccups (eg duplicates).

vyzo: see https://github.com/libp2p/go-libp2p-pubsub/pull/384/commits/d7a606bcdca7d9883a4fcf9d8db34c6b8481e23d

peer_gater.go
vyzo: so if we taint the IP and have a common stats object for all peers in the same IP, our memory usage per peer ID becomes just the peer ID itself and a pointer; attacker would have to generate  some significant fraction of a billion of peer IDs in the 6 hour interval to make a dent in memory usage.

Stebalien: OOM vector discussion: You'd need ~100M peer IDs to take ~4GiB of memory. That's over 4k new connections a second.

vyzo: I implemented per IP stat tracking in https://github.com/libp2p/go-libp2p-pubsub/pull/384/commits/cb4aae17443b622231f70a0c54ec3a6712fbce26.

gossip_tracer.go
```diff
@@ -14,14 +14,16 @@ import (
 // The tracking of promises is probabilistic to avoid using too much memory.
 type gossipTracer struct {
 	sync.Mutex
-	msgID    MsgIdFunction
-	promises map[string]map[peer.ID]time.Time
+	msgID        MsgIdFunction
+	promises     map[string]map[peer.ID]time.Time
+	peerPromises map[peer.ID]map[string]struct{}
```
raulk: Can we introduce type aliases here to give some meaning to these multi-layer maps? Alternatively, please document this data structure. Especially because one is called _peer_ promises, and the other isn't, but they both deal with peers...

If one is an index over the other, consider calling this `promiseIdx`.

raulk: I guess `type MessageID = string` would be enough to clear things out a bit...

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	manet "github.com/multiformats/go-multiaddr-net"
+)
+
+var PeerGaterRetainStats = 6 * time.Hour
+var PeerGaterQuiet = time.Minute
+
+type peerGater struct {
```
raulk, line 119: godocs please, even if it's private. Please explain how the logic here works.

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	manet "github.com/multiformats/go-multiaddr-net"
+)
+
+var PeerGaterRetainStats = 6 * time.Hour
+var PeerGaterQuiet = time.Minute
```
raulk: Godocs please; these are public.

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+
+	deliver, duplicate, ignore, reject float64
+}
+
+// WithPeerGater is a gossipsub router option that enables reactive validation queue
+// management.
+// The threshold parameter is the threshold of throttled/validated messages before the gating
+// kicks in.
+// The decay parameter is the (linear) decay of counters per second.
+func WithPeerGater(threshold, decay float64) Option {
```
raulk: Should we also pass in the decay interval?

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+	if !ok {
+		st = pg.getIPStats(p)
+		pg.peerStats[p] = st
+	}
+	return st
+}
+
+func (pg *peerGater) getIPStats(p peer.ID) *peerGaterStats {
+	ip := pg.getPeerIP(p)
+	st, ok := pg.ipStats[ip]
```
raulk, line 272: What would happen here if we have more than one peer group at many different IP addresses, all of which failed to parse and fell back to `<unknown>`? Wouldn't their stats intermingle?

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+
+	if time.Since(pg.lastThrottle) > PeerGaterQuiet {
+		return AcceptAll
+	}
+
+	if pg.throttle == 0 {
+		return AcceptAll
+	}
+
+	if pg.validate != 0 && pg.throttle/pg.validate < pg.threshold {
```
raulk: All this logic needs comments.

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+		return AcceptAll
+	}
+
+	st := pg.getPeerStats(p)
+
+	total := st.deliver + 0.25*st.duplicate + st.ignore + 4*st.reject
```
raulk: Same, comments please. These factors seem rather ad-hoc. We should document them and consider making them adjustable.

peer_gater_test.go
```diff
@@ -0,0 +1,122 @@
...
+	accepted := false
+	for i := 0; !accepted && i < 1000; i++ {
+		status = pg.AcceptFrom(peerA)
+		if status == AcceptControl {
+			accepted = true
+		}
+	}
+	if !accepted {
+		t.Fatal("expected AcceptControl")
+	}
```
raulk, line 75: Can't we simplify this and Fatal out if the condition doesn't match inside the loop?

pubsub.go
```diff
@@ -183,6 +183,17 @@ type PubSubRouter interface {
 	Leave(topic string)
 }
 
+type AcceptStatus int
+
+const (
+	// AcceptAll signals to accept the incoming RPC for full processing
+	AcceptNone AcceptStatus = iota
+	// AcceptControl signals to accept the incoming RPC only for control message processing
```
raulk: `control message processing` can mean a lot of things. We should explain this in godocs. Either here, or elsewhere. If elsewhere, link to it.

peer_gater.go
```diff
@@ -0,0 +1,302 @@
...
+	if pg == nil {
+		return AcceptAll
+	}
+
+	pg.Lock()
+	defer pg.Unlock()
+
+	if time.Since(pg.lastThrottle) > PeerGaterQuiet {
+		return AcceptAll
+	}
```
raulk: So this is a circuit breaker construction, right? At this point you close the circuit again, and if the peer misbehaves, they will be throttled again?

vyzo: right, if there hasn't been any queue throttling for 1min, we open it up again.

vyzo: if spam reappears and the queue starts throttling again, the circuit breaker will turn back on.

raulk (commented):
This is a great addition.

It needs a lot more godocs + a draft spec before we merge to master. We're introducing an entirely new mechanism here that works in synergy with the peer scoring, but it's not linked with it. It's not an incremental improvement, but a whole new feature, hence me being insistent about the level of docs/specs we require here.

The way I see it, this PR introduces a defensive frontline circuit breaker that reacts rapidly, while the scoring mechanism is a coalescing, background mechanism.

peer_gater_test.go
vyzo: it's probabilistic, we can't do it deterministically.

peer_gater.go
vyzo: yeah, I've just been tweaking them, but they should be configurable.

peer_gater.go
vyzo: ok.

peer_gater.go
vyzo: this is very difficult to happen, but if it does there is some pathology and we are probably ok with binning them altogether in `<unknown>`.

peer_gater.go
vyzo: sure, why not.

gossip_tracer.go
vyzo: I will add some documentation, the new map was added to have efficient handling of throttling events.

peer_gater.go
vyzo: Added a Warn so that we see it if it actually happens. Maybe some exotic transport will trigger that.

raulk (approved):
Accompanying spec extension is on https://github.com/libp2p/specs/pull/292.

PR #385, opened 2020-09-02
feat: update to go-libp2p-core 0.7.0

Stebalien:
Updates to the new stream interfaces. See https://github.com/libp2p/go-libp2p-core/releases/tag/v0.7.0.


protolambda:
Interested in this PR, currently pubsub is broken with latest libp2p core version, due to the FullClose missing helper function. Can I help?

Stebalien:
Please downgrade go-libp2p-core (i.e., use the version referenced by the latest go-libp2p). This patch set got stuck on some unexpected issues but should unstick within a few weeks.

protolambda:
You mean this PR? The CI just says that there are some unused imports. The helpers package is not used anymore. Or is there something else to wait for?

Stebalien:
This is waiting on https://github.com/libp2p/go-libp2p/pull/1001, which is blocked on some very tricky issues in protocol negotiation.

go.mod
```diff
@@ -8,12 +8,12 @@ require (
...
+	github.com/libp2p/go-libp2p-core v0.7.0
 	github.com/libp2p/go-libp2p-discovery v0.5.0
-	github.com/libp2p/go-libp2p-swarm v0.2.8
+	github.com/libp2p/go-libp2p-swarm v0.3.1
 	github.com/libp2p/go-msgio v0.0.6
-	github.com/multiformats/go-multiaddr v0.2.2
-	github.com/multiformats/go-multiaddr-net v0.1.5
-	github.com/multiformats/go-multistream v0.1.1
+	github.com/multiformats/go-multiaddr v0.3.1
+	github.com/multiformats/go-multiaddr-net v0.2.0
```
aschmahmann: This shouldn't be needed anymore, so there's probably a straggler

aschmahmann (approved):
Looks good although @vyzo may want to take a look.

Also, should `handleNewStream` call Close instead of Reset now or is leaving it as Reset the intended behavior?

Looks like there might be a test failure, but rerunning CI check.

Stebalien:
> Also, should handleNewStream call Close instead of Reset now or is leaving it as Reset the intended behavior?

It doesn't really matter in that case. That'll only happen when we're shutting down (who cares) or we've already disconnected from the peer (again, who cares).  We only call "close" in the case where the peer sends os a nice EOF for some reason (which will never happen in the go implementation, actually...).

vyzo:
Yeah, looks fine -- @stebalien will prod when it's time to merge (go-libp2p and every thing else must be ready first).

Stebalien:
@vyzo all done. Merge when ready.

vyzo (approved):

PR #387, opened 2020-09-12
make IPv6 subnet colocation optional as a peer score parameter

vyzo:


daviddias:
What's the rationale? was this being abused?

vyzo:
no, but it can create problems with providers like textile.

vyzo:
basically I want to make it an option, it's something we discussed with the eth2 folks as well.

raulk (commented):
Mind editing the PR body and explaining the observations from Eth2 and Filecoin that justify making this change? The patch is easy to understand, but the desired impact isn't, and that makes it obscure for the community to follow the direction. Thanks!

ribasushi:
cc @jleni this might be of interest too

PR #388, opened 2020-10-08
Remove support for multi-topic messages

vyzo:
This was a fringe / half-baked feature in the wire protocol that has no direct users or API exposure.

Not only we didn't provide an API to do multi-topic publishing, but also the recent introduction of the topic handle abstraction created impedance mismatch.

So this feature adds bloat with no visible gain, and this PR proposes simplification by removal.

## Migration path

There is no visible consequence for go-libp2p-pubsub users. At the wire level, due to the way that Protobuf serialises repeated fields, this causes no breakage. If direct wire protocol users (e.g. other libp2p implementations) were using the variadic topics field, this patch will make Protobuf pick the last topic in the list.


compat/compat.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+
+package compat.pb;
```
raulk, line 3: Nice! This is a great technique we should keep using across the board on protobuf changes ;-)

compat_test.go
```diff
@@ -0,0 +1,83 @@
...
+	if newMessage.GetTopic() != topic1 {
+		t.Fatalf("bad topic: expected %s, got %s", topic1, newMessage.GetTopic())
+	}
+
+	newMessage.Reset()
+	err = newMessage.Unmarshal(oldMessage2b)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if newMessage.GetTopic() != topic2 {
```
raulk, line 68: Cool, this verifies that indeed the last topic in the list survives.

gossipsub_spam_test.go
```diff
@@ -662,10 +662,10 @@ func TestGossipsubAttackInvalidMessageSpam(t *testing.T) {
 					// fail validation and reduce the attacker's score)
 					for i := 0; i < 100; i++ {
 						msg := &pb.Message{
-							Data:     []byte("some data" + strconv.Itoa(i)),
-							TopicIDs: []string{mytopic},
-							From:     []byte(attacker.ID()),
-							Seqno:    []byte{byte(i + 1)},
+							Data:  []byte("some data" + strconv.Itoa(i)),
+							Topic: &mytopic,
```
raulk, line 666: I think we should remove the `optional` trait from the protobuf file for the topic. It is compulsory (AFAIK) and it would remove this annoying indirection, and speed things up slightly.

pb/rpc.proto
```diff
@@ -18,7 +18,7 @@ message Message {
 	optional bytes from = 1;
 	optional bytes data = 2;
 	optional bytes seqno = 3;
-	repeated string topicIDs = 4;
+	optional string topic = 4;
```
raulk, line 21: Don't need to do here, but we should drop the optional here.

score.go
```diff
@@ -884,69 +883,61 @@ func (ps *peerScore) markFirstMessageDelivery(p peer.ID, msg *Message) {
...
-		// check against the mesh delivery window -- if the validated time is passed as 0, then
-		// the message was received before we finished validation and thus falls within the mesh
-		// delivery window.
-		if !validated.IsZero() && now.After(validated.Add(tparams.MeshMessageDeliveriesWindow)) {
-			continue
-		}
+	// check against the mesh delivery window -- if the validated time is passed as 0, then
+	// the message was received before we finished validation and thus falls within the mesh
+	// delivery window.
+	if !validated.IsZero() && time.Since(validated) > tparams.MeshMessageDeliveriesWindow {
```
raulk, line 933: This is much readable, thanks!

validation_test.go
```diff
@@ -300,471 +293,3 @@ func TestValidateAssortedOptions(t *testing.T) {
 		}
 	}
 }
-
-func TestValidateMultitopic(t *testing.T) {
```
raulk, line 304: 🗑️🚀

raulk (approved):
Nice! Glad to see this go. This feature complicated our code substantially, and it wasn't even exposed via the API.

In the future, we could provide some porcelain function to fan out a message on many topics.

gossipsub_spam_test.go
vyzo: Agreed. But it's a bit of work, so let's do it in a future pr.

PR #389, opened 2020-10-08
Topic Subscription filters

vyzo:
# Topic Subscription Filters

This adds support for subscription filters, which allows the user to explicitly specify topics of interest to the pubsub system.
Only topics that satisfy the filter will be tracked for other peers, which reduces processing load and memory requirements in the system.
Similarly, the user will be able to join only topics allowed by the filter.

## Usage

The filter is enabled by passing the `WithSubscriptionFilter` option to the pubsub constructor.
A couple of basic filters are provided as part of the library; the user may implement more elaborate filtering logic by implementing the `SubscriptionFilter` interface

### DIFF
--- a/subscription_filter.go
+++ b/subscription_filter.go
@@ -1,0 +1,149 @@
+package pubsub
+
+import (
+	"errors"
+	"regexp"
+
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// ErrTooManySubscriptions may be returned by a SubscriptionFilter to signal that there are too many
+// subscriptions to process.
+var ErrTooManySubscriptions = errors.New("too many subscriptions")
+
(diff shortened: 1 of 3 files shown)


subscription_filter.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, sub := range accept {
+		result = append(result, sub)
+	}
+
+	return result
+}
+
+// WrapLimitSubscriptionFilter wraps a subscription filter with a hard limit in the number of
+// subscriptions allowed in an RPC message.
+func WrapLimitSubscriptionFilter(filter SubscriptionFilter, limit int) SubscriptionFilter {
```
raulk, line 128: Nice, I like that the naming is idiomatic.

raulk (commented):
LGTM. We can add more useful subscription filters to the toolbox going forward.

raulk (approved):

PR #393, opened 2020-12-06
Add preconnected nodes to pubsub

aschmahmann:
PubSub only tries to add peers to the set of pubsub nodes when they connect. This meant that peers that were already connected to the host (e.g. tests, multiple nodes on the same machine, an application waiting a while to start up pubsub, etc.) would not get treated as pubsub nodes.

This resolves that by simply adding the list of already connected peers to pubsub.

Note: PubSub still does not utilize `Identify` events to figure out if a peer supports one of the valid pubsub protocols or not, that other implementations of PubSub need to run a similar script instead of just toggling on and off support for a protocol like FloodSub.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -292,6 +292,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 
 	go ps.processLoop(ctx)
 
+	(*PubSubNotif)(ps).Initialize()
+
 	return ps, nil
 }
 
--- a/floodsub_test.go
+++ b/floodsub_test.go
@@ -1137,3 +1137,66 @@ func TestWithInvalidMessageAuthor(t *testing.T) {
 		t.Fatal("expected error")
 	}
 }
+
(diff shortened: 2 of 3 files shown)


aschmahmann:
Note: CI on this branch seems to be failing sometimes - I think the `TestTagTracerDeliveryTags` test is flaky and that it's unrelated to this PR. lmk what you think @vyzo

vyzo:
yeah, that test is flaky and it's getting on my nerves.

pubsub.go
```diff
@@ -292,6 +292,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 
 	go ps.processLoop(ctx)
 
+	(*PubSubNotif)(ps).Initialize()
```
vyzo, line 295: should we do that in a goroutine perhaps?

vyzo: nah, it's fine.

vyzo (approved):

PR #396, opened 2021-01-12
inbound stream deduplication

vyzo:
This adds deduplication for inbound pubsub streams so that we always have only a single inbound stream from a peer; last one wins.

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -52,6 +52,14 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 	p.inboundStreams[peer] = s
 	p.inboundStreamsMx.Unlock()
 
+	defer func() {
+		p.inboundStreamsMx.Lock()
+		if p.inboundStreams[peer] == s {
+			delete(p.inboundStreams, peer)
+		}
+		p.inboundStreamsMx.Unlock()
+	}()
+
 	r := protoio.NewDelimitedReader(s, p.maxMessageSize)
 	for {
 		rpc := new(RPC)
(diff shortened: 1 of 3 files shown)


comm.go
```diff
@@ -41,6 +41,17 @@ func (p *PubSub) getHelloPacket() *RPC {
 }
 
 func (p *PubSub) handleNewStream(s network.Stream) {
+	peer := s.Conn().RemotePeer()
+
+	p.inboundStreamsMx.Lock()
+	other, dup := p.inboundStreams[peer]
+	if dup {
+		log.Debugf("duplicate inbound stream from %s; resetting other stream", peer)
+		other.Reset()
```
raulk, line 50: Does the muxer guarantee that an outstanding read from this stream will error out immediately as soon as we reset it? Can't remember.

comm.go
```diff
@@ -54,10 +65,17 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 				// but it doesn't hurt to send it.
 				s.Close()
 			}
+
+			p.inboundStreamsMx.Lock()
+			if p.inboundStreams[peer] == s {
+				delete(p.inboundStreams, peer)
+			}
+			p.inboundStreamsMx.Unlock()
```
raulk: Probably should put this in a `defer`, so that it's also run on other return paths.

comm.go
vyzo: I don't really know, but it shouldn't matter much. As long as it happens promptly enough we should be fine.

comm.go
vyzo: Well, there are no other relevant return code paths. The only other return is in the context cancellation case, where we don't care because we are shutting down.
Having said that, it is probably cleaner to do it in a defer so why not.

raulk (approved):

PR #397, opened 2021-02-09
Subnet whitelisting for IPColocation

vyzo:
This changes the IP whitelist to support subnets when considering the IPColocation factor.

Note: this is a (very minor) breaking change, but I don't see the reason of keeping the ip whitelist when we have subnets.

cc @travisperson

### DIFF
--- a/score.go
+++ b/score.go
@@ -336,10 +336,20 @@ func (ps *peerScore) ipColocationFactor(p peer.ID) float64 {
 	}
 
 	var result float64
+loop:
 	for _, ip := range pstats.ips {
 		_, whitelisted := ps.params.IPColocationFactorWhitelist[ip]
 		if whitelisted {
 			continue
+		}
+
+		if len(ps.params.IPColocationFactorWhitelistSubnets) > 0 {
+			ipObj := net.ParseIP(ip)
+			for _, ipNet := range ps.params.IPColocationFactorWhitelistSubnets {
+				if ipNet.Contains(ipObj) {
+					continue loop
+				}
+			}
 		}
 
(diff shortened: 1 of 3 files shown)


score.go
```diff
@@ -336,10 +336,15 @@ func (ps *peerScore) ipColocationFactor(p peer.ID) float64 {
...
-		_, whitelisted := ps.params.IPColocationFactorWhitelist[ip]
-		if whitelisted {
-			continue
+		if len(ps.params.IPColocationFactorWhitelist) > 0 {
+			ipObj := net.ParseIP(ip)
+			for _, ipNet := range ps.params.IPColocationFactorWhitelist {
+				if ipNet.Contains(ipObj) {
+					continue loop
+				}
+			}
```
raulk: Let's avoid this repeated computation. The worst case complexity for this parsing + subnet inclusion check is `O(N*M*P)` (N=number of IPs; M=number of whitelisted subnets; P=number of peers). Ideas:

- store the peer IPs as `[]net.IP` instead of `[]string` (prevents reparsing); this is should be OK as long as we don't use the IP as a map key.
- memoize if the IP is whitelisted or not in state; of course, we'd lose the ability to modify `IPColocationFactorWhitelist` at runtime, but that should never happen as it's not safe, IUC.

raulk (changes requested):

score.go
vyzo: Yeah, I was thinking about it too; will do something more reasonable.

vyzo: done, added memoization cache.

raulk (approved):

score.go
raulk: 👍

PR #403, opened 2021-03-10
Expose internal tracing with the RawTracer interface

### LINKED ISSUE #399: Tracing is uneasy with only the public interface
I wanted to monitor the messages dropped due to #398 but the tracing API make that difficult. I ended up writing the following (obviously not great) code:

```go
type tracer struct {
	ingressDropped *stats.Int64Measure
	egressDropped  *stats.Int64Measure
}

func (t tracer) Trace(evt *pubsub_pb.TraceEvent) {
	switch evt.GetType() {
	case pubsub_pb.TraceEvent_REJECT_MESSAGE:
		if evt.RejectMessage == nil {
			return
		}
		switch evt.RejectMessage.GetReason() {
		case "validation queue full", "validation throttled":
			stats.Record(context.Background(), t.ingressDropped.M(1))
		}
	case pubsub_pb.TraceEvent_DROP_RPC:
		stats.Record(context.Background(), t.egressDropped.M(1))
	}
}
```

This code:
- use string constants copied from the package as they are not public
- require a deep understanding of the package internals to catch the correct case
- imply that we are doing an unnecessary marshaling to a protobuf message

The internalTracer interface looks better in my opinion but it's also not publicly accessible.

vyzo:
Fixes #399

### DIFF
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -111,7 +111,7 @@ func (gt *gossipTracer) GetBrokenPromises() map[peer.ID]int {
 	return res
 }
 
-var _ internalTracer = (*gossipTracer)(nil)
+var _ RawTracer = (*gossipTracer)(nil)
 
 func (gt *gossipTracer) fulfillPromise(msg *Message) {
 	mid := gt.msgID(msg.Message)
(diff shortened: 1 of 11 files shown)


aschmahmann:
@marten-seemann if you've got time to review this that would be great, otherwise lmk.

marten-seemann:
@aschmahmann I don’t have any context on this, nor any experience with the tracing used here.

aschmahmann (approved):
LGTM, thanks 😄.

I ran into some issues testing on Windows, but they seem unrelated to this pretty straightforward PR so I'll just flag them separately.

PR #404, opened 2021-03-16
ci: add windows ci

aschmahmann:
I've noticed some issues when running tests on Windows, we should really track this in CI.

@vyzo how do you feel about moving away from Travis? @marten-seemann has been working on some CI standardization using GitHub actions in https://github.com/ipld/.github/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 os:
   - linux
+  - windows
 
 language: go
 


vyzo:
Sure, we are not married to travis.

marten-seemann:
We can include this repo in the new GitHub Actions workflow, as soon as we get it deployed PL-wide.

vyzo:
Sure let's do that.

@aschmahmann the windows ci is timing out; this is probably travis sucking, but maybe there is a real issue underneath.

aschmahmann:
@vyzo I think this is a real thing, not just Travis. This sometimes happens on my laptop too.

There are also a couple test failures related to writing to /tmp instead of Go's OS agnostic temporary directory.

vyzo:
care to fix those tests in the pr?

aschmahmann:
Yep, I can definitely do the `/tmp` fixes here. The hanging might take me some time before I can look into it though

vyzo:
we have dropped travis, as it is horribly broken; closing.

PR #405, opened 2021-03-18
fix: avoid validating messages from ourself before sending

### LINKED ISSUE #398: Published Messages Are Dropped In Gossipsub
Currently before a message can be broadcasted to all peers in a mesh, it has to pass a series of steps before it can be sent to its peers. 

- Host Publishes Message:
https://github.com/libp2p/go-libp2p-pubsub/blob/master/topic.go#L245
- Message is handled in the event queue:
https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L591

- While message is being processed, it is run by its validators: 
https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L1047

- However if the validation queue is full this message is dropped.
https://github.com/libp2p/go-libp2p-pubsub/blob/master/validation.go#L201

This can lead to messages going missing in the above case. While it does make sense to drop inbound messages in the event our validation queue is full, dropping messages originating from the host (rather than re-propagated) does not seem correct. As given the host is the source and originator of the particular message it should be special cased. 

The issue is similar to #217, something like `WaitUntilQueued` would be very useful here to specify the importance of the message.

Stebalien:
The potential downside is that we could get blocked for sending invalid messages. But the upside is that outbound messages:

* Don't incur a validation cost.
* Can't get dropped because our validation queue is overloaded.

fixes #398

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -1045,7 +1045,7 @@ func (p *PubSub) pushMsg(msg *Message) {
 		return
 	}
 
-	if !p.val.Push(src, msg) {
+	if src != self && !p.val.Push(src, msg) {
 		return
 	}
 


vyzo:
Discussed this synchronously with @stebalien and we decided that the right solution to this problem is to have a separate validation path for own messages that validates synchronously and returns an error to the user.

Closing.

PR #406, opened 2021-03-19
implement synchronous validation for locally published messages

### LINKED ISSUE #398: Published Messages Are Dropped In Gossipsub

vyzo:
So that locally published messages are never dropped because of full validation queues.

Closes #398

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -991,7 +991,7 @@ func (p *PubSub) pushMsg(msg *Message) {
 		return
 	}
 
-	err := p.checkSignature(msg)
+	err := p.checkSigningPolicy(msg)
 	if err != nil {
 		log.Debugf("dropping message from %s: %s", src, err)
 		return
@@ -1021,7 +1021,7 @@ func (p *PubSub) pushMsg(msg *Message) {
 	}
 }
 
-func (p *PubSub) checkSignature(msg *Message) error {
+func (p *PubSub) checkSigningPolicy(msg *Message) error {
 	// reject unsigned messages when strict before we even process the id
 	if p.signPolicy.mustVerify() {
 		if p.signPolicy.mustSign() {
--- a/pubsub.go
+++ b/pubsub.go
@@ -65,9 +65,6 @@ type PubSub struct {
 	// incoming messages from other peers
 	incoming chan *RPC
 
-	// messages we are publishing out to our peers
-	publish chan *Message
-
 	// addSub is a control channel for us to add and remove subscriptions
 	addSub chan *addSubReq
 
@@ -234,7 +231,6 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		signKey:               nil,
 		signPolicy:            StrictSign,
 		incoming:              make(chan *RPC, 32),
-		publish:               make(chan *Message),
 		newPeers:              make(chan peer.ID),
 		newPeerStream:         make(chan network.Stream),
 		newPeerError:          make(chan peer.ID),
@@ -588,10 +584,6 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			preq.resp <- peers
 		case rpc := <-p.incoming:
 			p.handleIncomingRPC(rpc)
-
-		case msg := <-p.publish:
-			p.tracer.PublishMessage(msg)
-			p.pushMsg(msg)
 
 		case msg := <-p.sendMsg:
 			p.publishMessage(msg)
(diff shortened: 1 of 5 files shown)


vyzo:
I also temporarily disabled the flaky delivery tag tracer test -- it's getting on my nerves. Follow up issue:  #407

pubsub.go
```diff
@@ -999,59 +991,66 @@ func (p *PubSub) pushMsg(msg *Message) {
...
+		p.tracer.RejectMessage(msg, RejectSelfOrigin)
+		return
+	}
+
+	// have we already seen and validated this message?
+	id := p.msgID(msg.Message)
+	if p.seenMessage(id) {
+		p.tracer.DuplicateMessage(msg)
+		return
+	}
```
Stebalien, line 1013: We should check if we've seen the message before we check the signature.

validation.go
```diff
@@ -191,6 +210,20 @@ func (v *validation) RemoveValidator(req *rmValReq) {
 	}
 }
 
+// Publish synchronously accepts a locally published message, performs applicable
+// validations and pushes the message for propagate by the pubsub system
+func (v *validation) Publish(msg *Message) error {
```
Stebalien: This is terrible.

pubsub.go
vyzo: the default path does it this way so that an attacker can't squat the (default) message ID of a target by guessing its seqno.

validation.go
vyzo: in terms of name or documentation? or the fact that it is exported? ie, what should change?

Stebalien: The fact that the validator module has a publish method. But I guess that abstraction was broken long ago so it's not like this is anything new.

pubsub.go
Stebalien: We can:

1. Check if we've already seen the message. It doesn't really matter if the signature is valid or not here. At worst, we reject a duplicate message with an invalid signature.
2. Validate the signature.
3. Mark that we've seen the message (aborting if some parallel process also saw the message).

As far as I can tell, that's what we did before as well.

vyzo: sure, we can add an early read only check before checking the signature.

validation.go
vyzo: I think we can just call it PushLocal, it's less yaky this way.

Stebalien: Ah, yes. That makes me feel _much_ better. I still kind of wish we didn't have services calling back into eachother this way... but such is life.

pubsub.go
vyzo: so here the signature check is consistency wrt to our signing policy; it doesn't actually verify the signature just yet. I will rename this method for clarity and also add an early read-only check before we verify the signature, since we can skip it for duplicates.

vyzo: ah wait, we already have the read only check; I will simply rename the `checktSignature` thing to `checkSigningPolicy`.

vyzo: done, renamed `checkSignature` to `checkSigningPolicy` to avoid confusion.

Here is a summary of how it works:
- when an incoming message is received (or published), we first check the signing policy for consistent presence of signatures; if it is not consistent, then it is rejected (we don't actually verify the signature in this step).
- we then perform a read only check in the seenMessages cache using the message id. If we have already marked the message, it is dropped as duplicate
- we validate the signature next. If signature verification fails, we reject the message -- not that we don't mark the message as seen in this step, as this would open an avenue for attack with the default signing policy.
- once signature validation passes, we mark the message as seen; if it was already marked (potentially because there is concurrent validation in progress), we drop it as duplicate.
- we then apply user validators, which happens exactly once for a given message.

validation.go
vyzo: done, I also feel much better about this name.

validation.go
```diff
@@ -15,6 +16,16 @@ const (
 	defaultValidateThrottle    = 8192
 )
 
+// ValidationError is an error that may be signalled from message publication when the message
+// fails validation
+type ValidationError struct {
+	Reason string
```
Stebalien, line 22: Nit: the nice way to do this is to make this "Reason" an error as well and expose it via `Cause()`. That way, one can check `xerrors.Is(error, ValidationError{})` and `xerrors.Is(error, UltimateCause{})`.

Stebalien (approved):

validation.go
vyzo: In principle this is indeed quite nice, but it's quite likely unnecessary -- the Reason strings are exported and part of the public API for tracers.

We _could_ add named errors for all of them but this seems more work than what is worth right now.

If you feel strongly about it, let's make an issue to address it.

Issue #408, opened 2021-03-23
Question: Extending Gossipsub protocol list


richard-ramos:
Is it possible to extend the list of protocols returned in:
https://github.com/libp2p/go-libp2p-pubsub/blob/5457a2845b06d859df6d3683867808e46a15c500/gossipsub.go#L401-L403
I'd like to include an additional ID besides `GossipSubID_v11, GossipSubID_v10, FloodSubID`, yet I find it difficult to extend GossipSubRouter to include my custom ID. 

Floodsub supports this feature with `NewFoodsubWithProtocols` 
https://github.com/libp2p/go-libp2p-pubsub/blob/5457a2845b06d859df6d3683867808e46a15c500/floodsub.go#L17-L22
Is adding a similar functionality something viable? I could create a PR that adds a `NewGossipsubWithProtocols(ctx context.Context, h host.Host, ps []protocol.ID, opts ...Option) (*PubSub, error)`

vyzo:
In principle we could, but there is some impedance mismatch in that the router uses the protocol ID to enable certain features.
For example, it checks for the floodsub ID to determine whether to just treat those peers as simple floodsub peers.
It also checks for the GSv1.1 ID to check whether to enable PX.

richard-ramos:
Here's the code for my previous idea:
https://github.com/libp2p/go-libp2p-pubsub/compare/master...richard-ramos:master

> There is some impedance mismatch in that the router uses the protocol ID to enable certain features.

Perhaps this coud be solved by always return `GossipSubID_v11, GossipSubID_v10, FloodSubID`, yet still allowing to add custom protocols? Something like this code:

```go
func (gs *GossipSubRouter) Protocols() []protocol.ID {
	ps := []protocol.ID{GossipSubID_v11, GossipSubID_v10, FloodSubID}
	return append(ps, gs.protocols...)
}
```

If having this logic is not desirable, would you mind sharing some tips on how could I achieve something similar with the current implementation of GossipSub?

vyzo:
Sure, but how should the router behave towards the custom protocol? Is it
gsv1.1 compatible?

Also, its not entirely clear to me why you want a custom protocol -- what
do you gain by it?

richard-ramos:
I'm attempting to implement this spec: https://specs.vac.dev/specs/waku/v2/waku-relay
The TLDR is:
- It's a thin layer over GossipSub
[...]

PR #409, opened 2021-03-23
Create GossipsubRouter with custom protocols

richard-ramos:
Allows setting up a custom protocol list in GossipSub similar to FloodSub.
Also changes the way a peer's protocol support is verified by introducing `peerHasAny` function which returns true if a peer supports any of the protocols passed to the function.

Related discussion here:
https://github.com/libp2p/go-libp2p-pubsub/issues/408

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -163,21 +163,22 @@ var (
 	GossipSubIWantFollowupTime = 3 * time.Second
 )
 
-// NewGossipSub returns a new PubSub object using GossipSubRouter as the router.
-func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
+// NewGossipSubWithProtocols returns a new PubSub object using GossipSubRouter as the router and the protocols specified in ps.
+func NewGossipSubWithProtocols(ctx context.Context, h host.Host, ps []protocol.ID, opts ...Option) (*PubSub, error) {
 	rt := &GossipSubRouter{
-		peers:    make(map[peer.ID]protocol.ID),
-		mesh:     make(map[string]map[peer.ID]struct{}),
-		fanout:   make(map[string]map[peer.ID]struct{}),
-		lastpub:  make(map[string]int64),
-		gossip:   make(map[peer.ID][]*pb.ControlIHave),
-		control:  make(map[peer.ID]*pb.ControlMessage),
-		backoff:  make(map[string]map[peer.ID]time.Time),
-		peerhave: make(map[peer.ID]int),
-		iasked:   make(map[peer.ID]int),
-		outbound: make(map[peer.ID]bool),
-		connect:  make(chan connectInfo, GossipSubMaxPendingConnections),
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -1732,7 +1749,7 @@ func (gs *GossipSubRouter) piggybackControl(p peer.ID, out *RPC, ctl *pb.Control
 }
 
 func (gs *GossipSubRouter) makePrune(p peer.ID, topic string, doPX bool) *pb.ControlPrune {
-	if gs.peers[p] == GossipSubID_v10 {
+	if gs.peerHasAny(p, GossipSubID_v10) {
```
vyzo, line 1752: I think the right approach here is to check for the PX feature, and passing the protocol.
So `gs.SupportsFeature(GossipSubFeaturePX, p)` and so on.

gossipsub.go
```diff
@@ -1609,7 +1626,7 @@ func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}
 	for p := range gs.p.topics[topic] {
 		_, inExclude := exclude[p]
 		_, direct := gs.direct[p]
-		if !inExclude && !direct && (gs.peers[p] == GossipSubID_v10 || gs.peers[p] == GossipSubID_v11) && gs.score.Score(p) >= gs.gossipThreshold {
+		if !inExclude && !direct && gs.peerHasAny(p, GossipSubID_v10, GossipSubID_v11) && gs.score.Score(p) >= gs.gossipThreshold {
```
vyzo, line 1629: and here again, the feature could be `GossipSubFeatureMesh`

gossipsub.go
```diff
@@ -1777,7 +1794,7 @@ func (gs *GossipSubRouter) getPeers(topic string, count int, filter func(peer.ID
 
 	peers := make([]peer.ID, 0, len(tmap))
 	for p := range tmap {
-		if (gs.peers[p] == GossipSubID_v10 || gs.peers[p] == GossipSubID_v11) && filter(p) {
+		if gs.peerHasAny(p, GossipSubID_v10, GossipSubID_v11) && filter(p) {
```
vyzo, line 1797: here again the feature would be `GossipSubFeatureMesh`

gossipsub.go
```diff
@@ -163,21 +163,22 @@ var (
 	GossipSubIWantFollowupTime = 3 * time.Second
 )
 
-// NewGossipSub returns a new PubSub object using GossipSubRouter as the router.
-func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
+// NewGossipSubWithProtocols returns a new PubSub object using GossipSubRouter as the router and the protocols specified in ps.
+func NewGossipSubWithProtocols(ctx context.Context, h host.Host, ps []protocol.ID, opts ...Option) (*PubSub, error) {
```
vyzo, line 167: this is much better handled with a `WithGossipSubProtocols(protocols, featureCheckFun)` option.

vyzo (commented):
Did you accidentally close this?

PR #410, opened 2021-03-31
Strengthen Validation Of Gossip Scoring Parameters

nisdas:
During further testing of our parameters for peer scoring, we came across some unusual scores for a significant amount of our peers, with a large amount giving their score as `NaN`. On further investigation it came about that when initializing the scoring parameter there had been an arithmetic error when determining the parameter. This lead to the parameter being initialized with a negative infinite value.  The parameter validation was unable to capture this, which is the main reason this PR has been opened up.

- [x] Add in validation for both `NaN` and `Infinite` floating point values and throw an error if any of the topic parameters are initialized as such.
- [x] Add regression test for it.

### DIFF
--- a/score_params.go
+++ b/score_params.go
@@ -32,20 +32,20 @@ type PeerScoreThresholds struct {
 }
 
 func (p *PeerScoreThresholds) validate() error {
-	if p.GossipThreshold > 0 {
-		return fmt.Errorf("invalid gossip threshold; it must be <= 0")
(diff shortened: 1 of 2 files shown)


vyzo:
@nisdas do you mind rebasing on master? I just merged #406 which disables the flaky test that keeps failing on travis.

vyzo (approved):

vyzo:
thank you!

PR #412, opened 2021-04-01
Ignore transient connections

vyzo:
So that we don't try to do pubsub over them.


notify.go
```diff
@@ -34,9 +39,22 @@ func (p *PubSubNotif) ListenClose(n network.Network, _ ma.Multiaddr) {
...
-	for _, pr := range p.host.Network().Peers() {
+	isTransient := func(pid peer.ID) bool {
+		for _, c := range p.host.Network().ConnsToPeer(pid) {
+			if !c.Stat().Transient {
+				return false
+			}
+		}
+
+		return true
+	}
```
Stebalien, line 50: Gah. We really need a better peerstore/database system.

Stebalien (approved):

notify.go
vyzo: yes, indeed!

vyzo:
rebased on master for merge.

Stebalien (approved):

raulk (approved):

PR #417, opened 2021-04-20
add go-check workflow, fix gofmt, go vet and staticcheck

marten-seemann:
Closes #415. Closes #416.

I see that running `go test` is more complicated on this repo, so we should probably not move it to GitHub Actions (yet).
@coryschwartz did an excellent job fixing all the `go vet` and `staticcheck` errors. We should merge these changes, and add the `go-check` workflow.

### DIFF
--- a/discovery_test.go
+++ b/discovery_test.go
@@ -292,6 +292,7 @@ func TestGossipSubDiscoveryAfterBootstrap(t *testing.T) {
 	}
 }
 
+//lint:ignore U1000 // function is only unused because TestGossipSubDiscoveryAfterBootstrap is skipped
 func waitUntilGossipsubMeshCount(ps *PubSub, topic string, count int) {
 	done := false
 	doneCh := make(chan bool, 1)
--- a/gossipsub_spam_test.go
+++ b/gossipsub_spam_test.go
@@ -198,6 +198,7 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
 				})
 
 				go func() {
+					defer close(errs)
 					defer cancel()
 
 					// Wait for a short interval to make sure the legit host
@@ -219,6 +220,7 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
 					iwc := getIWantCount()
 					if iwc > GossipSubMaxIHaveLength {
 						errs <- fmt.Errorf("Expecting max %d IWANTs per heartbeat but received %d", GossipSubMaxIHaveLength, iwc)
+						return
 					}
 					firstBatchCount := iwc
 
@@ -226,6 +228,7 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
 					score := ps.rt.(*GossipSubRouter).score.Score(attacker.ID())
 					if score != 0 {
 						errs <- fmt.Errorf("Expected 0 score, but got %f", score)
+						return
 					}
 
 					// Send a bunch of IHAVEs
@@ -242,10 +245,12 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
 					iwc = getIWantCount()
 					if iwc == firstBatchCount {
 						errs <- fmt.Errorf("Expecting to receive more IWANTs after heartbeat but did not")
+						return
 					}
 					// Should not be more than the maximum per heartbeat
 					if iwc-firstBatchCount > GossipSubMaxIHaveLength {
 						errs <- fmt.Errorf("Expecting max %d IWANTs per heartbeat but received %d", GossipSubMaxIHaveLength, iwc-firstBatchCount)
+						return
 					}
 
 					time.Sleep(GossipSubIWantFollowupTime)
@@ -254,8 +259,8 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
 					score = ps.rt.(*GossipSubRouter).score.Score(attacker.ID())
 					if score >= 0 {
 						errs <- fmt.Errorf("Expected negative score, but got %f", score)
+						return
 					}
-					close(errs)
 				}()
 			}
 		}
@@ -422,6 +427,7 @@ func TestGossipsubAttackGRAFTDuringBackoff(t *testing.T) {
 				})
 
 				go func() {
+					defer close(errs)
 					defer cancel()
 
 					// Wait for a short interval to make sure the legit host
@@ -432,6 +438,7 @@ func TestGossipsubAttackGRAFTDuringBackoff(t *testing.T) {
 					pc := getPruneCount()
 					if pc != 0 {
 						errs <- fmt.Errorf("Expected %d PRUNE messages but got %d", 0, pc)
+						return
 					}
 
 					// Send a PRUNE to remove the attacker node from the legit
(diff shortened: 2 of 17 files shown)


marten-seemann:
No idea why Circle is failing here. Seems to be unrelated to this PR. Some `git clone` failure?

gossipsub_spam_test.go
```diff
@@ -240,24 +244,31 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
...
-						t.Fatalf("Expected negative score, but got %f", score)
+						errs <- fmt.Errorf("Expected negative score, but got %f", score)
+						return
 					}
 				}()
 			}
 		}
 
+		for err := range errs {
+			t.Fatal(err)
```
mvdan: or just make the calls above `t.Errorf`. a Fatal here is a bit weird, as the first one will exit the entire test.

topic_test.go
```diff
@@ -167,7 +167,7 @@ func TestTopicReuse(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if bytes.Compare(msg.GetData(), firstMsg) != 0 {
+	if bytes.Equal(msg.GetData(), firstMsg) {
```
mvdan, line 170: should all these not be `!bytes.Equal`?

marten-seemann: I think it should. From `bytes.Compare` documentation: `The result will be 0 if a==b, -1 if a < b, and +1 if a > b.`

vyzo:
The CircleCI failure is odd, seems it can't checkout the code because of some key issue -- maybe credential rotation?

discovery_test.go
```diff
@@ -291,18 +302,3 @@ func TestGossipSubDiscoveryAfterBootstrap(t *testing.T) {
 		}
 	}
 }
-
-func waitUntilGossipsubMeshCount(ps *PubSub, topic string, count int) {
```
vyzo, line 295: can't we tell staticchek somehow to not barf here? I don't like having it inline.

gossipsub_spam_test.go
```diff
@@ -405,16 +416,18 @@ func TestGossipsubAttackGRAFTDuringBackoff(t *testing.T) {
 
 	newMockGS(ctx, t, attacker, func(writeMsg func(*pb.RPC), irpc *pb.RPC) {
 		// When the legit host connects it will send us its subscriptions
+		errs := make(chan error)
```
vyzo, line 419: we need to make this buffered or we might end up with stuck goroutines.

gossipsub_spam_test.go
```diff
@@ -188,15 +188,17 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
...
```
vyzo, line 191: this should be buffered i think.

gossipsub_spam_test.go
vyzo: yeah, `Errorf` is better here.

gossipsub_spam_test.go
```diff
@@ -531,12 +554,17 @@ func TestGossipsubAttackGRAFTDuringBackoff(t *testing.T) {
...
-						t.Fatal("Expected to not be in the mesh of the legitimate host")
+						errs <- fmt.Errorf("Expected to not be in the mesh of the legitimate host")
+						return
 					}
 				}()
 			}
 		}
 
+		for err := range errs {
+			t.Fatal(err)
```
vyzo, line 565: yeah, let's make that an `Errof`.

gossipsub_spam_test.go
```diff
@@ -677,21 +708,28 @@ func TestGossipsubAttackInvalidMessageSpam(t *testing.T) {
...
-						t.Fatal("Expected attacker node to be PRUNED when score drops low enough")
+						errs <- fmt.Errorf("Expected attacker node to be PRUNED when score drops low enough")
+						return
 					}
 				}()
 			}
 		}
 
+		for err := range errs {
+			t.Fatal(err)
```
vyzo, line 730: `Errorf` I guess too.

topic_test.go
vyzo: yeah, it's fine.

vyzo (commented):
some nits, other than that looks good.

I'll try to figure out the circle checkout issue, seems like they might have rotated some creds following the codecov debacle and we are htting that.

vyzo:
We might need to refresh the deploy key in the project settings? I don't have admin to do that however, @marten-seemann do you have perms?
Otherwise, @stebalien might be able to help.

discovery_test.go
marten-seemann: This is indeed a not very helpful staticcheck message. It says the function is unused, apparently because the test is skipped.

Smells like a false positive to me: https://github.com/dominikh/go-tools/issues/968.

I'll just add a `nolint` comment here, I agree that inlining is not a nice solution.

jacobheun:
> We might need to refresh the deploy key in the project settings? I don't have admin to do that however, @marten-seemann do you have perms?

I missed a few readds when I cycled keys due to the codecov incident last week. Remaining repos should have the new keys now

discovery_test.go
marten-seemann: Done.

gossipsub_spam_test.go
```diff
@@ -405,16 +416,18 @@ func TestGossipsubAttackGRAFTDuringBackoff(t *testing.T) {
...
-				graft := []*pb.ControlGraft{&pb.ControlGraft{TopicID: sub.Topicid}}
+				graft := []*pb.ControlGraft{{TopicID: sub.Topicid}}
 				writeMsg(&pb.RPC{
-					Subscriptions: []*pb.RPC_SubOpts{&pb.RPC_SubOpts{Subscribe: sub.Subscribe, Topicid: sub.Topicid}},
+					Subscriptions: []*pb.RPC_SubOpts{{Subscribe: sub.Subscribe, Topicid: sub.Topicid}},
 					Control:       &pb.ControlMessage{Graft: graft},
 				})
 
 				go func() {
+					defer close(errs)
```
marten-seemann, line 430: That will cause duplicate `close`s of the channel.

gossipsub_spam_test.go
```diff
@@ -188,15 +188,17 @@ func TestGossipsubAttackSpamIHAVE(t *testing.T) {
...
 				// Reply by subcribing to the topic and grafting to the peer
 				writeMsg(&pb.RPC{
-					Subscriptions: []*pb.RPC_SubOpts{&pb.RPC_SubOpts{Subscribe: sub.Subscribe, Topicid: sub.Topicid}},
-					Control:       &pb.ControlMessage{Graft: []*pb.ControlGraft{&pb.ControlGraft{TopicID: sub.Topicid}}},
+					Subscriptions: []*pb.RPC_SubOpts{{Subscribe: sub.Subscribe, Topicid: sub.Topicid}},
+					Control:       &pb.ControlMessage{Graft: []*pb.ControlGraft{{TopicID: sub.Topicid}}},
 				})
 
 				go func() {
+					defer close(errs)
```
marten-seemann, line 201: This will cause duplicate `close`s of the channel.

gossipsub_spam_test.go
```diff
@@ -642,20 +670,23 @@ func TestGossipsubAttackInvalidMessageSpam(t *testing.T) {
...
```
marten-seemann, line 683: This will cause duplicate `close`s of the channel.

vyzo:
the changes with the error channels seem to have broken some tests...

mvdan:
I think all the new channels should be unnecessary; just use Errorf where the original code used Fatalf :)

aschmahmann:
@marten-seemann @vyzo what's the status of this PR?

marten-seemann:
@aschmahmann Due to the resource requirements (especially fds) of the test suite here, we removed this repo from the automation: https://github.com/protocol/.github/pull/51. We can also close this PR.

PR #422, opened 2021-05-11
feat: re-generate protobufs with gogofaster

Stebalien:
This removes the pointer indirection from all values and saves us an allocation. We never actually try to distinguish between a field being missing and the "zero" value anyways.

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -32,8 +32,8 @@ func (p *PubSub) getHelloPacket() *RPC {
 
 	for t := range subscriptions {
 		as := &pb.RPC_SubOpts{
-			Topicid:   proto.String(t),
-			Subscribe: proto.Bool(true),
+			Topicid:   t,
+			Subscribe: true,
 		}
 		rpc.Subscriptions = append(rpc.Subscriptions, as)
 	}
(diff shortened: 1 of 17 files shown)


Stebalien:
Motivation: pubsub does a noticeable amount of allocation. We were noticing this on a gateway due to pubsub handshake messages.

Stebalien:
It looks like this is changing the number of messages we're sending, possibly because we're sending "zero values"? We _shouldn't_ be doing that, but it looks like that might be the case.

vyzo:
yeah, the test failure is interesting -- rerunning to see if it is a fluke.

vyzo (commented):
I am in favor of this, but we have to double check for nil checks that would indicate the missing value. 
I think that's what might be happening with the test failure.

Stebalien:
> I think that's what might be happening with the test failure.

Ah... we're likely checking "is this slice nil". Unfortunately, if I change:

```go
var x *[]byte
if x == nil
```
```go
var x []byte
if x == nil
```

Nothing will break. Will check.

Stebalien:
Ok, found it. It's because the topic ID will always be encoded now, even if unset. This is more complicated than it's worth for now.

PR #424, opened 2021-05-13
Update README to point to correct example directory

medmouine:


marten-seemann (approved):
Thank you @medmouine!
No idea what CircleCI is doing here, but I'm going to merge this anyway.

PR #429, opened 2021-07-12
expose more event for RawTracer and add an option for configuring subscription queue size

MichaelMure:
Possibly `RecvRPC` and `SendRPC` could be added as well as they follow the same format. After that, only `PublishMessage` seems to be missing in `RawTracer`.


vyzo (commented):
you also need to implement it in the various internal tracers! Just an empty stub will do.

vyzo (approved):
thank you!

vyzo:
not sure what's up with Circle -- it seems _stuck_.

vyzo (approved):
LGTM!

vyzo:
CI is still misbehaving, let me try to close and reopen.

PR #430, opened 2021-07-12
fix goroutine build up from connected notifications

vyzo:
As observed in the gateway.

Aggregates notifications and batch processes, while being careful not to block the event loop.

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -13,7 +13,6 @@ import (
 	"github.com/libp2p/go-msgio/protoio"
 
 	"github.com/gogo/protobuf/proto"
-	ms "github.com/multiformats/go-multistream"
 )
 
 // get the initial RPC containing all of our subscriptions to send to new peers
@@ -106,14 +105,11 @@ func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing <-chan
 	if err != nil {
 		log.Debug("opening new stream to peer: ", err, pid)
 
-		if err == ms.ErrNotSupported {
-			select {
-			case p.newPeerError <- pid:
-			case <-ctx.Done():
-			}
-		} else {
-			p.notifyPeerDead(pid)
+		select {
+		case p.newPeerError <- pid:
+		case <-ctx.Done():
 		}
+
 		return
 	}
 
--- a/comm.go
+++ b/comm.go
@@ -88,21 +88,31 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 	}
 }
 
+func (p *PubSub) notifyPeerDead(pid peer.ID) {
+	p.peerDeadPrioLk.RLock()
+	p.peerDeadMx.Lock()
+	p.peerDeadPend[pid] = struct{}{}
+	p.peerDeadMx.Unlock()
+	p.peerDeadPrioLk.RUnlock()
+
+	select {
+	case p.peerDead <- struct{}{}:
+	default:
+	}
+}
+
 func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing <-chan *RPC) {
 	s, err := p.host.NewStream(p.ctx, pid, p.rt.Protocols()...)
 	if err != nil {
 		log.Debug("opening new stream to peer: ", err, pid)
 
-		var ch chan peer.ID
 		if err == ms.ErrNotSupported {
-			ch = p.newPeerError
+			select {
+			case p.newPeerError <- pid:
+			case <-ctx.Done():
+			}
 		} else {
-			ch = p.peerDead
-		}
-
-		select {
-		case ch <- pid:
-		case <-ctx.Done():
+			p.notifyPeerDead(pid)
 		}
 		return
 	}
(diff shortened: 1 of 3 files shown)


pubsub.go
```diff
@@ -621,6 +614,43 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+	}
+
+	if len(p.newPeersPend) == 0 {
+		return
+	}
+
+	newPeers := p.newPeersPend
+	p.newPeersPend = make(map[peer.ID]struct{})
+
+	for pid := range newPeers {
```
aschmahmann, line 605: Are we concerned that if this map is large we could block the event loop?

pubsub.go
```diff
@@ -621,6 +614,43 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+func (p *PubSub) handlePendingPeers() {
+	select {
+	case <-p.newPeersSema:
+		defer func() {
+			p.newPeersSema <- struct{}{}
+		}()
+
+	default:
+		// contention, return and wait for the next notification without blocking the event loop
+		return
```
aschmahmann: This seems like it might be ok, although is it a problem if there's contention for a long time?

aschmahmann (commented):
Left a couple of comments/questions, generally I get the changes and trust your instincts here.

pubsub.go
vyzo: well, if it gets large enough to take a non trivial amount of time to process, we've got bigger problems! We are getting flooded with connections.

pubsub.go
vyzo: i don't think so, at worst some peers wont be processed for a while; we can live with that.

pubsub.go
```diff
@@ -260,6 +264,8 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		counter:               uint64(time.Now().UnixNano()),
 	}
 
+	ps.newPeersSema <- struct{}{}
```
Stebalien: nit: I generally prefer the pattern where _writing_ takes the lock, and reading unlocks. Just means you can avoid this initialization part.

pubsub.go
```diff
@@ -621,6 +614,43 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+		return
+	}
+
+	if len(p.newPeersPend) == 0 {
+		return
+	}
+
+	newPeers := p.newPeersPend
+	p.newPeersPend = make(map[peer.ID]struct{})
+
```
Stebalien, line 604: Can we release the "lock" right here? There's no reason to hold it.

pubsub.go
```diff
@@ -231,7 +233,9 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		signKey:               nil,
 		signPolicy:            StrictSign,
 		incoming:              make(chan *RPC, 32),
-		newPeers:              make(chan peer.ID),
+		newPeers:              make(chan struct{}, 1),
+		newPeersSema:          make(chan struct{}, 1),
```
Stebalien: I'd really just use a lock here. We're just protecting reads/writes to a single map.

pubsub.go
vyzo: sure

pubsub.go
vyzo: Not really -- we can block the event loop indefinitely if there is a lot of contention in the lock; better to use a semaphore that allows us to try lock.

vyzo:
rebased on master and also:
- added batching for dead peer notifications 
- fixed the handling of errors from opening new streams so that we can no longer go into loops of respawning writers because we failed to open a stream.

pubsub.go
vyzo: done

comm.go
```diff
@@ -88,22 +87,29 @@ func (p *PubSub) handleNewStream(s network.Stream) {
...
 func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing <-chan *RPC) {
 	s, err := p.host.NewStream(p.ctx, pid, p.rt.Protocols()...)
 	if err != nil {
 		log.Debug("opening new stream to peer: ", err, pid)
 
-		var ch chan peer.ID
-		if err == ms.ErrNotSupported {
-			ch = p.newPeerError
-		} else {
-			ch = p.peerDead
```
Stebalien, line 100: Why are we getting rid of this? Can't we fail to create a new stream if, e.g., the connection is dead? This will leave us without a peer if we try to use the wrong connection.

pubsub.go
```diff
@@ -621,6 +590,83 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+
+		close(ch)
+
+		if p.host.Network().Connectedness(pid) == network.Connected {
+			// still connected, must be a duplicate connection being closed.
+			// we respawn the writer as we need to ensure there is a stream active
+			log.Debugf("peer declared dead but still connected; respawning writer: %s", pid)
+			messages := make(chan *RPC, p.peerOutboundQueueSize)
+			messages <- p.getHelloPacket()
+			go p.handleNewPeer(p.ctx, pid, messages)
```
Stebalien, line 653: You know, _this_ might have been the source of backing up handleNewPeer goroutines... It's possible that we were seeing "unresponsive" peers that still appeared to be connected?

Maybe we need some form of connection blacklist? Or maybe a bit of backoff?

Stebalien (approved):

comm.go
vyzo: it can lead to some very nasty "peer declared dead; respawning writer" loops, which have been reported by lotus users in the wild.

Not sure how to properly handle this situation, let me give it a bit more thought -- it's subtle and what we have right now is very problematic.

pubsub.go
vyzo: yeah, same problem as above; we need a more principled approach.

comm.go
vyzo: and that's why I removed it.

Stebalien: Ah, I see. Yeah. This is a crap situation. I'm just concerned that this will lead to the _opposite_ problem, but I think this is fine for now. But we should put in a better fix later.

pubsub.go
Stebalien: :sob:

comm.go
vyzo: Yeah, I am concerned about it too. I'll open a follow up issue so that we don't forget.

vyzo: follow up issue in https://github.com/libp2p/go-libp2p-pubsub/issues/433

PR #432, opened 2021-07-13
More tracing, option to configure the Subscribe output queue length

MichaelMure:
Replace https://github.com/libp2p/go-libp2p-pubsub/pull/429 in an attempt to properly trigger the CI.


vyzo (approved):
thank you!

vyzo:
ok, the CI situation is not getting better; I am going to run the tests locally so that we can force merge.

vyzo:
tests work for me locally.

PR #435, opened 2021-07-19
cleanup: fix vet and staticcheck failures

iand:
Fixes the safe staticcheck failures. The following remain and need more thought or explicit ignores.

May indicate a bug, not clear what the intention is here:
```
gossipsub.go:904:4: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
```

Needs a decision on the alternative:
```
gossipsub.go:1300:8: log.EventBegin is deprecated: Stop using go-log for event logging  (SA1019)
```

Changing these may break clients who rely on error string text:
```
validation.go:146:15: error strings should not be capitalized (ST1005)
validation.go:173:15: error strings should not be capitalized (ST1005)
validation.go:209:15: error strings should not be capitalized (ST1005)
validation.go:387:9: error strings should not be capitalized (ST1005)
```

### DIFF
--- a/discovery_test.go
+++ b/discovery_test.go
@@ -292,6 +292,7 @@ func TestGossipSubDiscoveryAfterBootstrap(t *testing.T) {
 	}
 }
 
+//lint:ignore U1000 used only by skipped tests at present
 func waitUntilGossipsubMeshCount(ps *PubSub, topic string, count int) {
 	done := false
 	doneCh := make(chan bool, 1)
--- a/gossip_tracer_test.go
+++ b/gossip_tracer_test.go
@@ -18,12 +18,10 @@ func TestBrokenPromises(t *testing.T) {
 	peerB := peer.ID("B")
 	peerC := peer.ID("C")
 
-	var msgs []*pb.Message
 	var mids []string
 	for i := 0; i < 100; i++ {
 		m := makeTestMessage(i)
 		m.From = []byte(peerA)
-		msgs = append(msgs, m)
 		mid := DefaultMsgIdFn(m)
 		mids = append(mids, mid)
 	}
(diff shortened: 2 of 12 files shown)


vyzo:
> gossipsub.go:904:4: ineffective break statement

That thing with the break is really innocuous, we can just remove it.

> gossipsub.go:1300:8: log.EventBegin is deprecated: Stop using go-log for event logging  (SA1019)

We can just replace it with a `log.Infow` that logs the timing

> Changing these may break clients who rely on error string text:

That's just silly/stupid; just go ahead and change them.

vyzo (commented):
maybe add some short-circuit code for breaking the test when we have goroutines failing? Or wait for them. Not sure what's prudent.

iand:
> breaking

Is this a reference to the replacement of t.Fatal with t.Error+return in test goroutines? In these cases the test still fails.

vyzo:
yeah, just worried it might get blocked somewhere.

iand:
> yeah, just worried it might get blocked somewhere.

OK. In `TestGossipsubAttackSpamIHAVE` returning from the goroutine cancels the context. I could use that to ensure that the other goroutines exit too (we should do that anyway to avoid other sources of blocking). Will look at the other tests as well.

iand:
> > yeah, just worried it might get blocked somewhere.
> 
> OK. In `TestGossipsubAttackSpamIHAVE` returning from the goroutine cancels the context. I could use that to ensure that the other goroutines exit too (we should do that anyway to avoid other sources of blocking). Will look at the other tests as well.

Done this in https://github.com/libp2p/go-libp2p-pubsub/pull/435/commits/4238ca7c078c6c632641a9730aea9e1d689ba637

Stebalien (approved):

PR #439, opened 2021-07-28
fix: downgrade multiaddr

Stebalien:
The new version doesn't work with the current websocket release. We can upgrade in the next minor release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,8 +10,8 @@ require (
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
 	github.com/libp2p/go-libp2p-core v0.8.6
 	github.com/libp2p/go-libp2p-discovery v0.5.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
 	github.com/libp2p/go-libp2p-core v0.8.6
 	github.com/libp2p/go-libp2p-discovery v0.5.1
-	github.com/libp2p/go-libp2p-swarm v0.5.2
+	github.com/libp2p/go-libp2p-swarm v0.5.3
```
vyzo, line 13: is that intentional?

vyzo (approved):

go.mod
Stebalien: Yes. I needed to release new transports and a new swarm.

PR #440, opened 2021-07-29
add support for custom protocol matching function

richard-ramos:
This allows setting a custom protocol matching function.
```
fullTextMatch := func(expectedProtocol string) func(string) bool {
    return func(protocolToVerify string) bool {
	return protocolToVerify == expectedProtocol
    }
}

// create a new PubSub service using the GossipSub router
ps, err := pubsub.NewGossipSub(ctx, h, pubsub.WithProtocolMatchFunction(fullTextMatch))
```

Could potentially be used to allow semver matching
```go
protoMinor := protocol.ID("/testing/1.2.0")
mfunc, err := helpers.MultistreamSemverMatcher(protoMinor)
if err != nil {
    t.Fatal(err)
}
ps, err := pubsub.NewGossipSub(ctx, h, 
     pubsub.WithProtocolMatchFunction(mfunc),
     pubsub.WithGossipSubProtocols([]protocol.ID{pubsub.GossipSubID_v11, pubsub.GossipSubID_v10, pubsub.FloodSubID, protoMinor}, pubsub.GossipSubDefaultFeatures)
)

```

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -36,6 +36,8 @@ var (
 
 var log = logging.Logger("pubsub")
 
+type MatchingFunction func(string) func(string) bool
+
 // PubSub is the implementation of the pubsub system.
 type PubSub struct {
 	// atomic counter for seqnos
@@ -157,6 +159,9 @@ type PubSub struct {
 	// filter for tracking subscriptions in topics of interest; if nil, then we track all subscriptions
 	subFilter SubscriptionFilter
 
+	// protoMatchFunc is a matching function for protocol selection.
+	protoMatchFunc *MatchingFunction
+
 	ctx context.Context
 }
 
@@ -235,6 +240,7 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		peerOutboundQueueSize: 32,
 		signID:                h.ID(),
 		signKey:               nil,
+		protoMatchFunc:        nil,
 		signPolicy:            StrictSign,
 		incoming:              make(chan *RPC, 32),
 		newPeers:              make(chan struct{}, 1),
@@ -292,7 +298,11 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	rt.Attach(ps)
 
 	for _, id := range rt.Protocols() {
-		h.SetStreamHandler(id, ps.handleNewStream)
+		if ps.protoMatchFunc != nil {
+			h.SetStreamHandlerMatch(id, (*ps.protoMatchFunc)(string(id)), ps.handleNewStream)
+		} else {
+			h.SetStreamHandler(id, ps.handleNewStream)
+		}
 	}
 	h.Network().Notify((*PubSubNotif)(ps))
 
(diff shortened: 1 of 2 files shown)


pubsub.go
```diff
@@ -36,6 +36,8 @@ var (
 
 var log = logging.Logger("pubsub")
 
+type MatchingFunction func(string) func(string) bool
```
vyzo: Can you make this an alias and call it `ProtocolMatchFn`?

So
```
type ProtocolMatchFn = ...
```

pubsub.go
```diff
@@ -157,6 +159,9 @@ type PubSub struct {
 	// filter for tracking subscriptions in topics of interest; if nil, then we track all subscriptions
 	subFilter SubscriptionFilter
 
+	// protoMatchFunc is a matching function for protocol selection.
+	protoMatchFunc *MatchingFunction
```
vyzo: no need for a pointer here, you can test a function against nil.

pubsub.go
```diff
@@ -292,7 +298,11 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 	rt.Attach(ps)
 
 	for _, id := range rt.Protocols() {
-		h.SetStreamHandler(id, ps.handleNewStream)
+		if ps.protoMatchFunc != nil {
+			h.SetStreamHandlerMatch(id, (*ps.protoMatchFunc)(string(id)), ps.handleNewStream)
```
vyzo: no need for the indirection if you don't use a pointer.

pubsub.go
```diff
@@ -475,6 +485,15 @@ func WithMaxMessageSize(maxMessageSize int) Option {
 	}
 }
 
+// WithProtocolMatchFunction sets a custom matching function for protocol
+// selection to be used by the protocol handler on the Host's Mux
+func WithProtocolMatchFunction(m MatchingFunction) Option {
+	return func(ps *PubSub) error {
+		ps.protoMatchFunc = &m
```
vyzo: and here, no need to take a pointer.

vyzo (changes requested):
Thank you, this looks useful.

A couple of small changes are needed:
- It should probably be combined with the appropriate feature test function in gossipsub.
  Mind adding some about this in the documentation?
- No need to use a pointer! functions can be tested against nil with `!=`

vyzo:
Also, can you add a test?

pubsub.go
```diff
@@ -235,6 +240,7 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		peerOutboundQueueSize: 32,
 		signID:                h.ID(),
 		signKey:               nil,
+		protoMatchFunc:        nil,
```
vyzo: we don't need to initialize this, the default is fine.

pubsub.go
```diff
@@ -475,6 +485,16 @@ func WithMaxMessageSize(maxMessageSize int) Option {
 	}
 }
 
+// WithProtocolMatchFn sets a custom matching function for protocol selection to
+// be used by the protocol handler on the Host's Mux. Can be combined with
+// WithGossipSubProtocols feature function for checking if a protocol is supported
```
vyzo: I would say here "Should be combined ... for checking if certain protocol features are supported"

pubsub.go
vyzo: Note that we do it for `signKey` to be explicit about it, as these are grouped together.

vyzo (commented):
thank you, this is looking good. Just a couple of minor things to fix and we can merge.

richard-ramos:
I finished the required changes. Thank you for the code review!

vyzo (approved):

vyzo:
thank you!

PR #452, opened 2021-09-21
allow waiting for enough peers in gossipsub

### LINKED ISSUE #454: support waiting until a gossipsub topic has enough peers
Right now, most tests that build a gossipsub mesh use sleeps to wait for the mesh to build:

https://github.com/libp2p/go-libp2p-pubsub/blob/7ef0669764b94ed11e8b5b0128ee408942ebe5eb/gossipsub_test.go#L62-L63

The indexer-reference-provider codebase has the same problem:

https://github.com/filecoin-project/indexer-reference-provider/blob/3ec1151ca682a437c6e38727aa00b1087e5a0276/core/engine/engine_test.go#L186-L188

It seems like it would be enough to wait for enough peers to be connected before publishing advertisements.
Right now, the following options are available:

1) Sleep. This is what all the tests do, but one has to use a very long sleep to reduce the chances of flakes. Even with multi-second sleeps, the CI over at indexer-reference-provider still ran into failures at times.

2) Poll `PubSubRouter.EnoughPeers` until it returns true. Would work, but just like any other form of polling, it's a tradeoff between slowness and overhead.

3) Poll `Topic.ListPeers` until its length reaches a minimum. Same polling problems as 1.

4) Use `WithReadiness` when publishing. Unfortunately, this option only works when `WithDiscovery` is used, and the indexer doesn't use that. The option seems to be a no-op in those cases.

I think we should support some way to do this without sleeping nor polling, nor without requiring `WithDiscovery`. Here are two proposed API additions:

5) Modify `WithReadiness` so that it works in all configurations. Without `WithDiscovery`, it would simply make `Topic.Publish` block until enough peers are connected (or until Publish's `ctx` is cancelled).

6) Add another method, such as `func (*Topic) WaitEnoughPeers(context.Context, int) bool`, which would block until enough peers are connected or ctx is cancelled.

5 seems preferrable to 6, for the sake of making existing APIs more consistently useful and not adding redundant APIs. However, I'm not familiar enough with `WithReadiness` to tell if it's the best approach internally.

mvdan:
(see commit message)

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 
@@ -242,7 +243,44 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error 
 	}
 
 	if pub.ready != nil {
-		t.p.disc.Bootstrap(ctx, t.topic, pub.ready)
+		if t.p.disc.discovery != nil {
+			t.p.disc.Bootstrap(ctx, t.topic, pub.ready)
+		} else {
+			// TODO: we could likely do better than polling every 200ms.
+			// For example, block this goroutine on a channel,
+			// and check again whenever events tell us that the number of
+			// peers has increased.
+			var ticker *time.Ticker
+		readyLoop:
+			for {
+				// Check if ready for publishing.
+				// Similar to what disc.Bootstrap does.
+				res := make(chan bool, 1)
+				select {
+				case t.p.eval <- func() {
+					done, _ := pub.ready(t.p.rt, t.topic)
+					res <- done
+				}:
+					if <-res {
+						break readyLoop
+					}
+				case <-t.p.ctx.Done():
+					return t.p.ctx.Err()
+				case <-ctx.Done():
+					return ctx.Err()
+				}
+				if ticker == nil {
+					ticker = time.NewTicker(200 * time.Millisecond)
+					defer ticker.Stop()
+				}
+
+				select {
+				case <-ticker.C:
+				case <-ctx.Done():
+					return fmt.Errorf("router is not ready: %w", ctx.Err())
+				}
+			}
+		}
 	}
 
 	return t.p.val.PushLocal(&Message{m, t.p.host.ID(), nil})
--- a/topic_test.go
+++ b/topic_test.go
@@ -3,6 +3,7 @@ package pubsub
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"sync"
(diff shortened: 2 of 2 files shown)


mvdan:
FYI @adlrocha @masih @willscott for the time.Sleep TODOs in go-legs and indexer-reference-provider.

mvdan:
@willscott another question is how to expose this via go-legs. Right now it calls NewGossipSub, which creates a GossipSubRouter and feeds it to NewPubSub. Unfortunately, that is then hidden behind an unexported field in `PubSub.rt`, so neither go-legs nor its downstreams can access it to call methods like WaitEnoughPeers.

One option would be to add another method, like:

    func (*PubSub) Router() PubSubRouter

Thoughts, @vyzo?

mvdan:
Yet another option is to leave it up to the downstream to poll `EnoughPeers` in a loop, since that's already part of the `PubSubRouter` interface. That would certainly work, but then we'd be stuck with polling and sleeping forever, without the ability to do better like the TODO says. So I think supporting a standard method like "wait for enough peers" is worthwhile.

I'm new to this module and its design, though, so maybe there's a better design to be had here.

vyzo (changes requested):
This should be part of the pubsub interface, not the router; all routers implement enough peers.
Also, thread safety...

vyzo:
Actually it should be in the topic interface.

vyzo:
and there is also machinery for topic related events, which can make implementation of this much nicer.

vyzo:
and there is already an option that more or less does what you want, it's called `WithReadiness`.

mvdan:
> and there is already an option that more or less does what you want, it's called `WithReadiness`.

I saw that, but it doesn't do anything at all in our case, because we're not using `WithDiscovery`.
Maybe we could teach that option to also work without `WithDiscovery`? It's unclear to me what API would be nicest.

> Actually it should be in the topic interface.

That's fine by me; I was just following `EnoughPeers`. I can rewrite this PR if we can agree on what API to add or modify.

vyzo:
then open an issue to discuss and agree on an api.

mvdan:
> then open an issue to discuss and agree on an api.

Done: https://github.com/libp2p/go-libp2p-pubsub/issues/454

vyzo:
thanks.

mvdan:
I've updated the code to no longer be racy, I think - the added `WithReadiness` logic inside `Publish` now uses the `eval` channel to call the `ready` func, which then calls `EnoughPeers`. Can you take another look?

gossipsub.go
```diff
@@ -573,6 +573,7 @@ func (gs *GossipSubRouter) EnoughPeers(topic string, suggested int) bool {
 		suggested = gs.params.Dlo
 	}
 
+	// println(suggested, gs.params.Dhi, fsPeers+gsPeers, gsPeers)
```
vyzo: can we remove this?

topic.go
```diff
@@ -242,7 +243,39 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
...
+				case <-ctx.Done():
+					return ctx.Err()
+				}
+
+				// TODO: we could likely do better than sleeping 200ms.
+				// For example, block this goroutine on a channel,
+				// and check again whenever events tell us that the number of
+				// peers has increased.
+				select {
+				case <-time.After(200 * time.Millisecond):
```
vyzo: we should use a ticker here.

vyzo (commented):
i would say this is good enough.

However, please don't change the gossipsub tests to remove the sleep, we want to test mesh formation with heartbeats.

mvdan:
Hm, I'm not sure I understand removing the sleeps from tests. Sleeping for two seconds or sleep-polling until the equivalent condition is met seems basically the same, just that sleeping makes the tests slower. I'm happy to revert those changes though, as I'm not the maintainer.

If I'm not touching any of the existing tests, then it sounds like I need to add a new test to cover this feature.

vyzo:
it's not the same condition; one allows the mesh to form, the other waits
for number of peers.

Yes, please add a new test for the readiness condition.

topic.go
mvdan: done

mvdan:
All done, I think.

topic.go
```diff
@@ -242,7 +243,41 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
...
 	if pub.ready != nil {
-		t.p.disc.Bootstrap(ctx, t.topic, pub.ready)
+		if t.p.disc.discovery != nil {
+			t.p.disc.Bootstrap(ctx, t.topic, pub.ready)
+		} else {
+			// TODO: we could likely do better than polling every 200ms.
+			// For example, block this goroutine on a channel,
+			// and check again whenever events tell us that the number of
+			// peers has increased.
+			ticker := time.NewTicker(200 * time.Millisecond)
```
vyzo: an improvement here is to only instantiate the ticker after making an optimistic attempt; this way we only pay the price once.

vyzo (commented):
looks pretty good - we just need to avoid unnecessary ticker allocations once we've got enough peers.

topic.go
mvdan: done

vyzo (approved):
thank you!

PR #459, opened 2021-11-10
feat: plumb through context changes

guseggert:


marten-seemann (approved):

vyzo (approved):
do we have to wait before merfing?

guseggert:
> do we have to wait before merfing?

No I'll merge it and then release a new version since I need to plumb this through go-ipfs.

PR #463, opened 2021-12-22
discovery.go: Include topic size information in `MinTopicSize` godoc.

### LINKED ISSUE #462: discovery.go: `MinTopicSize` does not include publishing host in `TopicSize` count.
`MinTopicSize` is confusing as `TopicSize` is interpreted as **total** number of peers listening in on a topic, but in reality, the publishing host excludes itself from that number. It would be nice if this were either to be documented in the godocs or if `TopicSize` were to be changed so that it includes the publishing host in the count.

renaynay:
This PR resolves #462 .


vyzo (approved):
thank you!

PR #465, opened 2021-12-28
feat: custom per-topic MsgIdFunction

### LINKED ISSUE #464: Per-topic MsgIdFunction
Pubsub only allows setting global `MsgIdFunction`, which applies to all the topics. Still, It would be useful to have the ability to set custom per topic msg id generation, which, e.g., can be dependant on the msg format of some topic. For this to be implemented using the global `MsgIdFunction`, it would need to be aware of all the msg types of each topic and do some ugly switching to determine what `MsgIdFunction` to use for what msg. 

Fortunately, an unused `TopicOpt` can set custom values while joining the topic, so some per topic customizations are conceived by design, and the described feature can be implemented.

Wondertan:
An attempt to close #464

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -20,7 +20,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/protocol"
 
 	logging "github.com/ipfs/go-log"
-	timecache "github.com/whyrusleeping/timecache"
+	"github.com/whyrusleeping/timecache"
 )
 
 // DefaultMaximumMessageSize is 1mb.
@@ -213,6 +213,7 @@ const (
 
 type Message struct {
 	*pb.Message
+	ID            string
 	ReceivedFrom  peer.ID
 	ValidatorData interface{}
 }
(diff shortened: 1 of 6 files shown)


pubsub.go
```diff
@@ -213,6 +213,7 @@ const (
 
 type Message struct {
 	*pb.Message
+	ID            string
```
Wondertan, line 216: This allows msg id to be generated only once. Before it was generated 3 times in a pipeline, which can be expensive when hashing is used. Also, this simplifies code a bit and removes the requirement to MsgIdFunction into MessageCache.

Wondertan: Fun fact: One of the [Mina Protocol's motivations](https://github.com/MinaProtocol/mina/blob/rfc/bitswap/rfcs/0042-bitswap.md#motivation) to rework their protocol is high CPU spikes on msg id generation.

pubsub.go
```diff
@@ -1164,6 +1175,15 @@ type TopicOptions struct{}
 
 type TopicOpt func(t *Topic) error
 
+// WithCustomMessageIDFunc is an option to customize the way a message ID is computed for a pubsub message
+// within the Topic. By default, it uses global MsgIdFunction registered on a PubSub instance.
+func WithCustomMessageIDFunc(msgId MsgIdFunction) TopicOpt {
```
Wondertan, line 1180: Not sure about a good naming here. Added a work `Custom` to differ from global Pubsub option.

Wondertan:
I will add a test once this or another solution is considered good.

vyzo (commented):
See comment on issue.

Note that computing the ID once and caching it is valuabke on its own, you can cherry pick this for another pr.

Wondertan:
Ok, will do

vyzo:
thank you!

BigLep:
@vyzo : are we good to merge?

vyzo:
no, this should be closed. We have discussed a better implementatio (see the issue) which will supersede this.

Wondertan:
I am going to work over it on weekends.

PR #466, opened 2021-12-28
Compute msg id only once

Wondertan:
Cherry-picked from #465


Wondertan:
All the tests are passing locally btw

vyzo (commented):
so after reviewing this, lets go forward and implement the suggested design?
That way we dont have to change any of this code, we'll have msgid method on the new object do the cahing.
All the SetMsgId calls stay, and we pass the method ref.

What do you think?

Wondertan:
Ok. Replied in the issue.

PR #468, opened 2022-01-09
feat: msgIdGenerator

### LINKED ISSUE #464: Per-topic MsgIdFunction

Wondertan:
* Introduces `msgIdGenerator` 
* Ensures msg id is generated once only
	* Besides `Tracer `methods rcving `*RPC`
* Allows setting per-topic `MsgIDFunction`

Closes #464

> Personally, I still prefer the solution, where we compute and set Message.ID once in the pipeline and then assume it is set everywhere, instead of calling generator everywhere, but I might be missing something

### DIFF
--- a/gossip_tracer.go
+++ b/gossip_tracer.go
@@ -15,7 +15,7 @@ import (
 type gossipTracer struct {
 	sync.Mutex
 
-	msgID MsgIdFunction
+	idGen *msgIDGenerator
 
 	followUpTime time.Duration
 
@@ -29,7 +29,7 @@ type gossipTracer struct {
 
 func newGossipTracer() *gossipTracer {
 	return &gossipTracer{
-		msgID:        DefaultMsgIdFn,
+		idGen:        newMsgIdGenerator(),
 		promises:     make(map[string]map[peer.ID]time.Time),
 		peerPromises: make(map[peer.ID]map[string]struct{}),
 	}
@@ -40,7 +40,7 @@ func (gt *gossipTracer) Start(gs *GossipSubRouter) {
 		return
 	}
 
-	gt.msgID = gs.p.msgID
+	gt.idGen = gs.p.idGen
 	gt.followUpTime = gs.params.IWantFollowupTime
 }
 
@@ -117,7 +117,7 @@ func (gt *gossipTracer) GetBrokenPromises() map[peer.ID]int {
 var _ RawTracer = (*gossipTracer)(nil)
 
 func (gt *gossipTracer) fulfillPromise(msg *Message) {
-	mid := gt.msgID(msg.Message)
+	mid := gt.idGen.ID(msg)
 
 	gt.Lock()
 	defer gt.Unlock()
--- a/mcache.go
+++ b/mcache.go
@@ -3,8 +3,6 @@ package pubsub
 import (
 	"fmt"
 
-	pb "github.com/libp2p/go-libp2p-pubsub/pb"
-
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
@@ -27,23 +25,25 @@ func NewMessageCache(gossip, history int) *MessageCache {
 		panic(err)
 	}
 	return &MessageCache{
-		msgs:    make(map[string]*pb.Message),
+		msgs:    make(map[string]*Message),
 		peertx:  make(map[string]map[peer.ID]int),
 		history: make([][]CacheEntry, history),
 		gossip:  gossip,
-		msgID:   DefaultMsgIdFn,
+		msgID: func(msg *Message) string {
+			return DefaultMsgIdFn(msg.Message)
+		},
 	}
 }
 
 type MessageCache struct {
-	msgs    map[string]*pb.Message
+	msgs    map[string]*Message
 	peertx  map[string]map[peer.ID]int
 	history [][]CacheEntry
 	gossip  int
-	msgID   MsgIdFunction
+	msgID   func(*Message) string
 }
 
-func (mc *MessageCache) SetMsgIdFn(msgID MsgIdFunction) {
+func (mc *MessageCache) SetMsgIdFn(msgID func(*Message) string) {
 	mc.msgID = msgID
 }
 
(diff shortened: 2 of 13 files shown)


gossip_tracer.go
```diff
@@ -29,7 +29,7 @@ type gossipTracer struct {
 
 func newGossipTracer() *gossipTracer {
 	return &gossipTracer{
-		msgID:        DefaultMsgIdFn,
+		idGen:        newMsgIdGenerator(),
```
Wondertan, line 32: Can be removed, as anyway set on Start. Kept matching the existing pattern.

mcache.go
```diff
@@ -27,23 +25,25 @@ func NewMessageCache(gossip, history int) *MessageCache {
...
+	msgs    map[string]*Message
 	peertx  map[string]map[peer.ID]int
 	history [][]CacheEntry
 	gossip  int
-	msgID   MsgIdFunction
+	msgID   func(*Message) string
 }
 
-func (mc *MessageCache) SetMsgIdFn(msgID MsgIdFunction) {
+func (mc *MessageCache) SetMsgIdFn(msgID func(*Message) string) {
```
Wondertan, line 46: Slightly breaks API. `NewMessageCache` is public and this method as well, but there is no way around and I dought it is used somewhere.

mcache.go
```diff
@@ -27,23 +25,25 @@ func NewMessageCache(gossip, history int) *MessageCache {
...
 }
 
 type MessageCache struct {
-	msgs    map[string]*pb.Message
+	msgs    map[string]*Message
 	peertx  map[string]map[peer.ID]int
 	history [][]CacheEntry
 	gossip  int
-	msgID   MsgIdFunction
+	msgID   func(*Message) string
```
Wondertan, line 43: We can consider referencing here a `msgIdGenerator`, similarly to how it is done in tracers. Also make `SetMsgIdFn` a `SetMsgIdGen`, anyway it is getting changed.

mcache_test.go
```diff
@@ -18,7 +18,7 @@ func TestMessageCache(t *testing.T) {
 	}
 
 	for i := 0; i < 10; i++ {
-		mcache.Put(msgs[i])
+		mcache.Put(&Message{Message: msgs[i]})
```
Wondertan, line 21: To avoid changing `makeTestMessage`, but can do if considered good.

gossip_tracer.go
Wondertan: Same with other tracers

pubsub.go
```diff
@@ -326,11 +327,7 @@ type MsgIdFunction func(pmsg *pb.Message) string
 // but it can be customized to e.g. the hash of the message.
 func WithMessageIdFn(fn MsgIdFunction) Option {
 	return func(p *PubSub) error {
-		p.msgID = fn
-		// the tracer Option may already be set. Update its message ID function to make options order-independent.
-		if p.tracer != nil {
-			p.tracer.msgID = fn
-		}
```
Wondertan, line 333: Not relevant anymore as now `msgIdGenerator` keeping the default is shared

trace.go
```diff
@@ -344,7 +344,7 @@ func (t *pubsubTracer) traceRPCMeta(rpc *RPC) *pb.TraceEvent_RPCMeta {
 	var msgs []*pb.TraceEvent_MessageMeta
 	for _, m := range rpc.Publish {
 		msgs = append(msgs, &pb.TraceEvent_MessageMeta{
-			MessageID: []byte(t.msgID(m)),
+			MessageID: []byte(t.idGen.ID(&Message{Message: m})),
```
Wondertan: No way to get `Message`, but should not be a big issue.

gossip_tracer.go
vyzo: yeah, let's drop this.

vyzo: actually wait, tests might need it so let's keep it.

mcache.go
vyzo: not a problem, i doubt anyone uses outside the package.

mcache.go
vyzo: let's keep the func, easier testability.

midgen.go
```diff
@@ -0,0 +1,45 @@
...
+	"sync"
+)
+
+// msgIDGenerator handles computing IDs for msgs
+// It allows setting custom generators(MsgIdFunction) per topic
+type msgIDGenerator struct {
+	Default MsgIdFunction
+
+	topicGens   map[string]MsgIdFunction
+	topicGensLk sync.RWMutex
```
vyzo, line 14: nit: can you put the lock on top so that it is clear that it protects the member underneat?
This pattern is called the mutex hat.

vyzo (commented):
Thank you, this looks pretty good!

Left a couple of comments suggesting some minor improvements and then it's ready to merge.

pubsub.go
vyzo: :+1:

trace.go
vyzo: I think we should make a method in idGen that does this, no need to alloc here.

mcache_test.go
vyzo: maybe it will save some characters? Up to you, doesn't really matter.

vyzo:
Basically the one  small thing that we can and should improve on is provide a raw ID method in idGen, so that we don't have to allocate a wrapper message to compute it in the tracer.

This method could simply be refactored out of ID, and used by it.

trace.go
vyzo: Or we can expose a secondary method, refactor out of ID (so that it calls it) and call it RawID or something?
It will be nicer, so we might as well do it.

vyzo:
hey @Wondertan, when will you have some time to finish this? I would like to merge it soon.

It is very close, just needs to fix the allocation in the tracer by refactoring rawMsgInd.
It also needs a rebase.

Wondertan:
Hey @vyzo, planned for tomorrow.

midgen.go
Wondertan: Mutex hat, like this. Done!

trace.go
Wondertan: Good idea. Done

Wondertan:
Rebased on master

Wondertan:
One last thing I am concerned about is naming for options. 
`WithMessageIdFn` is global options
`WithMsgIdFunction` is for topic. Would be cool to unify them somehow or just append `Topic` in the end.

I will also add a test for the option later today.

vyzo:
yeah, lets call it `WithTopicMessageId` or something similar.

Wondertan:
Ready. 

Btw, why do you use real Swarm and not mocknet in tests? What does this give for testing?

vyzo (approved):
looks good, thank you!

vyzo:
> Btw, why do you use real Swarm and not mocknet in tests? What does this give for testing?

Well, we want some integration testing happening and observing effects of the actual network in use; mocknet would deny us that.

PR #473, opened 2022-02-07
Add Backoff For Pruned Peers

### LINKED ISSUE #367: Gossipsub Backoffs not stored in GossipSubRouter::sendPrune
Every time we send a prune via [`GossipSubRouter::sendPrune`](https://github.com/libp2p/go-libp2p-pubsub/blob/4ccb63821e80f5f05f4894f98c2f6a90e64a1894/gossipsub.go#L1007) we include our backoff specified in the config [here](https://github.com/libp2p/go-libp2p-pubsub/blob/4ccb63821e80f5f05f4894f98c2f6a90e64a1894/gossipsub.go#L1752). But we never actually add this backoff for the given peer and topic to our data structure (we add it in many other situations like [here](https://github.com/libp2p/go-libp2p-pubsub/blob/4ccb63821e80f5f05f4894f98c2f6a90e64a1894/gossipsub.go#L684), but not in the situation where we send a `PRUNE`).

This result of this is that if we `PRUNE` a peer we might try to re`GRAFT` immediately (or shortly after) which will result in a punishment by the receiving peer because of the backoff he received. Therefore I assume this is a bug.

A scenario, where this bug might occur heavily is if we leave a topic (this calls sendPrune on all peers of this topic) and then join it again shortly after.

nisdas:
This PR resolves the issue described in #367. 
- [x] To prevent unnecessary penalties due to eager re-grafting of peers, we now add a backoff when leaving a topic and pruning the respective peers from our topic mesh.  
- [x] Adds in a regression test for this case.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1093,7 +1093,7 @@ func (gs *GossipSubRouter) Leave(topic string) {
 		gs.sendPrune(p, topic)
 		// Add a backoff to this peer to prevent us from eagerly
 		// re-grafting this peer into our mesh if we rejoin this
-		// topic before the backoff period.
+		// topic before the backoff period ends.
 		gs.addBackoff(p, topic)
 	}
 }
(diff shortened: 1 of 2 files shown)


vyzo (approved):
thank you!

PR #474, opened 2022-02-08
Add in Backoff Filter When Grafting New Peers

nisdas:
This is a follow up from #473. 

When we rejoin a topic in a short duration, behavioural penalties can be triggered if we try to re-graft previously pruned peers. This is prevalent especially if we rejoin a topic before the prune backoff period ends.

- [x]  When joining a topic, we make sure we do not graft peers who have a backoff assigned to them. This brings the method in line with the behaviour and peer filter that is applied in our `heartbeat` method. 
- [x] Adds in a regression test for this particular case.

cc @vyzo

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1036,10 +1036,12 @@ func (gs *GossipSubRouter) Join(topic string) {
 
 	gmap, ok = gs.fanout[topic]
 	if ok {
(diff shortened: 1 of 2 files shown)


vyzo (approved):
thank you!

PR #477, opened 2022-03-17
Implement lazyHello (fix #460)

### LINKED ISSUE #460: About the possibility to skip the HelloPacket
At Berty we implement our conversation groups based on [go-orbit-db](https://github.com/berty/go-orbit-db) stores (go-orbit-db is a golang port of [js-orbit-db](https://github.com/orbitdb/orbit-db)) which themselves rely on libp2p pubsub.

Since Berty users will need to exchange messages asynchronously (send a message to a receiver who's offline), we want to provide some highly available nodes. Their role would be to replicate the messages of a group to which they have been added (NB: they won't be able to decipher the messages) and forward messages to members of the conversation who were offline at the time of sending.

To do this, we would need to have some kind of **supernode** capable of subscribing to a very large amount of pubsub topics. But we are facing a problem: due to the gossip implementation, each peers on the pubsub will exchange the topics they are subscribed to ([HelloPacket](https://github.com/libp2p/go-libp2p-pubsub/blob/3b98004eed436cce836cf3222e1c1db7c9c52c65/comm.go#L19)) and as mentionned on https://github.com/libp2p/go-libp2p-pubsub/issues/402, it will silently fail if too many topics have been joined and the list exceeds the maximum message size. The **supernodes** we are planing to setup will rapidly reach this limit.

Do you think it would be possible / desirable to have a supernode mode for the pubsub that would not require to exchange the list of topics?

Do you have any suggestion on this?

d-roak:
Does not emit the eager hello when a node has the lazy flag.
When a non-lazy node receives a graft message, it adds the TopicID to the subscriptions that are going to be processed.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/discovery"
@@ -163,6 +164,9 @@ type PubSub struct {
 
 	// protoMatchFunc is a matching function for protocol selection.
 	protoMatchFunc ProtocolMatchFn
+
+	// lazyHello is a flag to enable lazy hello.
+	lazyHello bool
 
 	ctx context.Context
 }
@@ -510,6 +514,13 @@ func WithProtocolMatchFn(m ProtocolMatchFn) Option {
 	}
 }
 
+func WithLazyHello(lazy bool) Option {
+	return func(ps *PubSub) error {
+		ps.lazyHello = lazy
+		return nil
+	}
+}
+
 // processLoop handles all inputs arriving on the channels
 func (p *PubSub) processLoop(ctx context.Context) {
 	defer func() {
@@ -654,7 +665,9 @@ func (p *PubSub) handlePendingPeers() {
 		}
 
 		messages := make(chan *RPC, p.peerOutboundQueueSize)
-		messages <- p.getHelloPacket()
+		if !p.lazyHello {
+			messages <- p.getHelloPacket()
+		}
 		go p.handleNewPeer(p.ctx, pid, messages)
 		p.peers[pid] = messages
 	}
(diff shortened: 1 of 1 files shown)


pubsub.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"google.golang.org/protobuf/proto"
```
vyzo, line 14: we use gogo, pls dont import this.

vyzo (changes requested):
this is not going to work, it will never receive a graft if it hasnt send a hello.

Also, i would prefer to not conflate with gs semantics, it is incompatible with floodsub. So we should not process any control events for this.

I think the right way foreard is to send back a hello packet if it receives one, and only include topics of interest.
These topics should be decided by a user defined function, passed through the lazy option.

I am guessing we would also want to Relay those topics (not subscribe), so the application should also be given the chance to do so, maybe indicated as the respone from the user supplied function (maybe object?).
Or it could emit an event to let the application decide whether to subscribe or relay to relevant topics.

We'll have to think about this a bit more.

vyzo:
there are also some security considerations, as this could become a DoS vector.

d-roak:
> this is not going to work, it will never receive a graft if it hasnt send a hello.

The non-lazy node receives a graft message even without the hello (i tested it).

> Also, i would prefer to not conflate with gs semantics, it is incompatible with floodsub. So we should not process any control events for this.

Yes, I agree!

> I think the right way foreard is to send back a hello packet if it receives one, and only include topics of interest. These topics should be decided by a user defined function, passed through the lazy option.
> 
> I am guessing we would also want to Relay those topics (not subscribe), so the application should also be given the chance to do so, maybe indicated as the respone from the user supplied function (maybe object?). Or it could emit an event to let the application decide whether to subscribe or relay to relevant topics.
> 
> We'll have to think about this a bit more.

What if when a lazy node receives a "subscriptions" rpc call, it calls the `handleNewPeer` subroutine with the subscription packet he received (if it's in the `mySubs` or `myRelays`)?
I am trying to avoid a `replyHelloPacket` approach, but if you think is the best approach, we can go with that.

> there are also some security considerations, as this could become a DoS vector.

What is the attack vector you are thinking?

d-roak:
Hi @vyzo,

Can you check the last comment, please?

PR #481, opened 2022-03-24
feat: WithLocalPublication option to enable local only publishing on a topic

### LINKED ISSUE #480: Option to enable local only publishing on a topic
There are use cases where it becomes handy to publish msgs only to local subscribers, e.g. some edge cases where the outdated payload needs processing through the local pipeline but is irrelevant to the network under certain conditions.

This can be simply implemented using `PubOpt` and by setting the new `Local` bool field to `Message`, so that it is not `Router.Publish`ed

Wondertan:
An attempt to close https://github.com/libp2p/go-libp2p-pubsub/issues/480

> I know that it is better to discuss the solution first, but waiting is not fun while coding is and I am ok to redo that if we find a better solution.

We can also sneak here similar cases like an option to publish to the mesh only, so doing something like the code below during validation is not needed anymore:
```go
if self == peer {
	return 
}
```

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -221,6 +221,7 @@ type Message struct {
 	ID            string
 	ReceivedFrom  peer.ID
 	ValidatorData interface{}
+	Local bool
 }
 
 func (m *Message) GetFrom() peer.ID {
@@ -1066,7 +1067,7 @@ func (p *PubSub) handleIncomingRPC(rpc *RPC) {
 				continue
 			}
 
-			p.pushMsg(&Message{pmsg, "", rpc.from, nil})
+			p.pushMsg(&Message{pmsg, "", rpc.from, nil, false})
 		}
 	}
 
@@ -1165,7 +1166,9 @@ func (p *PubSub) checkSigningPolicy(msg *Message) error {
 func (p *PubSub) publishMessage(msg *Message) {
 	p.tracer.DeliverMessage(msg)
 	p.notifySubs(msg)
-	p.rt.Publish(msg)
+	if !msg.Local {
+		p.rt.Publish(msg)
+	}
 }
 
 type addTopicReq struct {
--- a/topic.go
+++ b/topic.go
@@ -320,8 +320,11 @@ func WithReadiness(ready RouterReady) PubOpt {
 	}
 }
 
-// WithLocalPublication option tells Publish to *only* notify local subscribers about a message.
-// This option prevents messages publication to peers.
+// WithLocalPublication returns a publishing option to notify in-process subscribers only.
+// It prevents message publication to mesh peers.
+// Useful in edge cases where the msg needs to be only delivered to the in-process subscribers,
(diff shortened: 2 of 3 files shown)


pubsub.go
```diff
@@ -1143,7 +1144,9 @@ func (p *PubSub) checkSigningPolicy(msg *Message) error {
 func (p *PubSub) publishMessage(msg *Message) {
 	p.tracer.DeliverMessage(msg)
```
Wondertan, line 1167: Should we trace such local messages or not?

Wondertan:
@vyzo, gentle ping. Do you think this is a reasonable publish option? Should we proceed, or close the PR?

vyzo:
well, i am a little conflicted about it.
Do we have a concrete use case? If so, yes.

vyzo:
leave it open for now, i'd like to see others in the community chime in as well.
My concern is that i dont want people to use this as in process pubsub, there are specialized libs for that.
On the other hand, i can see it being useful in some cases, but i'd like to make that concrete.

Wondertan:
Got your concern. We don't use PubSub for in-process pubsubing only, but we use it for pubsubing when msgs need to be _both_ delivered to a local instance of some component and remote instances on peers. However, there is an edge case where the _local_ instance is out of sync and only it needs to rcv msgs, s.t. network is not spammed with no more relevant msgs. And we would like to solve this case over the same code path and without the need to spawn another internal pubsubing mechanism specifically for the case. I can give you more details and even point to sources if you want.

topic.go
```diff
@@ -319,6 +320,15 @@ func WithReadiness(ready RouterReady) PubOpt {
 	}
 }
 
+// WithLocalPublication option tells Publish to *only* notify local subscribers about a message.
+// This option prevents messages publication to peers.
```
vyzo: Lets improve the documentation a bit.

vyzo (commented):
ok, fair enough.
Lets improve the documentation a bit and i will merge.

topic.go
Wondertan: > // WithLocalPublication returns a publishing option to notify in-process subscribers only.
> // It prevents message publication to mesh peers.
> // Useful in *edge* cases where the msg needs to be only delivered to the in-process subscribers, 
> // e.g. not to spam the network with outdated msgs.
> // Should not be used specifically for in-process pubsubing.

Tried to be consistent with the wording on other pub opts and incorporate your concern.

vyzo: good enough!

Wondertan: Ok, will update the PR then

Wondertan: Done

pubsub.go
vyzo: the tracer can see it's a local publication from the Local field -- but we might want to have some special casing in some tracers.
I think we should leave it as for now.

vyzo (approved):
thank you!

PR #483, opened 2022-04-06
Adds exponential backoff to re-spawing new streams for supposedly dead peers

### LINKED ISSUE #482: Backoff / retry for writer stream respawn
Lately we’ve been seeing [this error](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/pubsub.go#L686) get printed a lot. It’s causing enormous periodic goroutine spikes, sometimes up to 200k goroutines, and we can see the same message being printed for the exact same peer multiple times within a very short time window.

We captured a profile of the program while it was running, and this is [the line](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/comm.go#L118) that all the goroutines are blocked on.

What’s I believe is happening is that when [this line](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/comm.go#L117) of [handleNewPeer](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/comm.go#L103) gets called, the newly created stream is for some reason erroring immediately, which causes [this](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/comm.go#L131), which eventually leads to [handleNewPeer being called again for the same peer](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/pubsub.go#L689). At this point, the previous call is still blocked [here](https://github.com/libp2p/go-libp2p-pubsub/blob/be065ce0510e81d820a2cdb9762e63fd012122ba/comm.go#L119) because the consumer of `p.newPeerStream` is slow (or simply because it is overwhelmed with a large number of messages to process). So now, we have two goroutines blocked on that channel send for the same peer. And this can repeat indefinitely.

We should implement some backoff / retry logic to prevent more than one goroutine simultaneously blocked on that channel send for the same peer.

thep2p:
Addresses https://github.com/libp2p/go-libp2p-pubsub/issues/482 by implementing an exponential backoff mechanism for re-spawning new streams to peers that are originally supposedly dead. 

closes https://github.com/libp2p/go-libp2p-pubsub/issues/482

### DIFF
--- a/backoff.go
+++ b/backoff.go
@@ -48,7 +48,7 @@ func newBackoff(ctx context.Context, sizeThreshold int, cleanupInterval time.Dur
 	return b
 }
 
-func (b *backoff) updateAndGet(id peer.ID) (time.Duration, bool) {
+func (b *backoff) updateAndGet(id peer.ID) time.Duration {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
@@ -71,13 +71,20 @@ func (b *backoff) updateAndGet(id peer.ID) (time.Duration, bool) {
 
 	h.lastTried = time.Now()
 	h.attempts += 1
-	if h.attempts > b.maxAttempts {
-		delete(b.info, id)
-		return 0, false
-	}
 
 	b.info[id] = h
-	return h.duration, true
+	return h.duration
+}
+
+func (b *backoff) peerExceededBackoffThreshold(id peer.ID) bool {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	h, ok := b.info[id]
+	if !ok {
+		return false // no record of this peer is still there, hence fine.
+	}
+	return h.attempts > b.maxAttempts
 }
 
 func (b *backoff) cleanup() {
--- a/backoff.go
+++ b/backoff.go
@@ -15,7 +15,7 @@ const (
 	TimeToLive             = 10 * time.Minute
 	BackoffCleanupInterval = 1 * time.Minute
 	BackoffMultiplier      = 2
-	MaxBackoffJitterCoff   = 1000
+	MaxBackoffJitterCoff   = 100
 )
 
 type backoffHistory struct {
--- a/backoff.go
+++ b/backoff.go
@@ -1,6 +1,7 @@
 package pubsub
 
 import (
+	"context"
 	"sync"
 	"time"
 
@@ -10,8 +11,9 @@ import (
 const (
 	MinBackoffDelay   = 100 * time.Millisecond
 	MaxBackoffDelay   = 10 * time.Second
-	TimeToLive        = 10 * time.Minute
-	BackoffMultiplier = 2
+	TimeToLive             = 10 * time.Minute
+	BackoffCleanupInterval = 1 * time.Minute
+	BackoffMultiplier      = 2
 )
 
 type backoffHistory struct {
@@ -22,15 +24,21 @@ type backoffHistory struct {
 type backoff struct {
 	mu   sync.Mutex
 	info map[peer.ID]*backoffHistory
-	ct   int // size threshold that kicks off the cleaner
+	ct   int           // size threshold that kicks off the cleaner
+	ci   time.Duration // cleanup intervals
 }
 
-func newBackoff(sizeThreshold int) *backoff {
-	return &backoff{
+func newBackoff(ctx context.Context, sizeThreshold int, cleanupInterval time.Duration) *backoff {
+	b := &backoff{
 		mu:   sync.Mutex{},
 		ct:   sizeThreshold,
+		ci:   cleanupInterval,
 		info: make(map[peer.ID]*backoffHistory),
 	}
+
+	go b.cleanupLoop(ctx)
+
+	return b
 }
 
 func (b *backoff) updateAndGet(id peer.ID) time.Duration {
@@ -55,17 +63,27 @@ func (b *backoff) updateAndGet(id peer.ID) time.Duration {
 	h.lastTried = time.Now()
 	b.info[id] = h
 
-	if len(b.info) > b.ct {
-		b.cleanup()
-	}
-
 	return h.duration
 }
 
 func (b *backoff) cleanup() {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
 	for id, h := range b.info {
 		if time.Since(h.lastTried) > TimeToLive {
 			delete(b.info, id)
+		}
+	}
+}
+
+func (b *backoff) cleanupLoop(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return // pubsub shutting down
+		case <-time.Tick(b.ci):
+			b.cleanup()
 		}
 	}
 }
--- a/backoff.go
+++ b/backoff.go
@@ -2,6 +2,7 @@ package pubsub
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"sync"
 	"time"
@@ -48,7 +49,7 @@ func newBackoff(ctx context.Context, sizeThreshold int, cleanupInterval time.Dur
 	return b
 }
 
-func (b *backoff) updateAndGet(id peer.ID) time.Duration {
+func (b *backoff) updateAndGet(id peer.ID) (time.Duration, error) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
@@ -60,6 +61,8 @@ func (b *backoff) updateAndGet(id peer.ID) time.Duration {
 			duration: time.Duration(0),
 			attempts: 0,
 		}
+	case h.attempts > b.maxAttempts:
+		return 0, fmt.Errorf("peer %s has reached its maximum backoff attempts", id)
 
 	case h.duration < MinBackoffDelay:
 		h.duration = MinBackoffDelay
@@ -72,22 +75,10 @@ func (b *backoff) updateAndGet(id peer.ID) time.Duration {
 		}
 	}
 
-	h.lastTried = time.Now()
 	h.attempts += 1
-
+	h.lastTried = time.Now()
 	b.info[id] = h
-	return h.duration
-}
-
-func (b *backoff) peerExceededBackoffThreshold(id peer.ID) bool {
-	b.mu.Lock()
-	defer b.mu.Unlock()
-
-	h, ok := b.info[id]
-	if !ok {
-		return false // no record of this peer is still there, hence fine.
-	}
-	return h.attempts > b.maxAttempts
+	return h.duration, nil
 }
 
 func (b *backoff) cleanup() {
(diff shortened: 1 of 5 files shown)


synzhu (commented):
We need some mechanisms to prune this new datastructure once peers get disconnected, otherwise it could be a memory leak.

We also should reset the backoff timer after a while. We can track the time when the last request happened, and if we are more than some threshold past that time, then we can reset the timer and just return the base timeout.

synzhu (commented):
Also, I'm not sure if we need to handle canceling a pending reconnect  if the peer becomes disconnected?

Maybe before calling `handleNewPeer` we need to first check if the peer is still connected? But not sure if this is needed.

backoff.go
```diff
@@ -0,0 +1,71 @@
...
+			h.duration = MaxBackoffDelay
+		}
+	}
+
+	h.lastTried = time.Now()
+	b.info[id] = h
+
+	if len(b.info) > b.ct {
+		b.cleanup()
+	}
```
synzhu: nit: if we have a lot of peers (more than `b.ct`), then we will end up running `b.cleanup()` on every call to this function (which requires looping though the entire map), even when there is nothing to cleanup. Perhaps we need a better way to handle this?

For example, if we maintain an additional heap datastructure which is sorted by the expiration time, then at every call to the function we can just pop all the expired entries from the heap one by one and remove them from the map. This would require us to track the explicit expiration time in each `backoffHistory`, rather than `lastTried`.

Will leave to @vyzo to decide whether this is necessary.

backoff.go
```diff
@@ -0,0 +1,71 @@
...
+	h, ok := b.info[id]
+	if !ok || time.Since(h.lastTried) > TimeToLive {
+		// first request goes immediately.
+		h = &backoffHistory{
+			duration: time.Duration(0),
+		}
+	} else if h.duration < MinBackoffDelay {
+		h.duration = MinBackoffDelay
+	} else if h.duration < MaxBackoffDelay {
+		h.duration = time.Duration(BackoffMultiplier * h.duration)
```
vyzo: can we add some jitter?

backoff.go
vyzo: Lets see if it is actually a problem in practice

vyzo: Actually it might be problematic, lets run a background goroutine that does it periodically (say once a minute).

vyzo (commented):
looks mostly good, just a couple of things:
- lets add some jitter in the backoff
- lets do the cleanup in a bacground goroutine.

backoff.go
thep2p: Added in https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/a9f4edfbfdf537be3fa16d4792e2a81b25dfefee

backoff.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/a77d4355e47d85fc0e7cb6373f9e1927e7add3de

comm.go
```diff
@@ -121,6 +122,16 @@ func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing <-chan
 	}
 }
 
+func (p *PubSub) handleNewPeerWithBackoff(ctx context.Context, pid peer.ID, outgoing <-chan *RPC) {
+	delay := p.deadPeerBackoff.updateAndGet(pid)
```
vyzo: I think we need to add a failure more if we have backed off too much and simply give up; say we try up to 10 times and then `updateAndGet` returns an error and we close the channel and forget the peer.

How does that sound?

vyzo (commented):
Looks pretty good, but see comment about giving up eventually.

I am concerned about attackability of the code without this failsafe, if an attacker figured out a way to trigger it.

comm.go
vyzo: maybe 10 is even too much, 3-4 attempts should be enough.

backoff.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return b
+}
+
+func (b *backoff) updateAndGet(id peer.ID) (time.Duration, bool) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	h, ok := b.info[id]
+	if !ok || time.Since(h.lastTried) > TimeToLive {
```
vyzo: let's write this if/else sequence with a switch, will be nicer.

backoff.go
```diff
@@ -0,0 +1,103 @@
...
+		h.duration = MinBackoffDelay
+	} else if h.duration < MaxBackoffDelay {
+		jitter := rand.Intn(MaxBackoffJitterCoff)
+		h.duration = (BackoffMultiplier * h.duration) + time.Duration(jitter)*time.Millisecond
+		if h.duration > MaxBackoffDelay || h.duration < 0 {
+			h.duration = MaxBackoffDelay
+		}
+	}
+
+	h.lastTried = time.Now()
```
vyzo: let's get the time after checking the max attempts, will avoid the gettimeofday call in that case.

comm.go
```diff
@@ -121,6 +123,22 @@ func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing <-chan
 	}
 }
 
+func (p *PubSub) handleNewPeerWithBackoff(ctx context.Context, pid peer.ID, outgoing <-chan *RPC) error {
+	delay, valid := p.deadPeerBackoff.updateAndGet(pid)
+	if !valid {
+		return fmt.Errorf("backoff attempts to %s expired after reaching maximum allowed", pid)
+	}
```
vyzo: let's return the error directly from `updateAndGet` instead of a bool, makes for simpler code.

vyzo (commented):
Looks pretty good, some cosmetic suggestions for cleaner code.

backoff.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/7f815f0a9c4e957c775ce3c1f10a704221671284

comm.go
thep2p: I decoupled updating the backoff history from checking for the number of backoff attempts. Hence now `updateAndGet` [only increments](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/7f815f0a9c4e957c775ce3c1f10a704221671284/backoff.go#L76) the backoff attempts without returning any boolean or error indicating the maximum attempts reached. Instead, I introduced a separate [`peerExceededBackoffThreshold`](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/7f815f0a9c4e957c775ce3c1f10a704221671284/backoff.go#L82-L90) method that checks whether the backoff attempts of a peer exceeds the defined threshold. The reasons for this change are: 

1. We want to close the channel and forget the peer if the backoff attempts go beyond a threshold. The [`updateAndGet` is called on a goroutine](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/6ebc292245fe234c3c078ef87362697ab2c1ed6a/pubsub.go#L694) while the `messages` channel and `peers` map are residing on a separate `goroutine`. So, if we let `updateAndGet` return a backoff attempt exceeding error and the `goroutine` that it resides on attempts on closing the `messages` channel and forgetting peer from `peers` map, it results in a race condition between the two `goroutines`, and also the code is vulnerable to panic when this [`defer`](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/6ebc292245fe234c3c078ef87362697ab2c1ed6a/pubsub.go#L518-L522) function is executed, as it is trying to close an already closed channel, i.e.,the `messages` channel that we closed on `updateAndGet`. 

2. By this decoupling, we [invoke `peerExceededBackoffThreshold` on the parent `goroutine`](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/6ebc292245fe234c3c078ef87362697ab2c1ed6a/pubsub.go#L687), hence we give up on backing off if the peer exceeds the backoff threshold rightaway, without opening any channel, spawning any child `goroutine` for `updateAndGet`. Moreover, the peer [is forgotten](https://github.com/yhassanzadeh13/go-libp2p-pubsub/blob/6ebc292245fe234c3c078ef87362697ab2c1ed6a/pubsub.go#L699) by the subsequent lines. Hence, no vulnerability for race conditions and less resource allocation-deallocations. 

Please let me know how does it sound?

backoff.go
thep2p: Please read my reply to the below comment as this part has got changed.

comm.go
thep2p: Done

comm.go
vyzo: sounds better than what we had :)

pubsub.go
```diff
@@ -683,19 +683,15 @@ func (p *PubSub) handleDeadPeers() {
 
 		close(ch)
 
-		if p.host.Network().Connectedness(pid) == network.Connected {
+		if p.host.Network().Connectedness(pid) == network.Connected &&
+			!p.deadPeerBackoff.peerExceededBackoffThreshold(pid) {
```
vyzo: we might want to (debug) log this.

vyzo (commented):
yes, this looks better and we avoid spawning the goroutine if we are going to stop.

One minor point is that the check/get is not atomic; I would suggest taking the base delay before spawning the goroutine, with an error return if it is execeeded.
Then check the error and log if this is exceeded (debug, no need to spam), and then spawn with the delay we got.

Does that sounds reasonable?

pubsub.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/8b649669f077268770b6efd85333ff55053e5b97

https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/c00510e104b020dc61a7d19cf01c3011f3081888

backoff.go
```diff
@@ -0,0 +1,104 @@
...
+	defer b.mu.Unlock()
+
+	for id, h := range b.info {
+		if time.Since(h.lastTried) > TimeToLive {
+			delete(b.info, id)
+		}
+	}
+}
+
+func (b *backoff) cleanupLoop(ctx context.Context) {
```
vyzo, line 95: you know, we should use a ticker object here to reduce allocation/goro pressure.
```
ticker := time.NewTicker(...)
defer ticker.Stop()
```

vyzo (commented):
Ok, i am pretty happy with this.
Lets use a ticker object in the bg loop and it is ready to merge.

backoff.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/483/commits/eede9ba0b8d26ab38e8314c7b9723efe62be8dac

vyzo (commented):
LGTM. Thank you!

vyzo (approved):

PR #484, opened 2022-04-30
fix unused GossipSubHistoryGossip, make seenMessages ttl configurable

protolambda:
This PR fixes a few minor issues:
- The `GossipSubHistoryGossip` global was unused, the config was defaulting `HistoryGossip` to the same `GossipSubHistoryLength` as `HistoryLength`
- Make `seenMessages` cache TTL configurable without having to change the global. Eth2 and optimism both customize this differently, it should not only just be a global var.
- Make `PeerScoreParams` not use the global `TimeCacheDuration` by adding a SeenMsgTTL field. It will default to the global value if the new field is not specified, so no breaking changes.
- Update dependencies: the tests would not run in Go 1.18 because of the (indirect) quic dependency not yet supporting Go 1.18. Some other libp2p libs already updated. Not sure about coordination of releases, but doesn't look like these dependency updates break anything.


vyzo (approved):
Sorry for the delay, i only noticed it now. Can you tag me for review in futute prs?

PR #486, opened 2022-05-25
Publishing option for signing a message with a custom private key

iulianpascalau:
- added publishing option for signing a message with a custom private key


vyzo (commented):
This looks fine, but would it make more sense to introduce a topic option `WithPublisher` and using that?

topic.go
```diff
@@ -286,6 +293,54 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
 	return t.p.val.PushLocal(&Message{m, "", t.p.host.ID(), nil})
 }
 
+// PublishWithSk publishes data to topic using the provided secret key and generated peer ID. Providing also the peer ID
+// (which can be cached) will improve the performance as it does not require the generation of the public key
+// and the resulting peer ID.
+// This method can be used as to be able to publish messages on the network without having a "real", connectable host.
+func (t *Topic) PublishWithSk(ctx context.Context, data []byte, signKey crypto.PrivKey, pid peer.ID, opts ...PubOpt) error {
```
vyzo: let's not abbreviate the Sk and say `PublishWithKey` or `PublishWithSecretKey`.

iulianpascalau: > This looks fine, but would it make more sense to introduce a topic option `WithPublisher` and using that?

I will explore that. I was on the verge of pushing a new commit for code duplication removal.

iulianpascalau:
> This looks fine, but would it make more sense to introduce a topic option `WithPublisher` and using that?

I will explore that. I was on the verge of pushing a new commit for code duplication removal.

topic.go
iulianpascalau: will refactor

topic.go
```diff
@@ -216,27 +227,40 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
...
+		}
+	}
+
 	m := &pb.Message{
 		Data:  data,
 		Topic: &t.topic,
 		From:  nil,
 		Seqno: nil,
 	}
 	if t.p.signID != "" {
```
vyzo: BUG: this should check the local pid variable.

topic.go
```diff
@@ -216,27 +227,40 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
...
-		m.From = []byte(t.p.signID)
+		m.From = []byte(pid)
 		m.Seqno = t.p.nextSeqno()
 	}
 	if t.p.signKey != nil {
```
vyzo: BUG: this should check the local key variable

vyzo (changes requested):
there are a couple of bugs, but otherwise looks good.

topic.go
iulianpascalau: right, fixed. Thanks

iulianpascalau: right, fixed. Thanks

topic.go
iulianpascalau: right, fixed. Thanks

topic_test.go
```diff
@@ -920,3 +921,101 @@ func TestWithTopicMsgIdFunction(t *testing.T) {
 		t.Fatal("msg ids are equal")
 	}
 }
+
+func TestTopic_PublishWithKeyInvalidParameters(t *testing.T) {
```
vyzo: can we remove the underscore?

topic_test.go
```diff
@@ -920,3 +921,101 @@ func TestWithTopicMsgIdFunction(t *testing.T) {
...
+	t.Run("empty peer ID should error", func(t *testing.T) {
+		withVirtualKey := WithSecretKeyAndPeerId(virtualPeer.PrivKey, "")
+		err := topics[0].Publish(ctx, []byte("buff"), withVirtualKey)
+		if err != ErrEmptyPeerID {
+			t.Fatal("error should have been of type errEmptyPeerID")
+		}
+	})
+}
+
+func TestTopicRelay_PublishWithKey(t *testing.T) {
```
vyzo: here too

vyzo (commented):
just some final nits and it is ready to merge.

topic_test.go
iulianpascalau: sure

iulianpascalau: this is the Elrond pattern used in tests :)

iulianpascalau: done

topic_test.go
iulianpascalau: done

vyzo (approved):
thank you!

Issue #489, opened 2022-06-14
Transient connections control


chux0519:
It would be nice if we can control whether to use transient connections.

[libp2p  v0.19.0](https://github.com/libp2p/go-libp2p/releases/tag/v0.19.0) published circuit-v2, if we want to use those circuit-v2 nodes in pubsub , transient connections should be enabled.
 
For now, pubsub doesnt support transient connections (https://github.com/libp2p/go-libp2p-pubsub/pull/412)

I made some change to support it, will PR later.

vyzo:
using transient connections in pubsub is meaningless unfortunately.

chux0519:
> using transient connections in pubsub is meaningless unfortunately.

Thanks, I see.
So if I want to use pubsub while using circuit v2, I better to configure with "no limit" in the relay service to make sure the connection is not transient.

vyzo:
for now yes, eventually we want to improve the api and allow apps to use high limit v2 conns.

PR #490, opened 2022-06-14
feat: support transient connections control

chux0519:
pubsub now will call `network.GetUseTransient(ctx)` to decide whether to
use transient connections

issue: https://github.com/libp2p/go-libp2p-pubsub/issues/489


vyzo (changes requested):
no, this is not supported by design. Pubsub needs stable connections.

Issue #494, opened 2022-07-19
Publish call high latency


aratz-lasa:
Hello,

The `Publish` call, in my application, is taking in average `5ms` to complete. Moreover, it makes use of an internal lock. Therefore, if you execute `Publish` N times in parallel, it takes `5ms+N` to complete all of them. 

Taking this into account, I have a question: What is the lock used for, apart for accessing the `t.closed` bool? Is there a way to get rid of it? Right now the Publish call it is a huge bottleneck.

vyzo:
That's unfortunate.

Are you using the topic publish method or the pubsub level Publish method? 

Can you profile this?
Also, can you check with 0.6/0.7?
Lets see if there is a regression.

aratz-lasa:
First of all, thank you for your fast response!

I have profiled it and the most time takes it at signing messages. If I disable signing it reduces from 5ms to 0.3ms. So I guess it's not bad. However, I do want to sign the messages. Moreover, there is no really a regression from 0.6 to 0.7.

However, I would like to know, what is the lock needed for?

vyzo:
Are you using RSA keys for some reason?

Can you try with ed25529 keys?
It should be considerably faster.

vyzo:
The lock is used to prevent data races, maybe we can get rid of it indeed.

aratz-lasa:
I checked and we do use RSA, because when you create a new `Host` in libp2p, if you don't specify the `Identity` option it fallbacks into RSA. So I will change that, thanks! 

On the other hand, what is the data race that we prevent using the lock? I checked and I could not find any, apart from the `closed` boolean, which can be replaced by a uber `atomic.bool`.

vyzo:
Or a CAS with sync/atomic; care for a patch?

aratz-lasa:
Okay, patch done. I used the uber package since internally makes use of `uint32` and the `sync.atomic` operations

PR #495, opened 2022-07-21
Replaced topic lock by atomic boolean to avoid lock contention

aratz-lasa:
PR opened in response to #494.

In the `topic` struct, there is currently lock contention due to the internal lock that is used to check if the topic has been closed or not. This generates lock contention, in those cases where the `publish` call takes a long time, such as if messages are signed with RSA.

Therefore, in this PR is proposed to replace the lock with an atomic boolean, allowing concurrent calls to Publish, and relieving lock contention.

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -11,6 +11,7 @@ import (
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 
 	"github.com/libp2p/go-libp2p-core/peer"
+	"go.uber.org/atomic"
 )
 
 // ErrTopicClosed is returned if a Topic is utilized after it has been closed
@@ -30,8 +31,7 @@ type Topic struct {
 	evtHandlerMux sync.RWMutex
 	evtHandlers   map[*TopicEventHandler]struct{}
 
-	mux    sync.RWMutex
-	closed bool
+	closed atomic.Bool
 }
 
 // String returns the topic associated with t
@@ -47,10 +47,7 @@ func (t *Topic) SetScoreParams(p *TopicScoreParams) error {
 		return fmt.Errorf("invalid topic score parameters: %w", err)
 	}
 
-	t.mux.Lock()
-	defer t.mux.Unlock()
-
-	if t.closed {
+	if t.closed.Load() {
 		return ErrTopicClosed
 	}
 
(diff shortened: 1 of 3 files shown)


topic.go
```diff
@@ -215,16 +206,14 @@ type ProvideKey func() (crypto.PrivKey, peer.ID)
 type PublishOptions struct {
 	ready     RouterReady
 	customKey ProvideKey
-	local bool
+	local     bool
```
vyzo, line 209: go fmt?

vyzo (commented):
It is not entirely the same, but i think this is fine.
I am a little surprised there is contention with the RW lock.

Can you gofmt?

topic.go
aratz-lasa: Yes

aratz-lasa:
I ran `gofmt` to all the source code and pushed again.

Lock contention occurs because the pubsub producers go faster than the time it takes to sign with Ed25519, therefore, they end up queueing up, since they cannot enter the publish call until the lock is free.

What do you mean by "it's not entirely the same"?

vyzo:
its an RLock, so they should not content, they can hold it concurrently.
Only close takes the exclusive lock.

It is not entirely the same because an active publish would prevent the topic from being closed concurrently.

vyzo:
Let me think about it a bit.

I dont think there will be any deleterious effects from the change, but it is still a change in behaviour.

aratz-lasa:
Aaah you are right! I thought that the `RLock` could only be acquired by goroutines in the same thread (since that's how it works in other languages). However I looked it up and I was wrong. There can be concurrent publish calls despite the goroutines being in different threads. So it's your call!

topic.go
```diff
@@ -347,9 +336,7 @@ func WithSecretKeyAndPeerId(key crypto.PrivKey, pid peer.ID) PubOpt {
 // Close closes down the topic. Will return an error unless there are no active event handlers or subscriptions.
 // Does not error if the topic is already closed.
 func (t *Topic) Close() error {
-	t.mux.Lock()
-	defer t.mux.Unlock()
-	if t.closed {
+	if t.closed.Load() {
```
SaveTheRbtz, line 339: There is a race here: you want to do `Swap` here and if old result was `false` then proceed with closing.

Otherwise 2 goroutines can hit Close, load false and proceed.

gammazero:
The RW lock appears to be the correct synchronization mechanism, and it should not be replaced with an `atomic.Bool`. Therefore, closing this PR.

PR #497, opened 2022-08-11
chore: ignore signing keys during WithLocalPublication publishing

vgonkivs:
We can ignore signing keys as there is no sense to verify ourselves during local publication.


Wondertan:
@vyzo, what do you think of this change?

vyzo (approved):
i think its fine.

PR #498, opened 2022-08-19
update go-libp2p to v0.22.0

marten-seemann:
Depends on https://github.com/libp2p/go-libp2p-testing/pull/66.


vyzo (commented):
tests are failing...

marten-seemann:
@vyzo Can you have a look please? I'm not familiar with this code base, might be probably a problem in test case itself.

vyzo:
it panics with a stream reset, and this used to work fine before -- so some regression in libp2p?

vyzo:
note that the tests use blankhost, maybe something broke there?

MarcoPolo:
Fix here: https://github.com/libp2p/go-libp2p/pull/1720

I don't think this is a regression. The issue was that connmanager was trimming connections. I'm actually surprised this test ever passed.

The fix is to use a mock clock for both the decayer and the connmanager. Otherwise the connmanager cleanup goroutine will kick before we expect it to.

I think we could backport https://github.com/libp2p/go-libp2p/pull/1720 into v0.22.1 and use that here or we can disable this test with a TODO in our release checklist for v0.23 to re-enable it along with some context.

vyzo:
I am fine with temporarily disabling this.

vyzo (approved):

PR #499, opened 2022-09-07
Enables non-atomic validation for peer scoring parameters

thep2p:
In the context of peer scoring, looking at [how the parameters and thresholds are currently set](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L270) at GossipSub, leaving some parameters unset, yields an error upon initialization of the GossipSub. For some applications and experimentations, we need to selectively set some of the parameters, e.g., only enabling staked participation in a permission setup. However, individual configuration of the scoring parameters is not feasible at the current state of GossipSub without setting all parameters and threshold values at their valid range. Missing even one parameter with a valid configuration fails the GossipSub instantiation.

This PR enables _non-atomic_ validation mode for the parameters and thresholds related to the peer scoring by adding a `SkipAtomicValidation` member field to the parameters and thresholds struct. By default, this variable is evaluated as `false`, i.e., zero value for `bool` data type. `!SkipAtomicValidation` indicates an atomic mode for validating thresholds and parameters which means all values must be set at their valid ranges. This is synonymous to how the validation is currently done at GossipSub. On the other hand, `SkipAtomicValidation` indicates a non-atomic mode for validating thresholds and parameters, which relaxes the validation criteria by allowing the application developer to selectively set _some of_ the parameters and threshold values. 

In contrast to the atomic validation mode which imposes all parameters and thresholds having a valid value, in the non-atomic mode, the parameters and thresholds can be left at their default (i.e., Golang zero) values as long as either all or none of the parameters related to the same context are set (e.g., mesh failure).

### DIFF
--- a/score_params.go
+++ b/score_params.go
@@ -183,41 +183,55 @@ func (p *PeerScoreParams) validate() error {
 		}
 	}
 
-	// check that the topic score is 0 or something positive
-	if p.TopicScoreCap < 0 || isInvalidNumber(p.TopicScoreCap) {
-		return fmt.Errorf("invalid topic score cap; must be positive (or 0 for no cap) and a valid number")
+	if !p.SkipAtomicValidation || p.TopicScoreCap != 0 {
+		// check that the topic score is 0 or something positive
+		if p.TopicScoreCap < 0 || isInvalidNumber(p.TopicScoreCap) {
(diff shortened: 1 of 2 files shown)


vyzo (approved):
this is a pretty good solution i would day.

PR #500, opened 2022-09-11
perf: use msgio pooled buffers for received msgs

### LINKED ISSUE #350: Reduce memory usage for receive buffers
We don't need to keep around large receive buffers. Instead, we can use msgio's message reader to:

1. Read the message length with a small buffer.
2. Get a larger buffer from the buffer pool to read the rest of the message.
3. Put the larger buffer back into the pool.

See https://github.com/libp2p/go-libp2p-kad-dht/blob/c245d836ab82922314dcbc364b3763d114938aad/dht_net.go#L76-L116 for how we do this in the DHT.

buffers

Wondertan:
The issue mentions only receiving buffers, so this PR only changes those. However, there might be value in changing sending buffers as well. Happy to update this in a subsequent commit if needed.

Closes #350

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -6,14 +6,14 @@ import (
 	"io"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
+
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
-
-	pb "github.com/libp2p/go-libp2p-pubsub/pb"
-
+	"github.com/libp2p/go-msgio"
 	"github.com/libp2p/go-msgio/protoio"
 
-	"github.com/gogo/protobuf/proto"
+	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 )
 
 // get the initial RPC containing all of our subscriptions to send to new peers
@@ -60,11 +60,11 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 		p.inboundStreamsMx.Unlock()
 	}()
 
-	r := protoio.NewDelimitedReader(s, p.maxMessageSize)
+	r := msgio.NewVarintReaderSize(s, p.maxMessageSize)
 	for {
-		rpc := new(RPC)
-		err := r.ReadMsg(&rpc.RPC)
+		msgbytes, err := r.ReadMsg()
 		if err != nil {
+			r.ReleaseMsg(msgbytes)
 			if err != io.EOF {
 				s.Reset()
 				log.Debugf("error reading rpc from %s: %s", s.Conn().RemotePeer(), err)
@@ -74,6 +74,15 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 				s.Close()
 			}
 
+			return
+		}
+
+		rpc := new(RPC)
+		err = rpc.Unmarshal(msgbytes)
+		r.ReleaseMsg(msgbytes)
+		if err != nil {
+			s.Reset()
+			log.Warnf("bogus rpc from %s: %s", s.Conn().RemotePeer(), err)
 			return
 		}
 


Wondertan:
@vyzo, any interest in this?

vyzo:
ah sorry, didnt notice it.
Can you tag me for review in future prs?

comm.go
```diff
@@ -77,9 +77,18 @@ func (p *PubSub) handleNewStream(s network.Stream) {
 			return
 		}
 
+		var rpc RPC
```
vyzo: a bit concerned about this being allocated on the stack, but i dont think the go compiler is that stupid.

comm.go
```diff
@@ -77,9 +77,18 @@ func (p *PubSub) handleNewStream(s network.Stream) {
...
+		if err != nil {
+			s.Reset()
+			log.Warnf("bogus rpc from %s: %s", s.Conn().RemotePeer(), err)
+			return
+		}
+
 		rpc.from = peer
 		select {
-		case p.incoming <- rpc:
+		case p.incoming <- &rpc:
```
vyzo: concern is this escape here.

vyzo (commented):
Thank you for this, it looks like it's going to help quite a bit with memory consumption.

Lets discuss a bit the rpc thingie, and then I am happy to merge.

vyzo:
@Wondertan ping

vyzo:
Also, lets do send buffers while at it; it can be a separate pr.

comm.go
Wondertan: Allocating this on the stack from my side was stupid... 
Making it like it was in the past `new(RPC)` to allocate on the heap directly

Wondertan:
> ah sorry, didnt notice it.

nw, someone has to make FVM sound 😉

Wondertan:
> ping

@vyzo, done

> Also, lets do send buffers while at it; it can be a separate pr.

👌🏻

comm.go
Wondertan: It is not that stupid, actually. `go build -gcflags="-m"` showed no escape. Still, better to keep it clear via `new`

Wondertan:
Also, updated `handlePeerEOF` to use msgio

comm.go
```diff
@@ -131,12 +140,11 @@ func (p *PubSub) handleNewPeerWithBackoff(ctx context.Context, pid peer.ID, back
...
 
-func (p *PubSub) handlePeerEOF(ctx context.Context, s network.Stream) {
+func (p *PubSub) handlePeerEOF(_ context.Context, s network.Stream) {
 	pid := s.Conn().RemotePeer()
-	r := protoio.NewDelimitedReader(s, p.maxMessageSize)
-	rpc := new(RPC)
+	r := msgio.NewVarintReaderSize(s, p.maxMessageSize)
 	for {
-		err := r.ReadMsg(&rpc.RPC)
+		_, err := r.ReadMsg()
```
vyzo: we are leaking from the pool here, lets immediately release.

vyzo (commented):
need to release the buffer in handleEOF.

comm.go
vyzo: It might be even better to not use a buffer at all, just read the length and skip, but non normative.

vyzo: Also note that it is probably a bug (or worse, a malicious peer?) in the other side if they write in their input stream.

I think we should not use a buffer, just read a varint, and if something comes in just kill the peer.

But lets not do this in this pr, we might want to discuss this in an issue and also bring it up in specs as the dual stream requirement recently came up.

Wondertan: Agree that we should not use buffer or release it here.

Wondertan: > But lets not do this in this pr, we might want to discuss this in an issue and also bring it up in specs as the dual stream requirement recently came up.

The dual stream is definitely something that needs to be "speced out" and I saw a PR from ChainSafe guy to fix this. 

Also, do we need to spec out the way we handle dead peers? It seems like this can vary depending on the implementation, in Go pubsub we only check if the peer is dead by awaiting `EOF` on the stream we write to, but in other implementations, there might be another way to do so. 

Or do you say that specifically, the fact of _reading_ out of _write stream_ has to be handled in a special way and this needs to be discussed and put into specs? Should we make an issue in this repo first?

And another question. Should I just remove/revert the changes from `handleEOF` from the PR so we can merge it?

Wondertan: cc @vyzo

vyzo:
@Wondertan ping -- lets move this forward?

Wondertan:
@vyzo, I am currently on the two-week team onsite, and it's almost the end. Going to bring this over the finish line right after.

The are some unanswered questions in the thread, IIRC

comm.go
vyzo: reverting the handleEOF change seems like the easiest way to merge for now.

However, we will need a follow up to improve the handling so that we cant possibly allocate a 1M buffer per peer unnecessarily.

Lets open an issue to discuss, and involve other people as well, and then we can escalate to spec once we have a reasonable implementation.

vyzo:
yeah, whenever you can.

I think reverting the eof change for now, or releasing the buffer, will do for merge.

But lets create a follow up issue to discuss the general handling and come up with a good solution that can later be speced for other impls to follow.

Wondertan:
@vyzo, so I reverted the EOF change and started making an issue for it. Now it's only the reading side with pooled buffers. I can also try to squeeze in the write today if we want to.

vyzo (approved):
LGTM for now, lets merge and do follow up prs.

Wondertan:
@vyzo, see https://github.com/libp2p/go-libp2p-pubsub/issues/506

PR #503, opened 2022-10-28
Enables injectable GossipSub router

thep2p:
This PR provides an alternative constructor for `GossipSub`, which is called `NewGossipSubWithRouter`. In contrast to the existing `NewGossipSub` constructor that builds a `GossipSub` instance with the default router, the `NewGossipSubWithRouter` alternative is capable of building a `GossipSub` instance with arbitrary `PubsubRouter`. This empowers building more application-specific routers and injecting them into `GossipSub`, e.g., the router with an application-specific metric collection.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1917,7 +1917,11 @@ func (gs *GossipSubRouter) getPeers(topic string, count int, filter func(peer.ID
 	return peers
 }
 
-func (gs *GossipSubRouter) WithTagTracerPubsubOption() Option {
+// WithDefaultTagTracer returns the tag tracer of the GossipSubRouter as a PubSub option.
+// This is useful for cases where the GossipSubRouter is instantiated externally, and is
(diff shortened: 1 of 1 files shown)


gossipsub.go
```diff
@@ -1909,6 +1917,10 @@ func (gs *GossipSubRouter) getPeers(topic string, count int, filter func(peer.ID
 	return peers
 }
 
+func (gs *GossipSubRouter) WithTagTracerPubsubOption() Option {
```
vyzo: document this please!
Also, lets call it `WithDefaultTagTracer`.

vyzo (commented):
LGTM, just a naming/documentation request.

gossipsub.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/503/commits/03b0d805018ea5e406a157eda8d708cce8b65b14

vyzo (approved):
thank you!

PR #505, opened 2022-11-08
Abstracts GossipSubRouter interface

thep2p:
## Problem Statement

Although in https://github.com/libp2p/go-libp2p-pubsub/pull/503 we introduced injectible `PubSubRouter` into the `GossipSub`, there are several places in the code that initializing a `GossipSub` instance with a customized `PubSubRouter` is failing due to the strict type assertions, e.g., [example-1](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L326), [example-2](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L360), and [example-3](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L387). 

Although the `rt` (router) attribute of the `GossipSub` is [deemed as a `PubSubRouter` interface type](https://github.com/libp2p/go-libp2p-pubsub/blob/1e161006c43f20435c2735bf24560f4d2d7eb429/pubsub.go#L54), nevertheless through the initialization phase of the `GossipSub` the `rt` attribute is strictly asserted as a `GossipSubRouter` value type. This is a blocker for initializing `GossipSub` with customized wrappers around `GossipSubRouter` as it no longer is going to be a `GossipSubRouter` strict value type. Hence the initialization fails with an `pubsub router is not gossipsub` error. 

## Contributions
In order to resolve the initialization failure due to the strict type assertions with `GossipSubRouter` this PR introduces the following changes:
- We define the `GossipPubSubRouter` interface type, and replaced strict type assertions with interface type assertions. 
- We decouple the option functions that deal with `GossipSubRouter` `struct` type from those deal with the `PubSub`.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -224,7 +224,7 @@ type GossipSubRouterOption func(*GossipSubRouter) error
 // DefaultGossipSubRouter returns a new GossipSubRouter with default parameters.
 func DefaultGossipSubRouter(h host.Host, opts ...GossipSubRouterOption) (*GossipSubRouter, error) {
 	params := DefaultGossipSubParams()
-	rt := &GossipSubRouter{
+	return &GossipSubRouter{
 		peers:     make(map[peer.ID]protocol.ID),
 		mesh:      make(map[string]map[peer.ID]struct{}),
 		fanout:    make(map[string]map[peer.ID]struct{}),
@@ -1927,10 +1927,6 @@ func (gs *GossipSubRouter) getPeers(topic string, count int, filter func(peer.ID
 	return peers
 }
 
-func (gs *GossipSubRouter) WithTagTracerPubsubOption() Option {
-	return WithRawTracer(gs.tagTracer)
-}
-
 func (gs *GossipSubRouter) SetPeerScore(score *PeerScore) {
 	gs.score = score
 }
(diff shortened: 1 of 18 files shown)


gossipsub.go
```diff
@@ -206,7 +206,10 @@ type GossipSubParams struct {
 
 // NewGossipSub returns a new PubSub object using the default GossipSubRouter as the router.
 func NewGossipSub(ctx context.Context, h host.Host, opts ...Option) (*PubSub, error) {
-	rt := DefaultGossipSubRouter(h)
+	rt, err := DefaultGossipSubRouter(h)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create default gossipsub router: %w", err)
+	}
 	opts = append(opts, WithRawTracer(rt.tagTracer))
```
vyzo, line 213: We need to find a way to absorb this (the tag tracer) into the defaults, it is very error prone to require it to be explicitly instantaited.

gossipsub.go
```diff
@@ -294,21 +307,17 @@ func WithPeerScore(params *PeerScoreParams, thresholds *PeerScoreThresholds) Opt
...
-		gs.gossipThreshold = thresholds.GossipThreshold
-		gs.publishThreshold = thresholds.PublishThreshold
-		gs.graylistThreshold = thresholds.GraylistThreshold
-		gs.acceptPXThreshold = thresholds.AcceptPXThreshold
-		gs.opportunisticGraftThreshold = thresholds.OpportunisticGraftThreshold
+		gs.SetPeerScore(newPeerScore(params))
+		gs.SetPeerScoreThresholds(thresholds)
 
-		gs.gossipTracer = newGossipTracer()
+		gs.SetGossipTracer(newGossipTracer())
```
vyzo, line 313: this very ugly.

vyzo (commented):
Well, day of reckoning.... sigh.

I am really not fond of reifying and exposing the internal types like that, but i can accept it if there is no other way around it.

Are you sure this isnt the case?
Cant we just reify the router type and offer an interface you can use, without changing the internal plumbing?

These abstractions are not public proof and thre is cost associated with having to go through the interface potentially (hundreds of) thousands times per second.

thep2p:
Thanks @vyzo, closing this PR in the favor of https://github.com/libp2p/go-libp2p-pubsub/pull/509

PR #507, opened 2022-11-19
perf: use msgio pooled buffers for message writes

Wondertan:
Continuation of #500

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -1,12 +1,14 @@
 package pubsub
 
 import (
-	"bufio"
 	"context"
+	"encoding/binary"
 	"io"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	pool "github.com/libp2p/go-buffer-pool"
+	"github.com/multiformats/go-varint"
 
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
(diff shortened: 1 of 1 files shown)


comm.go
```diff
@@ -156,16 +158,20 @@ func (p *PubSub) handlePeerEOF(ctx context.Context, s network.Stream) {
...
+
+		n := binary.PutUvarint(buf, size)
+		_, err := rpc.MarshalTo(buf[n:])
 		if err != nil {
 			return err
 		}
 
-		return bufw.Flush()
+		_, err = s.Write(buf)
+		return err
```
Wondertan, line 174: Notes for @vyzo
* I assume the bufio was used as protois's Writer could do two writes, and we wanted to avoid that with a single Flush.
	* Now we do only one write, so I removed the buffer
	* Pls correct me, if my assumption is wrong.
* I decided not to reuse the msgio Writer as that would require getting buffer from the pool twice
	* To marshal proto first
	* And then inside the Writer's `WriteMsg`
	* With this, less abstracted way, we do it only once
* The writeRpc logic is mostly copied from  https://github.com/celestiaorg/go-libp2p-messenger. It's [serde](https://github.com/celestiaorg/go-libp2p-messenger/blob/main/serde/serde.go#L74) pkg provides similar functionality to msgio, but it abstracts the message, not the Reader. Subsequently, it doesn't have an issue from the point above and provides a simpler API, IMO. It uses the go-buffer-pool. Just sharing.

vyzo: yup, makes sense.

vyzo (approved):
thank you!

PR #509, opened 2022-11-24
Adds Application Specific RPC Inspector

thep2p:
This PR implements an application-specific RPC inspector for `GossipSubRouter`. This is an _optional_ and _auxilary_ call-back method that can be set by the application so that all _accepted_ RPCs are inspected and only approved ones resume processing. 

When there is no inspected set, handling an accepted RPC goes as normal. This app-specific RPC inspector allows the application developer to: 
- Collect RPC statistics for application-level decision makings. 
- Implement application-level BFT mitigations at a lower level. 
- Implement application-specific white box tests involving the GossipSub.


vyzo (commented):
Much better!

Lets move this to the base pubsub object, it is not limited to gossipsub!

thep2p:
> Much better!
> 
> Lets move this to the base pubsub object, it is not limited to gossipsub!

@vyzo applied

pubsub.go
```diff
@@ -170,6 +170,11 @@ type PubSub struct {
 	protoMatchFunc ProtocolMatchFn
 
 	ctx context.Context
+
+	// appSpecificRpcInspector is an auxiliary that may be set by the application to inspect incoming RPCs prior to
+	// processing them. The inspector is invoked on an accepted RPC right prior to handling it.
+	// The return value of the inspector function is a boolean indicating whether the RPC should be processed or not.
+	appSpecificRpcInspector func(peer.ID, *RPC) bool
```
vyzo: lets return a loggable error to provide some context.

pubsub.go
```diff
@@ -527,6 +532,13 @@ func WithSeenMessagesTTL(ttl time.Duration) Option {
 	}
 }
 
+func WithAppSpecificRpcInspector(inspector func(peer.ID, *RPC) bool) Option {
```
vyzo: Also, lets make a type alias for the user inspector fun, returning error.

pubsub.go
```diff
@@ -1005,6 +1017,14 @@ func (p *PubSub) notifyLeave(topic string, pid peer.ID) {
 }
 
 func (p *PubSub) handleIncomingRPC(rpc *RPC) {
+	// pass the rpc through app specific validation (if any available).
+	if p.appSpecificRpcInspector != nil {
+		// check if the RPC is allowed by the external inspector
+		if accept := p.appSpecificRpcInspector(rpc.from, rpc); !accept {
```
vyzo:
```
If err := ....; err != nil {
  Log.Debugf(...)
  return
}
```

vyzo (commented):
lets fix the func signature and also type alias it, other than that LGTM.

pubsub.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/509/commits/776d8592b871718e6f63228f5ce8f9c919f941ea

pubsub.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/509/commits/776d8592b871718e6f63228f5ce8f9c919f941ea

pubsub.go
thep2p: https://github.com/libp2p/go-libp2p-pubsub/pull/509/commits/776d8592b871718e6f63228f5ce8f9c919f941ea

vyzo (approved):
This is great, thank you.

pubsub.go
```diff
@@ -527,6 +533,13 @@ func WithSeenMessagesTTL(ttl time.Duration) Option {
 	}
 }
 
+func WithAppSpecificRpcInspector(inspector func(peer.ID, *RPC) error) Option {
```
lthibault, line 536: I know this is already merged, but we probably want a docstring on this.  Just wanted to flag this 🙂 

cc @yhassanzadeh13

vyzo: Good point.

I am ready to cut a release,but i'll wait for a docstring here.

lthibault: 👍 PR incoming.

PR #513, opened 2023-01-06
feat: expire messages from the cache based on last seen time

### LINKED ISSUE #502: seenMessages TTL should be based on last observed time, not first
The [seenMessage](https://github.com/libp2p/go-libp2p-pubsub/blob/01ab84a809990a5ceefd34bca98139617fcc5389/pubsub.go#L152) cache prevents us from processing and re-broadcasting a pubsub message that we have already seen within the `seenMsgTTL` (default 2 minutes).  Accessing an existing message, however, [does not update the timestamp associated with the entry in the cache](https://github.com/whyrusleeping/timecache/blob/cfcb2f1abfee846c430233aef0b630a946e0a5a6/timecache.go#L57-L60).

This means that if a single pubsub message is being received frequently and consistently, we'll re-broadcast it every `seenMsgTTL`, rather than never re-broadcasting it as is the desired behavior.  If every time we received a duplicate pubsub message we pushed back the expiration time in the cache, that would give us the desired behavior of never re-broadcasting a message that we have already seen within the last `seenMsgTTL` seconds.

smrz2001:
This PR attempts to fix https://github.com/libp2p/go-libp2p-pubsub/issues/502.

It also includes a thread-safe copy of the original time cache [implementation](https://github.com/whyrusleeping/timecache).

Both the "first seen" (original) and "last seen" (new) implementations have been abstracted behind the `TimeCache` interface in a new `timecache` package.

cc @vyzo

### DIFF
--- a/blacklist.go
+++ b/blacklist.go
@@ -1,11 +1,11 @@
 package pubsub
 
 import (
-	"sync"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/peer"
-	"github.com/whyrusleeping/timecache"
+
+	"github.com/libp2p/go-libp2p-pubsub/timecache"
 )
 
 // Blacklist is an interface for peer blacklisting.
@@ -34,8 +34,7 @@ func (b MapBlacklist) Contains(p peer.ID) bool {
 
 // TimeCachedBlacklist is a blacklist implementation using a time cache
 type TimeCachedBlacklist struct {
-	sync.RWMutex
-	tc *timecache.TimeCache
+	tc timecache.TimeCache
 }
 
 // NewTimeCachedBlacklist creates a new TimeCachedBlacklist with the given expiry duration
@@ -46,8 +45,6 @@ func NewTimeCachedBlacklist(expiry time.Duration) (Blacklist, error) {
 
 // Add returns a bool saying whether Add of peer was successful
 func (b *TimeCachedBlacklist) Add(p peer.ID) bool {
-	b.Lock()
-	defer b.Unlock()
 	s := p.String()
 	if b.tc.Has(s) {
 		return false
@@ -57,8 +54,5 @@ func (b *TimeCachedBlacklist) Add(p peer.ID) bool {
 }
 
 func (b *TimeCachedBlacklist) Contains(p peer.ID) bool {
-	b.RLock()
-	defer b.RUnlock()
-
 	return b.tc.Has(p.String())
 }
--- a/go.mod
+++ b/go.mod
@@ -4,13 +4,15 @@ go 1.17
 
 require (
 	github.com/benbjohnson/clock v1.3.0
+	github.com/emirpasic/gods v1.18.1
 	github.com/gogo/protobuf v1.3.2
 	github.com/ipfs/go-log v1.0.5
+	github.com/libp2p/go-buffer-pool v0.1.0
 	github.com/libp2p/go-libp2p v0.22.0
 	github.com/libp2p/go-libp2p-testing v0.12.0
 	github.com/libp2p/go-msgio v0.2.0
 	github.com/multiformats/go-multiaddr v0.6.0
-	github.com/whyrusleeping/timecache v0.0.0-20160911033111-cfcb2f1abfee
+	github.com/multiformats/go-varint v0.0.6
 )
 
 require (
@@ -36,7 +38,6 @@ require (
 	github.com/jbenet/go-temp-err-catcher v0.1.0 // indirect
 	github.com/klauspost/compress v1.15.1 // indirect
 	github.com/klauspost/cpuid/v2 v2.1.0 // indirect
-	github.com/libp2p/go-buffer-pool v0.1.0 // indirect
 	github.com/libp2p/go-flow-metrics v0.1.0 // indirect
 	github.com/libp2p/go-netroute v0.2.0 // indirect
 	github.com/libp2p/go-openssl v0.1.0 // indirect
@@ -62,7 +63,6 @@ require (
 	github.com/multiformats/go-multicodec v0.5.0 // indirect
 	github.com/multiformats/go-multihash v0.2.1 // indirect
 	github.com/multiformats/go-multistream v0.3.3 // indirect
-	github.com/multiformats/go-varint v0.0.6 // indirect
 	github.com/nxadm/tail v1.4.8 // indirect
 	github.com/onsi/ginkgo v1.16.5 // indirect
 	github.com/opencontainers/runtime-spec v1.0.2 // indirect
--- a/pubsub.go
+++ b/pubsub.go
@@ -33,7 +33,7 @@ var (
 
 	// TimeCacheStrategy specifies which type of lookup/cleanup strategy is used by the seen messages cache.
 	// Use WithSeenMessagesStrategy to configure this per pubsub instance, instead of overriding the global default.
-	TimeCacheStrategy = timecache.Strategy_LastSeen
+	TimeCacheStrategy = timecache.Strategy_FirstSeen
 
 	// ErrSubscriptionCancelled may be returned when a subscription Next() is called after the
 	// subscription has been cancelled.
--- a/pubsub.go
+++ b/pubsub.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"github.com/libp2p/go-libp2p-pubsub/timecache"
 
 	"github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/discovery"
@@ -20,7 +21,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/protocol"
 
 	logging "github.com/ipfs/go-log"
-	"github.com/whyrusleeping/timecache"
 )
 
 // DefaultMaximumMessageSize is 1mb.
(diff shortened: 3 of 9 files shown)


blacklist.go
```diff
@@ -34,8 +34,7 @@ func (b MapBlacklist) Contains(p peer.ID) bool {
 
 // TimeCachedBlacklist is a blacklist implementation using a time cache
 type TimeCachedBlacklist struct {
-	sync.RWMutex
-	tc *timecache.TimeCache
+	tc timecache.TimeCache
```
smrz2001, line 37: Since the ported time cache implementation is thread-safe, I thought it would be ok to remove the `RWMutex` from here.

smrz2001: There is a minor change in the atomicity of operations though. Before, the lock was held across the entire `TimeCachedBlacklist.Add()` call, and thus across both the `tc.Has()` and `tc.Add()` calls.

Now, the time cache lock will be held only across the individual `tc.Has()` and `tc.Add()` calls. I think that should be ok, but I can totally just revert this change.

go.mod
```diff
@@ -4,13 +4,15 @@ go 1.17
 
 require (
 	github.com/benbjohnson/clock v1.3.0
+	github.com/emirpasic/gods v1.18.1
```
smrz2001, line 7: I only added this package, the other changes are from `go mod tidy`.

pubsub.go
```diff
@@ -31,6 +31,10 @@ var (
 	// Use WithSeenMessagesTTL to configure this per pubsub instance, instead of overriding the global default.
 	TimeCacheDuration = 120 * time.Second
 
+	// TimeCacheStrategy specifies which type of lookup/cleanup strategy is used by the seen messages cache.
+	// Use WithSeenMessagesStrategy to configure this per pubsub instance, instead of overriding the global default.
+	TimeCacheStrategy = timecache.Strategy_LastSeen
```
smrz2001: As you suggested, I've defaulted the strategy to be "last seen", i.e. the new implementation.

blacklist.go
smrz2001: One more thing - I retained the original (first-seen) time cache implementation here, but could also switch it to the new one.

vyzo:
thanks, i was on vacation. I will take a look soon.

timecache/first_seen_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+func (tc FirstSeenCache) Has(s string) bool {
+	tc.guard.RLock()
+	defer tc.guard.RUnlock()
+
+	ts, ok := tc.m[s]
+	// Only consider the entry found if it was present in the cache AND hadn't already expired.
+	return ok && time.Since(ts) <= tc.span
```
smrz2001, line 70: The expiration check was [absent](https://github.com/whyrusleeping/timecache/blob/master/timecache.go#L59) in the original implementation, which was a bug I [discovered](https://github.com/ipfs/kubo/pull/9372/files#diff-a75543978c37023721256291000855b02598e5d72668e203f7372cc4b8c159cfR222) while implementing the `SeenMessagesTTL` option in `kubo`.

I also added test cases for both cache implementations covering this scenario.

BigLep:
@vyzo : after this is reviewed/merged, can you cut a new release?   (I see this repo isn't part of unified ci: https://github.com/protocol/.github/blob/master/configs/go.json).  I'm asking so it can be bubbled up to Kubo for https://github.com/ipfs/kubo/pull/9372/files#diff-a75543978c37023721256291000855b02598e5d72668e203f7372cc4b8c159cfR222

vyzo:
sure

BigLep:
@vyzo : I know you're juggling a lot.  Would it be helpful if someone from the EngRes libp2p Stewards team takes a look?

vyzo:
It has been a little hectic in fvm land recently, but i do have some time to look at this during the weekend.

vyzo:
also, feel free to add steward reviewers, it would help disseminate knowledge.

pubsub.go
lidel: @smrz2001  bit confusing:  the old default is still present in 
https://github.com/libp2p/go-libp2p-pubsub/blob/04bfcf58514f59bddb650fec5b2edac018c8c406/timecache/time_cache.go#L19
and 
https://github.com/libp2p/go-libp2p-pubsub/blob/04bfcf58514f59bddb650fec5b2edac018c8c406/timecache/time_cache.go#L28-L30

lidel (dismissed):
@vyzo I eyeballed it, but any chance you take a look at this week? 

@smrz2001 quick ask:  unify default strategy (right now it is different https://github.com/libp2p/go-libp2p-pubsub/pull/513/files#r1071344615)

vyzo:
yes, yes, my weekend plan didnt materialize, but i have it on the list.

pubsub.go
smrz2001: Good point! Fixed.

blacklist.go
smrz2001: > One more thing - I retained the original (first-seen) time cache implementation here, but could also switch it to the new one.

Ok, scratch this. As per @lidel's comment, I aligned everything to use the new implementation.

smrz2001:
> @vyzo I eyeballed it, but any chance you take a look at this week?
> 
> @smrz2001 quick ask: unify default strategy (right now it is different https://github.com/libp2p/go-libp2p-pubsub/pull/513/files#r1071344615)

Fixed! Thanks, @lidel.

BigLep:
@MarcoPolo @marten-seemann @Wondertan : any chance you can please help take a look at this?  I'd like to get this over the line to get @smrz2001 and team unblocked.

Wondertan:
Happy to review this on the weekends if it will be relevant until then.

pubsub.go
vyzo: I would prefer to keep the existing behaviour as default.

I dont like surprising users with subtle changes, esp since this might effect memory usage.

Note that kubo is a fringe user, not surprising filecoin and eth is much more important here.

timecache/first_seen_cache.go
```diff
@@ -0,0 +1,70 @@
...
+func (tc FirstSeenCache) Add(s string) {
+	tc.guard.Lock()
+	defer tc.guard.Unlock()
+
+	_, ok := tc.m[s]
+	if ok {
+		panic("putting the same entry twice not supported")
+	}
+
+	tc.sweep()
```
vyzo, line 36: This should probably happen in the backround instead of every add, it is inefficient.

We dont have to do it in this pr, but if we dont we should add a TODO and open an issue for continuation.

timecache/last_seen_cache.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (tc *LastSeenCache) Add(s string) {
+	tc.guard.Lock()
+	defer tc.guard.Unlock()
+
+	tc.add(s)
+
+	// Garbage collect expired entries
+	tc.gc()
```
vyzo, line 44: same here.

timecache/last_seen_cache.go
```diff
@@ -0,0 +1,83 @@
...
+		// entries hereafter will be unexpired.
+		if time.Since(*ts) <= tc.span {
+			return
+		}
+		tc.m.Remove(key)
+	}
+}
+
+func (tc *LastSeenCache) Has(s string) bool {
+	tc.guard.Lock()
```
vyzo, line 73: RLock?

vyzo (commented):
A couple of issues to decide and a blocker: please dont change the default behaviour!

vyzo:
Please retain the default behaviour of first seen.

oed:
> Please retain the default behaviour of first seen.

@vyzo you said previously (https://github.com/libp2p/go-libp2p-pubsub/issues/502#issuecomment-1292586489) that it should be changed. This was the point of the bug that was reported.

vyzo:
Yes, i know, but i had achange of heart regarding the _default_ behaviour.

It is true that it is desirable to use the last seen behaviour for many applications.

However, some of the biggest users of this library are blockchains which can quite easily post reject and deduplicate messages once the timecache period has expired, because of consensus.
That is, after two minutes almost all blockchains can reject older messages reappearing and replays with their consensus based validators (nonce, block height, and so on).

For these applications, changing the default under their feet will not bring any tangible benefit for the increased memory usage.
At the very least i think we should be conservative and let the user choose.

Note that we can also consider making last seen the default a bit further down the line, say in a release or two.
Also, maturity of implementation concerns.

Does this make sense?

oed:
It does make sense. Thanks for explaining your reasoning @vyzo

timecache/first_seen_cache.go
smrz2001: What are your thoughts around where/how to do this? I could do it in this PR, if not complicated.

Or, if you would prefer to take some time and discuss the implementation, I could add a TODO for later.

If you meant in a goroutine with some kind of timed garbage collection, then would make sense to do in a separate PR.

timecache/last_seen_cache.go
smrz2001: Didn't do an `RLock` or even use a `RWMutex` (just using a plain `Mutex`) because the entry, if present and unexpired, would need to be "moved", which would need a write lock anyway.

Technically, I might have been able to do a `RLock`, checked for the entry and, if present, converted to `Lock` in order to move it.

From what I understand, "upgrading" locks like this without introducing race conditions is complicated and error-prone in Golang. Moreover, for an operation as small as delete + re-add (without GC), it felt like overkill.

This doesn't apply to the "first seen" cache implementation since it doesn't modify the cache when `Has` is called, unlike the "last seen" implementation.

timecache/first_seen_cache.go
vyzo: Lets add a TODO and do follow up, discussing in issue.

vyzo: and yes, a separate goroutine periodically running a sweep is what i have in mind.

timecache/last_seen_cache.go
vyzo: ok, good.

vyzo (approved):
ok, lets merge it and follow up for background gc.

vyzo:
@lidel you are blocking merge.

smrz2001:
> ok, lets merge it and follow up for background gc.

Great, thanks so much @vyzo and @lidel!

I opened #515 for the GC follow-up discussion.

PR #516, opened 2023-02-02
chore: update go-log dependency

richard-ramos:


go.mod
```diff
@@ -1,93 +1,90 @@
...
 	github.com/libp2p/go-openssl v0.1.0 // indirect
 	github.com/libp2p/go-reuseport v0.2.0 // indirect
-	github.com/libp2p/go-yamux/v3 v3.1.2 // indirect
-	github.com/lucas-clemente/quic-go v0.28.1 // indirect
-	github.com/marten-seemann/qtls-go1-16 v0.1.5 // indirect
-	github.com/marten-seemann/qtls-go1-17 v0.1.2 // indirect
-	github.com/marten-seemann/qtls-go1-18 v0.1.2 // indirect
-	github.com/marten-seemann/qtls-go1-19 v0.1.0 // indirect
+	github.com/libp2p/go-yamux/v4 v4.0.0 // indirect
+	github.com/lucas-clemente/quic-go v0.31.1 // indirect
```
chenrui333: why not just upgrade this into the latest version `quic-go/quic-go`, so that it can be built against go 1.20?

vyzo: it's implicit with go-libp2p, we never explicitly upgrade this.

vyzo (commented):
Uhm, in principle this is ok, but it will force us to cut a new minor release instead of patch, and also force downstream to upgrade libp2p, which is not necessarily desirable.

So lets punt a bit.

richard-ramos:
@vyzo I rebased against latest `master`, since https://github.com/libp2p/go-libp2p-pubsub/pull/517 got merged which did most of the changes that this PR originally proposed, so right now only go-log is pending for being updated. What do you think of merging it now?

vyzo:
sure

vyzo (approved):

PR #518, opened 2023-02-11
Expose Deep Copy of Mesh, Fanout, and Peers' Scores in GossipSub to Application Layer

thep2p:
This pull request implements a new feature in GossipSub that exposes a deep copy of the mesh, fanout, and peers' scores to the application layer. The deep-copy feature ensures that the internal state of GossipSub remains protected from interference by the application layer, while still providing the following benefits:

- Application-layer developers can perform comprehensive testing on the scoring mechanism of GossipSub. This includes setting scoring parameters, running test scenarios, and observing the updates to peers' scores in the GossipSub layer in real time.

- These metadata can be used in the application layer's decision-making process, such as allowing or disallowing protocol-level participation based on the behavior in the GossipSub protocol.

- Providing forensic insights into incidents and attacks.

By exposing these deep copies, the application layer can gain greater visibility into the inner workings of GossipSub, allowing for more robust testing and decision-making processes.


p-shahi:
Hi @yhassanzadeh13, thanks for this! ~~In the usual libp2p workflow (for maintainers and community members alike), new features are proposed as an update to the specifications ([gossipsub specs here](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub)) with an accompanying PR. Would you be able to create a specs PR outlining what you already included in the PR description here as well but in an md file. If you can provide some added motivation on whether/how you plan to use this in Flow would be really helpful as well.~~

On second thought, this change looks like it doesn't need a spec update.

vyzo:
This is perfectly fine and something we have been discussing.

Specs are irrelevant here, this is a library specific introspection mechanism and not a protocol change.

gossipsub.go
```diff
@@ -216,10 +218,88 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
 	return NewPubSub(ctx, h, rt, opts...)
 }
 
+// ReadOnlyRoutingMetadata is a read-only view of the routing metadataTracer maintained by the GossipSubRouter.
```
vyzo: maybe move this out of the way, in a separate file?

gossipsub.go
```diff
@@ -216,10 +218,88 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
...
+		r.fanout[topic] = make(map[peer.ID]struct{}, len(peers))
+		for p := range peers {
+			r.fanout[topic][p] = struct{}{}
+		}
+	}
+}
+
+type GossipSubRouterOption func(*GossipSubRouter)
+
+// WithReadOnlyRoutingMetadata returns a GossipSubRouterOption that enables the read-only routing metadataTracer.
```
vyzo: ditto

gossipsub.go
```diff
@@ -216,10 +218,88 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
...
+
+	for topic, peers := range fanout {
+		r.fanout[topic] = make(map[peer.ID]struct{}, len(peers))
+		for p := range peers {
+			r.fanout[topic][p] = struct{}{}
+		}
+	}
+}
+
+type GossipSubRouterOption func(*GossipSubRouter)
```
vyzo: the general pattern is to make a pubsub option that checks the router and asserts it to be gs.
Lets keep doing it that way.

gossipsub.go
```diff
@@ -216,10 +218,88 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
...
+// The metadataTracer is updated every k-heartbeats by the router.
+func WithReadOnlyRoutingMetadata(k uint64) GossipSubRouterOption {
+	return func(r *GossipSubRouter) {
+		r.metadataTracer = newReadOnlyRoutingMetadataTracer(k)
+	}
+}
+
 // DefaultGossipSubRouter returns a new GossipSubRouter with default parameters.
-func DefaultGossipSubRouter(h host.Host) *GossipSubRouter {
+func DefaultGossipSubRouter(h host.Host, opts ...GossipSubRouterOption) *GossipSubRouter {
```
vyzo: lets drop this, pubsub options are enough and there is no need to break the interface.

vyzo (commented):
I think we can do all this with a raw tracer instead of changing the router.

thep2p:
Reverting and closing this PR as there is seemingly a way to inject tracers for score and topology at the application layer. Thanks for the feedbacks!

PR #522, opened 2023-02-21
fix(timecache): remove panic in first seen cache on Add

Wondertan:
Closes #521

Issue #521, opened 2023-02-20
panic: putting same entry twice not supported


AskAlexSharov:
<img width="885" alt="Screenshot 2023-02-20 at 13 57 54" src="https://user-images.githubusercontent.com/46885206/220034833-78618440-2501-4088-82d5-e8b114f7c285.png">

vyzo:
Ouch, seems our newly incorporated cache has a bug.

Wondertan:
Our nodes are crashing because of this, and we need this fixed ASAP. If no one is working on this, we can take it,

vyzo:
sure, go ahead.

Correct behaviour is to ignore the message.

Wondertan:
The `Has` on cache is checked [here](https://github.com/libp2p/go-libp2p-pubsub/blob/cda3df70f42af335322cf4cf77f58fff84b32cff/pubsub.go#L998) with correct synchronization. However, false from `Has` [can be returned](https://github.com/libp2p/go-libp2p-pubsub/blob/cda3df70f42af335322cf4cf77f58fff84b32cff/timecache/first_seen_cache.go#L70) even if the message already in the cache, because the message is outdated(time since first seen > TTL)

Wondertan:
We can either: 
* Simply don't panic and ignore the message
* Or sweep the message in place on Has when the time since first seen > TTL
	* This way panic will never be triggered

Wondertan:
I am fine with both, but my intuition inclines toward the second option. WDYT, @vyzo

vyzo:
just ignore the message i would say, panic is very bad.

vyzo:
dont do the sweep in Has, will need xlock.

vyzo:
Also, we need to implement background sweeping at some point, for both cache impls, as this eager sweep business has implications for locking.

Wondertan:
There are many places where processLoop is locked unnecessarily :) 
Network.Connectedness can look it for a while

Wondertan:
> dont do the sweep in Has, will need xlock.

I know, but it seemed fine. Though background sweeping is a better approach so, let's just remove panic for now, as you've said

vyzo:
yeah, although we try to make it fast.

But point taken, ok, lets fix Has.

But please do remove the panic, it needs to ignore the message (and maybe log something in debug).

vyzo:
We are racing ;)

To summarize:
Lets fix Has while at it.
[...]

PR #523, opened 2023-02-21
Refactor timecache implementations

### LINKED ISSUE #515: fix: do time cache garbage collection in the background
Ref: https://github.com/libp2p/go-libp2p-pubsub/pull/513/files#r1084372794

vyzo:
Refactors the timechache implementations for a much simpler implementation and good behaviour.

In light of #521 and follow up to #522.
Closes #515

### DIFF
--- a/timecache/first_seen_cache.go
+++ b/timecache/first_seen_cache.go
@@ -1,72 +1,85 @@
 package timecache
 
 import (
-	"container/list"
+	"context"
 	"sync"
 	"time"
 )
 
-// FirstSeenCache is a thread-safe copy of https://github.com/whyrusleeping/timecache.
+var backgroundSweepInterval = time.Minute
+
+// FirstSeenCache is a time cache that only marks the expiry of a message when first added.
 type FirstSeenCache struct {
-	q     *list.List
-	m     map[string]time.Time
-	span  time.Duration
-	guard *sync.RWMutex
+	lk  sync.RWMutex
+	m   map[string]time.Time
+	ttl time.Duration
+
+	done func()
 }
 
-func newFirstSeenCache(span time.Duration) TimeCache {
-	return &FirstSeenCache{
-		q:     list.New(),
-		m:     make(map[string]time.Time),
-		span:  span,
-		guard: new(sync.RWMutex),
+var _ TimeCache = (*FirstSeenCache)(nil)
+
+func newFirstSeenCache(ttl time.Duration) *FirstSeenCache {
+	tc := &FirstSeenCache{
+		m:   make(map[string]time.Time),
+		ttl: ttl,
 	}
+
+	ctx, done := context.WithCancel(context.Background())
+	tc.done = done
+	go tc.background(ctx)
+
+	return tc
 }
 
-func (tc FirstSeenCache) Add(s string) {
-	tc.guard.Lock()
-	defer tc.guard.Unlock()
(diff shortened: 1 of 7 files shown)


timecache/first_seen_cache_test.go
```diff
@@ -17,22 +17,27 @@ func TestFirstSeenCacheFound(t *testing.T) {
...
+	backgroundSweepInterval = time.Second
+
 	tc := newFirstSeenCache(time.Second)
-	for i := 0; i < 11; i++ {
+	for i := 0; i < 10; i++ {
 		tc.Add(fmt.Sprint(i))
 		time.Sleep(time.Millisecond * 100)
 	}
 
+	time.Sleep(2 * time.Second)
```
MarcoPolo, line 28: Would be nice if we could use a mock clock here.

timecache/time_cache.go
```diff
@@ -16,23 +16,24 @@ const (
 )
 
 type TimeCache interface {
-	Add(string)
+	Add(string) bool
```
MarcoPolo, line 27: Can we document what this bool is?

MarcoPolo (approved):

timecache/time_cache.go
vyzo: sure, will do.

timecache/first_seen_cache_test.go
vyzo: yeah... future pr ;)

timecache/time_cache.go
vyzo: done.

timecache/time_cache.go
```diff
@@ -8,31 +8,45 @@ import (
...
+	// Depending on the implementation strategy, it may or may not update the expiry of
+	// an existing entry.
+	Add(string) bool
+	// Has checks the cache for the presence of an id.
+	// Depending on the implementation strategy, it may or may not update the expiry of
+	// an existing entry.
 	Has(string) bool
+	// Done signals that the user is done with this cache, which it may stop background threads
+	// and relinquish resources.
+	Done()
```
MarcoPolo, line 34: Is `.Close()` not the preferred convention? `.Done()` is used in `context.Context` which returns a channel that's closed when the context is finished. `.Done()` is also used in waitgroups to signal completion (but not cleanup).

`.Close()` seems closer to the intent. 

That said. This is minor, really just a nit

MarcoPolo (approved):

timecache/time_cache.go
vyzo: close returns error... there is none here, and there is no point.

vyzo: Initially i used Closer actually, but then i realize i am making work for myself with the error ;)

timecache/first_seen_cache.go
```diff
@@ -1,72 +1,56 @@
...
 }
 
-func (tc FirstSeenCache) Add(s string) {
-	tc.guard.Lock()
-	defer tc.guard.Unlock()
+var _ TimeCache = (*FirstSeenCache)(nil)
 
-	_, ok := tc.m[s]
-	if ok {
-		log.Debug("first-seen: got same entry")
```
Wondertan, line 32: Since we removed the log here, we can also remove the global log

vyzo: i guess... doesnt hurt all that much though.

Issue #524, opened 2023-02-25
Pubsub flood due to same message propagated multiple times


stbrody:
<h1>Incident report from 3Box Labs (Ceramic) Team </h1>

<h2>Incident summary</h2>
The Ceramic pubsub topic has been experiencing a flood of pubsub messages beyond our usual load for the last several days now.  We log every pubsub message we receive on the nodes that we run, and running analysis on those logs using LogInsights shows us that we are receiving messages with the exact same `seqno` multiple times - one message can show up upwards of 15 times in an hour.  During normal operation we do not see this issue with seqnos showing up multiple times.  This dramatic increase in the number of messages that need processing is causing excess load on our nodes that is causing major performance problems, even with as much caching and de-duplication as we can do at our layer.

<h2>Evidence of the issue</h2>

Graph of our incoming pubsub message activity showing how the number of messages spiked way up a few days ago.  The rate before 2/20 was our normal, expected amount of traffic:
<img width="1358" alt="Screen Shot 2023-02-25 at 1 47 14 PM" src="https://user-images.githubusercontent.com/870767/221379056-d1f8d222-c69c-403a-9a84-ef9beeb9b791.png">

AWS LogInsights Query demonstrating how the majority of this increased traffic is due to seeing the same message (with the same seqno) re-delivered multiple times.  Before the spike we never saw a `msg_count` greater than 2.

<img width="1026" alt="Screen Shot 2023-02-25 at 1 49 09 PM" src="https://user-images.githubusercontent.com/870767/221379083-607e8122-9625-499c-a499-c99034c9a130.png">
<img width="617" alt="Screen Shot 2023-02-25 at 1 49 27 PM" src="https://user-images.githubusercontent.com/870767/221379089-612e1702-ee89-4d7c-ab66-b1056cacf323.png">

<h2>Steps to reproduce</h2>
Connect to the gossipsub topic `/ceramic/mainnet`.  Observe the messages that come in, keep track of the number of times you see a message with each `seqno`.  You'll see that over the span of an hour you see the same message with the same `seqno` delivered multiple times

<h2>Historical context</h2>
[...]

PR #525, opened 2023-02-28
Default validator support

vyzo:
Adds support for default validators and provides an implementation that validates the message seqno as a per peer nocne to prevent infinite propagation when the network diameter exceeds the seen cache span.
See #524
Implements the logic described in https://github.com/ipfs/kubo/issues/9665

TBD:
- [x] unit test

### DIFF
--- a/validation_builtin.go
+++ b/validation_builtin.go
@@ -35,10 +35,11 @@ type BasicSeqnoValidator struct {
 }
 
 // NewBasicSeqnoValidator constructs a BasicSeqnoValidator using the givven PeerMetadataStore.
-func NewBasicSeqnoValidator(meta PeerMetadataStore) *BasicSeqnoValidator {
-	return &BasicSeqnoValidator{
+func NewBasicSeqnoValidator(meta PeerMetadataStore) ValidatorEx {
+	val := &BasicSeqnoValidator{
 		meta: meta,
 	}
+	return val.validate
 }
 
 func (v *BasicSeqnoValidator) validate(ctx context.Context, _ peer.ID, m *Message) ValidationResult {
--- a/validation_builtin.go
+++ b/validation_builtin.go
@@ -1,0 +1,93 @@
+package pubsub
+
+import (
+	"context"
+	"encoding/binary"
+	"sync"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+// PeerMetadataStore is an interface for storing and retrieving per peer metadata
+type PeerMetadataStore interface {
+	// Get retrieves the metadata associated with a peer;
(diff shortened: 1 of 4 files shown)


validation_builtin.go
```diff
@@ -0,0 +1,93 @@
...
+	Put(context.Context, peer.ID, []byte) error
+}
+
+// BasicSeqnoValidator is a basic validator, usable as a default validator, that ignores replayed
+// messages outside the seen cache window. The validator uses the message seqno as a peer-specific
+// nonce to decide whether the message should be propagated, comparing to the maximal nonce store
+// in the peer metadata store. This is useful to ensure that there can be no infinitely propagating
+// messages in the network regardless of the seen cache span and network diameter.
+// It requires that pubsub is instantiated with a strict message signing policy and that seqnos
+// are not disabled, ie it doesn't support anonymous mode.
```
MarcoPolo, line 26: Let's put a note that this currently breaks interop against rust nodes per: https://github.com/libp2p/rust-libp2p/issues/3453. 

I know the Rust impl is not spec compliant, but it would be nice to users to let them know about this rather than a painful debugging process.

MarcoPolo (approved):
One suggestion to the doc comment. Would be nice to have some tests here as well, but understood that this is urgent so we can defer to a future PR.

validation_builtin.go
vyzo: yeah, good point, I'll add a link to the issue with a warning.

MarcoPolo (approved):

validation_builtin.go
vyzo: done.

stbrody:
Is there a risk here if messages are delivered out of order? Like if a node receives message 5 before message 4 from the same peer, will that node then never forward message 4 on to its other peers? Does that risk some messages not being well propagated to the whole network?

vyzo:
This can only happen if one message from the peer is dropped (because of a full queue) and the next message propagates before the dropped message arrives from other paths.

So in theory there _is_ some risk, but this would be quite difficult to trigger.
Maybe with an extremely high sending  rate from some peer.

Worth noting, but i am not convinced it is a problem.

vyzo:
There is a simple way to alleviate concerns here: keep a bitvector (say 32 or 64 bits, as an integer) that encodes whether the last few messages have been delivered.

How does that sound?

stbrody:
@vyzo I don't think I fully understand how the bitvector would work, but I think keeping a small cache of the last few seqnos and whether they've been delivered once before does make sense and would address the issue I brought up!

vyzo:
Ok, i will implement something along these lines.

PR #536, opened 2023-07-06
use event bus for monitoring peer connections and protocol updates

jefft0:
This fixes a bug where another peer is not added to a topic after subscribing. This bug happens, for example, when a discovery system (like MDNS) is set up before pubsub initialization. Even though the peers have been discovered, they are not added to the new pubsub topics.

The bug can be seen by running [TestNotifyPeerProtocolsUpdated](https://github.com/jefft0/go-libp2p-pubsub/blob/15e9deb37922aa06420eb905fab1e898eefe90bc/notify_test.go#L11) in the master branch:

    git clone https://github.com/libp2p/go-libp2p-pubsub
    cd go-libp2p-pubsub
    curl https://raw.githubusercontent.com/jefft0/go-libp2p-pubsub/fix/notify-protocol-update/notify_test.go -o notify_test.go
    go test -run TestNotifyPeerProtocolsUpdated .

It prints "topic1 should at least have 1 peer". In our code we had to workaround this by disabling discovery (MDNS) and deferring discovery until after the pubsub protocols are registered. However, this pull request fixes the problem so that a workaround is not needed. It uses the event bus to monitor for peer connections and protocol updates. When `hosts[1]` joins `topic1`, the other peer is automatically added and the test passes.

### DIFF
--- a/notify.go
+++ b/notify.go
@@ -1,75 +1,96 @@
 package pubsub
 
 import (
+	"fmt"
+
+	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
-	ma "github.com/multiformats/go-multiaddr"
+	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 )
-
-var _ network.Notifiee = (*PubSubNotif)(nil)
 
 type PubSubNotif PubSub
 
-func (p *PubSubNotif) OpenedStream(n network.Network, s network.Stream) {
-}
-
-func (p *PubSubNotif) ClosedStream(n network.Network, s network.Stream) {
-}
-
-func (p *PubSubNotif) Connected(n network.Network, c network.Conn) {
-	// ignore transient connections
-	if c.Stat().Transient {
-		return
+func (p *PubSubNotif) startMonitoring() error {
+	sub, err := p.host.EventBus().Subscribe([]interface{}{
+		new(event.EvtPeerConnectednessChanged),
+		new(event.EvtPeerProtocolsUpdated),
+	}, eventbus.Name("pubsub/peers-notify")) // @NOTE(gfanton): is the naming is correct ?
+	if err != nil {
+		return fmt.Errorf("unable to subscribe to EventBus: %w", err)
 	}
+
+	// add current peers
+	p.addPeers(p.host.Network().Peers()...)
 
 	go func() {
-		p.newPeersPrioLk.RLock()
-		p.newPeersMx.Lock()
-		p.newPeersPend[c.RemotePeer()] = struct{}{}
-		p.newPeersMx.Unlock()
-		p.newPeersPrioLk.RUnlock()
+		defer sub.Close()
 
-		select {
-		case p.newPeers <- struct{}{}:
-		default:
-		}
-	}()
-}
+		for {
+			var e interface{}
+			select {
+			case <-p.ctx.Done():
+				return
+			case e = <-sub.Out():
+			}
 
-func (p *PubSubNotif) Disconnected(n network.Network, c network.Conn) {
-}
+			switch evt := e.(type) {
+			case event.EvtPeerConnectednessChanged:
+				// send record to connected peer only
+				if evt.Connectedness == network.Connected {
+					go p.addPeers(evt.Peer)
+				}
+			case event.EvtPeerProtocolsUpdated:
+				supportedProtocols := p.rt.Protocols()
 
-func (p *PubSubNotif) Listen(n network.Network, _ ma.Multiaddr) {
-}
-
-func (p *PubSubNotif) ListenClose(n network.Network, _ ma.Multiaddr) {
-}
-
-func (p *PubSubNotif) Initialize() {
-	isTransient := func(pid peer.ID) bool {
-		for _, c := range p.host.Network().ConnsToPeer(pid) {
-			if !c.Stat().Transient {
-				return false
+			protocol_loop:
+				for _, addedProtocol := range evt.Added {
+					for _, wantedProtocol := range supportedProtocols {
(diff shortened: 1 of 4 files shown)


jefft0:
Some background if it helps: We discovered and created a fix for this issue while working on this simple example of our networking library that is built on libp2p.
https://wesh.network/posts/share-contact-and-send-message

pubsub.go
```diff
@@ -327,13 +327,14 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
 		}
 	}
-	h.Network().Notify((*PubSubNotif)(ps))
 
 	ps.val.Start(ps)
 
 	go ps.processLoop(ctx)
 
-	(*PubSubNotif)(ps).Initialize()
+	if err := (*PubSubNotif)(ps).startMonitoring(); err != nil {
```
vyzo: Let's do this check before spawning the processLoop, as if it fails the system will not work at all.

vyzo (commented):
Thank you for this pr.

This is something that we have had to do for a while, so it is a step on the right direction.

At first glance it looks good, but I will do a second more thorough pass to make sure we don't break stuff.

pubsub.go
jefft0: Hello @vyzo . With the help of @gfanton , I added a second commit to address this comment. It more closely follows the logic which was there before.

jefft0: Thanks for the thumbs up. Do we need to address any other issues with this PR?

notify.go
```diff
@@ -1,75 +1,93 @@
...
-
-func (p *PubSubNotif) Connected(n network.Network, c network.Conn) {
-	// ignore transient connections
-	if c.Stat().Transient {
-		return
+func (p *PubSubNotif) startMonitoring() error {
+	sub, err := p.host.EventBus().Subscribe([]interface{}{
+		new(event.EvtPeerConnectednessChanged),
+		new(event.EvtPeerProtocolsUpdated),
+	}, eventbus.Name("pubsub/peers-notify")) // @NOTE(gfanton): is the naming is correct ?
```
vyzo: let's call it `libp2p/pubsub/notify` or something similar.

notify.go
```diff
@@ -1,75 +1,93 @@
...
 
-func (p *PubSubNotif) ListenClose(n network.Network, _ ma.Multiaddr) {
-}
-
-func (p *PubSubNotif) Initialize() {
-	isTransient := func(pid peer.ID) bool {
-		for _, c := range p.host.Network().ConnsToPeer(pid) {
-			if !c.Stat().Transient {
-				return false
-			}
```
vyzo, line 53: We want to keep this code around; please don't remove it and run it right after initializing the bus and before starting the monitoring goroutine.

notify.go
```diff
@@ -1,75 +1,93 @@
...
+			}
 
-		select {
-		case p.newPeers <- struct{}{}:
-		default:
+			switch evt := e.(type) {
+			case event.EvtPeerConnectednessChanged:
+				// send record to connected peer only
+				if evt.Connectedness == network.Connected {
+					go p.AddPeers(evt.Peer)
```
vyzo: we should process the disconnect events as well and get rid of the nasty code that handles current peer disconnections.

vyzo (changes requested):
This is a good start, left some comments requiring changes.

pubsub.go
```diff
@@ -327,13 +327,19 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
...
+		return nil, fmt.Errorf("unable to start pubsub monitoring: %w", err)
+	}
 
 	ps.val.Start(ps)
 
 	go ps.processLoop(ctx)
 
-	(*PubSubNotif)(ps).Initialize()
+	// add current peers to notify system
+	notify.AddPeers(h.Network().Peers()...)
```
vyzo, line 342: ah ok, it is still here.

notify.go
vyzo: So the initialization code is still there (below), but I believe it is racey.

pubsub.go
vyzo: This is racey; it needs to happen inside `startMonitoring`, after we have initialized the bus but _before_ we have spawned the monitoring goroutine.

notify.go
vyzo: See comments below.

notify.go
jefft0: Fixed the event bus name in the latest commit.

notify.go
jefft0: The latest commit handles the NotConnected event and calls RemovePeers. How does it look?

pubsub.go
jefft0: In the latest commit, the following test passes:

    go test -v -run=TestSimpleDiscovery -count=1 .

As you suggest, we could move `AddPeers` to inside `startMonitoring` before the goroutine as `p.AddPeers(p.host.Network().Peers()...)` . But with this change, `TestSimpleDiscovery` hangs and times out. What do you think?

jefft0:
Hello @vyzo . [You suggested](https://github.com/libp2p/go-libp2p-pubsub/pull/536#discussion_r1279074229) moving `AddPeers` to inside `startMonitoring` before the goroutine as `p.AddPeers(p.host.Network().Peers()...)` . But with this change, `TestSimpleDiscovery` hangs and times out. What do you suggest to do to resolve this? (Right now we have to use a fork of this repo with the fix.)

vyzo:
Uhm, fix the deadlock?

Maybe add an extra goroutine for startup.

Stebalien:
I re-implemented this in https://github.com/libp2p/go-libp2p-pubsub/pull/564 without even noticing that there was a PR in-progress. Oops...

But, that's merged now. So I've extracted the test to https://github.com/libp2p/go-libp2p-pubsub/pull/567.

PR #540, opened 2023-08-18
fix: topicscore params can't be set for dynamically subscribed topic

### LINKED ISSUE #533: `SetTopicScoreParams` after join a new pubsub topic
I noticed [here](https://github.com/libp2p/go-libp2p-pubsub/blob/master/score.go#L205-L210C2) that if a topic has not been setup before as part of the `WithPeerScore` option, then the `exist` variable will always be false, so it seems that it isn't it possible to set the `TopicScoreParams` with `SetTopicScoreParams` for new topics **after** GossipSub has been instantiated. 
```go
	old, exist := ps.params.Topics[topic]
	ps.params.Topics[topic] = p

	if !exist {
		return nil
	}
```
Is this a bug or intended behavior? because if remove the `if !exist {` condition, the logic seems to work fine when joining any new pubsub topic.

chaitanyaprem:
topicscore params can't be set for a topic subscribed after gossipsub is initialized.

Fixed #533

### DIFF
--- a/score.go
+++ b/score.go
@@ -202,7 +202,7 @@ func (ps *peerScore) SetTopicScoreParams(topic string, p *TopicScoreParams) erro
 	ps.Lock()
 	defer ps.Unlock()
 
-	old, _ := ps.params.Topics[topic]
+	old := ps.params.Topics[topic]
 	ps.params.Topics[topic] = p
 
 	// check to see if the counter Caps are being lowered; if that's the case we need to recap them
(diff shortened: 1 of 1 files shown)


score.go
```diff
@@ -202,13 +202,9 @@ func (ps *peerScore) SetTopicScoreParams(topic string, p *TopicScoreParams) erro
 	ps.Lock()
 	defer ps.Unlock()
 
-	old, exist := ps.params.Topics[topic]
+	old, _ := ps.params.Topics[topic]
```
vyzo: you can just remove the underscore completely.

vyzo (commented):
Thank you.

Just a small nit, otherwise looks good.

score.go
chaitanyaprem: Done

vyzo (approved):
thank you!

PR #541, opened 2023-08-26
Revert "fix: topicscore params can't be set for dynamically subscribed topic"

vyzo:
Reverts libp2p/go-libp2p-pubsub#540

This needs to be reverted; the check was necessary and not impeding the setting.
Without the check, if the topic score params did not previously exist, it will crash with nil pointer dereference.


vyzo:
cc @chaitanyaprem

MarcoPolo (approved):

PR #553, opened 2024-01-15
GossipSub v1.2: IDONTWANT control message and priority queue.

haxxpop:
Sending IDONTWANT
- [x] Implement a smart queue
- [x] Add priorities to the smart queue
- [x] Put IDONTWANT packets into the smart priority queue as soon as the node gets the packets

Handling IDONTWANT
- [x] Use a map to remember the message ids whose IDONTWANT packets have been received
- [x] Implement `max_idontwant_messages` (ignore the IDONWANT packets if the max is reached)
- [x] Clear the message IDs from the cache after 3 heartbeats
- [x] Hash the message IDs before putting them into the cache.

More requested features
- [x] Add a feature test to not send IDONTWANT if the other side doesnt support it

Spec: https://github.com/libp2p/specs/pull/548


comm.go
```diff
@@ -172,19 +172,15 @@ func (p *PubSub) handleSendingMessages(ctx context.Context, s network.Stream, ou
...
 	defer s.Close()
 	for {
-		select {
-		case rpc, ok := <-outgoing:
-			if !ok {
-				return
-			}
+		rpc, err := outgoing.Pop(ctx)
+		if err != nil {
+			return
```
vyzo, line 181: We might want to log the error (at very low priority) if it is not a closed indicator.

rpc_queue.go
```diff
@@ -0,0 +1,127 @@
...
+		default:
+		}
+		q.dataAvailable.Wait()
+		// It can receive a signal because the queue is closed.
+		if q.IsClosed() {
+			return nil, ErrQueueClosed
+		}
+	}
+	rpc := q.queue[0]
+	q.queue = q.queue[1:]
```
vyzo: this might leak space.

rpc_queue.go
```diff
@@ -0,0 +1,127 @@
...
+			q.spaceAvailable.Wait()
+			// It can receive a signal because the queue is closed.
+			if q.IsClosed() {
+				panic(ErrQueuePushOnClosed)
+			}
+		} else {
+			return ErrQueueFull
+		}
+	}
+	q.queue = append(q.queue, rpc)
```
vyzo: this might leak space.

vyzo (commented):
So far looking good overall.

However, I think using a slice to back the queue is problematic at many levels as it might leak space with a growing array, and also hang on to too much space.

I would very much prefer using a dequeue for this.

haxxpop:
> However, I think using a slice to back the queue is problematic at many levels as it might leak space with a growing array, and also hang on to too much space.

The unused space will be freed eventually https://stackoverflow.com/a/26863706

haxxpop:
I'm not really sure if we need dfd81e80848e5ce4e64fbf30cfdaafb4dc56b9d3 or not. I guess adding a non-standard flag to `pb/trace.proto` would break some users of the tracers?

vyzo:
> I'm not really sure if we need [dfd81e8](https://github.com/libp2p/go-libp2p-pubsub/commit/dfd81e80848e5ce4e64fbf30cfdaafb4dc56b9d3) or not. I guess adding a non-standard flag to `pb/trace.proto` would break some users of the tracers?

Yeah, let's not break compatibility with existing code, we don't need it.

Any progress other than that?

haxxpop:
> Any progress other than that?

I haven't implemented the second half. Does the first half looks good to you?

vyzo:
yes, looks reasonable so far.

haxxpop:
@vyzo I finished everything already. Sorry for the delay. I was in vacation and was a bit busy lately. Thank you for your patience.

gossipsub.go
```diff
@@ -56,7 +56,9 @@ var (
 	GossipSubGraftFloodThreshold              = 10 * time.Second
 	GossipSubMaxIHaveLength                   = 5000
 	GossipSubMaxIHaveMessages                 = 10
+	GossipSubMaxIDontWantMessages             = 10
```
vyzo: isnt that too low?

gossipsub.go
```diff
@@ -1617,6 +1693,13 @@ func (gs *GossipSubRouter) clearIHaveCounters() {
 	}
 }
 
+func (gs *GossipSubRouter) clearIDontWantCounters() {
```
vyzo, line 1794: Thats too aggressive and can clear the counters at unfortunate times.
Lets keep them for heartbeat or two, maybe 3 to match the cache sliding window.

vyzo (commented):
Generally looks good, but lets make the counter clear less aggressive.

gossipsub.go
haxxpop: Yeah, I didn't think enough about this. I just copied it from `GossipSubMaxIHaveMessages`. I think I will change it to 1,000 which will makes the IDONTWANT cache grows by 60,000 per minute (because `GossipSubHeartbeatInterval` is 1 second)

gossipsub.go
haxxpop: I don't understand this. When is the unfortunate time?

This only clear the counters not the cache that keeps the message ids from IDONTWANTs. So technically, the slower you clear the cache, the more you prevent the DoS attack. The faster you clear the cache, the more efficiency you have (because you process more IDONTWANTs).

vyzo: Lets say you just got an IDW, heartbeat clear, and then the message arrives... and you send it even though it is not wanted.

haxxpop: > and you send it even though it is not wanted.

This is not true.

The heartbeat clear only clears the counter, not the cache. The IDONTWANT is still in the cache so the message is not forwarded. In other words, the heartbeat clear clears `peerdontwant`, while only `unwanted` is used to decide whether to forward it or not.

vyzo: when is the cache cleared?

haxxpop: > when is the cache cleared?

Currently, it's not cleared at all. It's currently not specified in the spec (even if it probably should be).

As I read from the spec, `max_idontwant_messages` is only used for limiting the number of IDONTWANTs to be processed per heartbeat per peer.

If the intention is really to limit the number of IDONTWANTs the node should remember in the cache to save the memory, I think the current spec doesn't make any sense because you should do an LRU with a max size instead of doing the heartbeat thing.

vyzo: we need to clear the cache, otherwise it is compleltely broken.

vyzo: keep a counter of how many heartbeats each IDW is kept for, and clear it after 3.

haxxpop:
The message ids from IDONTWANT can be very large, so hashing it before keeping it in memory quite makes sense. The same thing is already implemented in nim-libp2p https://github.com/vacp2p/nim-libp2p/pull/1090

vyzo:
sure, but it needs to be cleared, otherwise its  memroy leak and dos waiting to happen. Or at best the feature breaks itself for long running nodes.

I suggest for each IDW keep a counter of how many heartbeats it survived, and after 3 (to match the IHAVE ads) it should be cleared.

gossipsub.go
haxxpop: > keep a counter of how many heartbeats each IDW is kept for, and clear it after 3.

That makes sense. I will do it, but I think it's considered go-libp2p specific and out of the scope of the spec, right?

vyzo: implementation detail perhaps.

haxxpop: Wait. If the purpose is to just limit the memory used, why not just define a parameter for the maximum number of IDONTWANTs to keep in the cache per peer?

If the number in the cache grows beyond the limit, just remove the least recently seen IDONTWANT from the cache before adding a new one.

I think this is simpler than dealing with the heartbeats.

vyzo: you misunderatand.

We need to keep the IDW for some heartbeats and then clear them.

It is both for controlling memory use and for correctness.

vyzo: reagardless, an LRA (least recently added) cache for IDW could work and keep memory bounded and it is possibly simpler.

However we need to consider cache sizes carefully.

vyzo: and the problem here is a sybil attack, and we get to keep all the garbage indefinitely.

So no, this is not the correct solution.

We need to explicitly clean old IDWs.

vyzo: In fact we should go one step further and use a global limit for IDWs across all peers in the cache.

haxxpop: > It is both for controlling memory use and for correctness.

Alright, got it. Thanks.

haxxpop:
@vyzo It's all done. Sorry for the delay.

vyzo:
ok, can you rebase on master?

chaitanyaprem:
@vyzo wondering if this PR ready to be merged or are you waiting for anything to be fixed/addressed? 

This seems to improve bandwidth usage and we want to take advantage of that.

vyzo (commented):
I think in general it looks good, but i do want to do another pass.

In the meantime:
- can you bump the version to 1.2
- can you add a feature test to not send IDONTWANT if the other side doesnt support it?

I will also invoke @Stebalien for review.

chaitanyaprem:
> I think in general it looks good, but i do want to do another pass.
> 
> In the meantime:
> 
>     * can you bump the version to 1.2
> 
>     * can you add a feature test to not send IDONTWANT if the other side doesnt support it?
> 
> 
> I will also invoke @Stebalien for review.

Haven't looked at the PR in detail, but will give it a try tomorrow.

rpc_queue.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (q *priorityQueue) Pop() *RPC {
+	var rpc *RPC
+
+	if len(q.priority) > 0 {
+		rpc = q.priority[0]
+		q.priority = q.priority[1:]
+	} else if len(q.normal) > 0 {
+		rpc = q.normal[0]
+		q.normal = q.normal[1:]
```
Stebalien:
```suggestion
		rpc = q.priority[0]
		q.normal[0] = nil
		q.priority = q.priority[1:]
	} else if len(q.normal) > 0 {
		rpc = q.normal[0]
		q.normal[0] = nil
		q.normal = q.normal[1:]
```

I'd also consider using ring-buffers to avoid constantly re-allocating/moving these slices, but that can be done later after some profiling.

rpc_queue.go
```diff
@@ -0,0 +1,170 @@
...
+
+type rpcQueue struct {
+	dataAvailable  *sync.Cond
+	spaceAvailable *sync.Cond
+	// Mutex used to access queue
+	queueMu *sync.Mutex
+	queue   *priorityQueue
+
+	// RWMutex used to access closed
+	closedMu *sync.RWMutex
```
Stebalien: The pointers are unnecessary.

```go
q := &rpcQueue{maxSize: maxSize}
q.dataAvailable.L = &q.queueMu
q.spaceAvailable.L = &q.queueMu
```

rpc_queue.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}()
+
+	defer func() {
+		// Tell the other routine that this function is finished.
+		select {
+		case finished <- struct{}{}:
+		default:
+		}
+	}()
```
Stebalien: Use `context.AfterFunc`.

```suggestion
	defer context.AfterFunc(ctx, func() { q.dataAvailable.Broadcast() })()
```

rpc_queue.go
```diff
@@ -0,0 +1,170 @@
...
+	return rpc, nil
+}
+
+func (q *rpcQueue) Close() {
+	q.closedMu.Lock()
+	q.closed = true
+	q.closedMu.Unlock()
+
+	q.dataAvailable.Broadcast()
+	q.spaceAvailable.Broadcast()
```
Stebalien: This races with the `IsClosed` checks. I'd just use a single mutex (the queue mutex). You're not getting anything out of the read-write mutex anyways because all interesting functions take the queue mutex.

rpc_queue.go
Stebalien: (assuming I'm not missing something important)

rpc_queue.go
Stebalien: Holding the lock while broadcasting may also fix it? But I'd just go with the one-lock solution, there's no need for any additional complexity.

comm.go
```diff
@@ -175,19 +175,15 @@ func (p *PubSub) handleSendingMessages(ctx context.Context, s network.Stream, ou
 
 	defer s.Close()
 	for {
```
Stebalien:
```suggestion
	for ctx.Err() == nil {
```

(existing bug, but the new queue implementation could exacerbate it).

gossipsub.go
```diff
@@ -917,6 +967,27 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
 	}
 }
 
+func (gs *GossipSubRouter) handleIDontWant(p peer.ID, ctl *pb.ControlMessage) {
+	if gs.unwanted[p] == nil {
+		gs.unwanted[p] = make(map[checksum]int)
+	}
+
+	// IDONTWANT flood protection
```
Stebalien, line 1003: We need global maximums as well, right? Or do we limit the number of grafted peers?

gossipsub.go
```diff
@@ -917,6 +967,27 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+	gs.peerdontwant[p]++
+	if gs.peerdontwant[p] > gs.params.MaxIDontWantMessages {
+		log.Debugf("IDONWANT: peer %s has advertised too many times (%d) within this heartbeat interval; ignoring", p, gs.peerdontwant[p])
+		return
+	}
+
+	// Remember all the unwanted message ids
+	for _, idontwant := range ctl.GetIdontwant() {
+		for _, mid := range idontwant.GetMessageIDs() {
+			hashed := sha256.Sum256([]byte(mid))
```
Stebalien: Can we assume that message IDs are small? Lotus, at least, uses hashes already.

We can also inline if and only if the ID is small enough:

```go
type id struct {
	payload [32]byte
	length uint8
}

func newId(mid string) id {
	var out id
	if len(id) > 32 || len(id) == 0 {
		out.payload = sha256.Sum256([]byte(mid))
	} else {
		out.length = uint8(copy(out.payload, mid))
	}
	return id
}
```

gossipsub.go
vyzo: that is limited by the mesh size, so we probably don't need a global maximum.

rpc_queue.go
vyzo: I agree that we would like a more efficient implementation that doesn't stress the garbage collector, but it gets a bit more complicated with the need of avoiding memory overcommit.

Specifically, we would like the queue to shrink if there is no buildup -- a ring buffer will not have that property and use a constant amount of memory on the queue size, which can add up with a lot of peers, most of whom are not likely to have a build up.

We discussed this with @ppopth above, and I think this is fine for now and we shouldn't block merging on this -- rather make a follow up issue to write a more optimize implementation that reduces memory consumption and gc pressure.

rpc_queue.go
vyzo: I agree that we would like a more efficient implementation that doesn't stress the garbage collector, but it gets a bit more complicated with the need of avoiding memory overcommit.

Specifically, we would like the queue to shrink if there is no buildup -- a ring buffer will not have that property and use a constant amount of memory on the queue size, which can add up with a lot of peers, most of whom are not likely to have a build up.

We discussed this with @ppopth above, and I think this is fine for now and we shouldn't block merging on this -- rather make a follow up issue to write a more optimize implementation that reduces memory consumption and gc pressure.

rpc_queue.go
vyzo: agreed.

gossipsub.go
vyzo: however, we should probably ignore IDONTWANT's from peers that are not in the mesh -- so there is something to fix here.

rpc_queue.go
ppopth: > (assuming I'm not missing something important)

`q.dataAvailable.Broadcast()` is still called after the function returns even if the contenxt is not done yet. So I don't think `Afterfunc` is the right solution.

haxxpop: > (assuming I'm not missing something important)

If you use `AfterFunc`, `q.dataAvailable.Broadcast()` will be called for sure, right? but it may not be called if `finished` is triggered (i.e. the function returns). So I don't think `AfterFunc` will work.

gossipsub.go
haxxpop: Alright, let's ignore IDONTWANT's from non-mesh peers.

rpc_queue.go
Stebalien: > but it may not be called if finished is triggered

It won't be called if the function returns, no. But that's by design, right? That's what the current code does and you only want to broadcast on cancel.

rpc_queue.go
```diff
@@ -0,0 +1,161 @@
...
+
+type rpcQueue struct {
+	dataAvailable  sync.Cond
+	spaceAvailable sync.Cond
+	// Mutex used to access queue
+	queueMu sync.Mutex
+	queue   priorityQueue
+
+	// RWMutex used to access closed
+	closedMu sync.RWMutex
```
Stebalien: removeme

rpc_queue.go
haxxpop: > That's what the current code does and you only want to broadcast on cancel.

Yes, it's by design. The point of broadcasting is to wake the `q.dataAvailable.Wait()` call up and let the `Pop` routine return.

If it doesn't wake up, the `Pop` routine will not return when the context is cancelled.

We don't want to wake it up, if the function returns normally and the context is not cancelled because the `Pop` routine already return.

gossipsub.go
haxxpop: That makes sense. Let's do that.

rpc_queue.go
Stebalien: Ok, that's what I thought. That line of code is short for:

```go
	unregisterAfterFunc := context.AfterFunc(ctx, func() { q.dataAvailable.Broadcast() })
	defer unregisterAfterFunc()
```

That is, the function will be unregistered on return.

haxxpop: Alright, that makes sense.

gossipsub.go
haxxpop: Wait. IDONTWANT is not topic-specific and there is no way to discover the topic by looking at the IDONTWANT messages, so I think it's better not to handle this.

vyzo:
seems like we have some flaky tests :(

vyzo:
@ppopth can you look at the go-check CI failure? Looks like the linter is complaining about some things.

haxxpop:
@vyzo could you rerun the CI? thanks

vyzo:
done

gossipsub.go
Stebalien: We don't have to make it topic specific, just only accept said messages from peers to which we are grafted (on any topic) and, ~ideally, limit the number of IDONTWANTs per peer~ *(which is what we're doing now)*.

Note: this is assuming that there's a limit to the number of "grafted peers". My goal is to prevent an attacker from just re-connecting with many different peer IDs to fill up these maps.

vyzo: there is and if it is exceeded it gets cleared on next heartbeat.

Stebalien: But I don't see any limits on the number of peers.

Also, we're currently dealing with an issue in F3 where we're getting absolutely _hammered_ by ~IHAVE~ IWANT messages on topics we're not subscribed to.

vyzo: yeah, maybe we should track the topic in the cache and avoid sending IHAVEs if the peer is not in the topic.

But thats a larger change, that will probably need a spec update, and certainly out of scope of this pr.

Do you want to open an issue about this, maybe in specs?

vyzo: as for the peers, the meshes are kept within degree bounds, maintained in the heartbeat.

So i am not convinced it is an effective  attack vector.

Lets open a follow up issue for this?

haxxpop:
@vyzo The CI error log is not informative. Do you have any idea?

```
...
--- PASS: TestTopicPublishWithKeyInvalidParameters (0.08s)
    --- PASS: TestTopicPublishWithKeyInvalidParameters/nil_sign_private_key_should_error (0.00s)
    --- PASS: TestTopicPublishWithKeyInvalidParameters/empty_peer_ID_should_error (0.00s)
=== RUN   FuzzAppendOrMergeRPC
--- PASS: FuzzAppendOrMergeRPC (0.00s)
FAIL
FAIL	github.com/libp2p/go-libp2p-pubsub	385.852s
-test.shuffle 1723651417916320350
=== RUN   TestLastSeenCacheSlideForward
    last_seen_cache_test.go:36: timing is too fine grained to run in CI
--- SKIP: TestLastSeenCacheSlideForward (0.00s)
=== RUN   TestFirstSeenCacheNotFoundAfterExpire
--- PASS: TestFirstSeenCacheNotFoundAfterExpire (2.00s)
=== RUN   TestFirstSeenCacheFound
--- PASS: TestFirstSeenCacheFound (0.00s)
=== RUN   TestLastSeenCacheExpire
--- PASS: TestLastSeenCacheExpire (3.10s)
=== RUN   TestFirstSeenCacheExpire
--- PASS: TestFirstSeenCacheExpire (3.00s)
=== RUN   TestLastSeenCacheFound
--- PASS: TestLastSeenCacheFound (0.00s)
=== RUN   TestLastSeenCacheNotFoundAfterExpire
--- PASS: TestLastSeenCacheNotFoundAfterExpire (2.00s)
PASS
ok  	github.com/libp2p/go-libp2p-pubsub/timecache	10.111s
FAIL
```

vyzo:
I think these are just flaky tests, i dont think your pr is breaking them.

gossipsub.go
haxxpop: > Also, we're currently dealing with an issue in F3 where we're getting absolutely _hammered_ by ~IHAVE~ IWANT messages on topics we're not subscribed to.

Could reference the issue on IWANT? or explain it in detail. I can't imagine how you can be dos'ed by receiving too many IWANTs (or IHAVEs).

haxxpop: > yeah, maybe we should track the topic in the cache and avoid sending IHAVEs if the peer is not in the topic.

I will be surprised if go-libp2p sends IHAVEs to some peers that are not in the topic. I think it's probably already in the spec and I always thought that no implementation sends IHAVEs to non-subscribing peers.

cortze: > I will be surprised if go-libp2p sends IHAVEs to some peers that are not in the topic. I think it's probably already in the spec and I always thought that no implementation sends IHAVEs to non-subscribing peers.

Doesn't the spec already contemplate sending messages on topics that it isn't subscribed to using the fanout peers?

I know that this is a common practice in Ethereum CL, and I don't see any kind of filter on the [GossipSub publishing side](https://github.com/libp2p/go-libp2p-pubsub/blob/19ffbb3a482caecabb8520917c631e3047a78094/gossipsub.go#L1037) that prevents adding a message to the cache of messages if we are not subscribed to the topic 🤷🏽

vyzo: I think it might make sense to specify the behavior.
It makes more sense to track the topic and only gossip IHAVEs to peers in the topic.

vyzo: At any rate, it is out of scope of this pr.

gossipsub.go
```diff
@@ -58,10 +65,15 @@ var (
 	GossipSubOpportunisticGraftPeers          = 2
 	GossipSubGraftFloodThreshold              = 10 * time.Second
 	GossipSubMaxIHaveLength                   = 5000
-	GossipSubMaxIHaveMessages                 = 10
+	GossipSubMaxIHaveMessages                 = 1000
```
vyzo: wait, why are we increasing the max IHAVE?

gossipsub.go
```diff
@@ -931,6 +990,26 @@ func (gs *GossipSubRouter) handlePrune(p peer.ID, ctl *pb.ControlMessage) {
...
+	if gs.unwanted[p] == nil {
+		gs.unwanted[p] = make(map[checksum]int)
+	}
+
+	// IDONTWANT flood protection
+	gs.peerdontwant[p]++
+	if gs.peerdontwant[p] > gs.params.MaxIDontWantMessages {
+		log.Debugf("IDONWANT: peer %s has advertised too many times (%d) within this heartbeat interval; ignoring", p, gs.peerdontwant[p])
+		return
+	}
```
vyzo: the check has to be prior to incrementing I think

vyzo (commented):
It generally looks good, left a couple of comments.
Once @Stebalien is happy as well, I'll tick and merge.

@ppopth can you disable the flaky test and open a follow up issue to fix it?

vyzo:
Regarding the test failures:
- TestGossipsubConnTagMessageDeliveries this seems consistently flaky; let's add a t.Skip and open follow up issue to fix it.
- FuzzAppendOrMergeRPC @ppopth this might need to be looked upon; if it is a flake, we can skip as well and open follow up issue, but might be a problem (with the test perhaps).

gossipsub.go
haxxpop: > At any rate, it is out of scope of this pr.

yeah, let's close this thread

haxxpop: @cortze so it's like sending IHAVEs after publishing (not forwarding) a message using the fanout? got it thanks, but my intuition is that the fanout publishers shouldn't consequently send IHAVEs. Having only the forwarders send them should be enough.

cortze: That should be something that could be discussed. If you published the message, even if you are not subscribed to the topic, you definitely have it and I don't see why you wouldn't share it on an IHAVE (assuming that the remote peer IS subscribed to the topic).
I was just introducing the edgy case where you could receive IWANT messages on topics that you are not subscribed to.

gossipsub.go
ppopth: It's a follow-up from this comment https://github.com/libp2p/go-libp2p-pubsub/pull/553#discussion_r1587434359

haxxpop: Oh no, I intended to increase the GossipSubMaxIDontWantMessages. Sorry.

haxxpop:
@vyzo I didn't get any error with FuzzAppendOrMergeRPC, did you?
```
$ go test -v -run FuzzAppendOrMergeRPC
=== RUN   FuzzAppendOrMergeRPC
--- PASS: FuzzAppendOrMergeRPC (0.00s)
PASS
ok  	github.com/libp2p/go-libp2p-pubsub	0.005s
```

vyzo:
thats a flake then.
Ok, disable and open follow up issue!

haxxpop:
done

vyzo:
CI failures - seems like a faulty commit.

vyzo:
@ppopth let's reenable the Fuzz test, it seems it is fine: see #574

gossipsub.go
```diff
@@ -663,6 +691,36 @@ func (gs *GossipSubRouter) AcceptFrom(p peer.ID) AcceptStatus {
 	return gs.gate.AcceptFrom(p)
 }
 
+// PreValidation sends the IDONTWANT control messages to all the mesh
+// peers. They need to be sent right before the validation because they
+// should be seen by the peers as soon as possible.
+func (gs *GossipSubRouter) PreValidation(msgs []*Message) {
```
Stebalien, line 702: We should document that `IDontWantMessageThreshold` should be greater than `GossipSubDhi * MessageIDSize` to prevent traffic amplification attacks given that this is pre-validation.

gossipsub.go
```diff
@@ -2037,3 +2148,13 @@ func shuffleStrings(lst []string) {
 		lst[i], lst[j] = lst[j], lst[i]
 	}
 }
+
+func computeChecksum(mid string) checksum {
+	var hashed checksum
+	if len(mid) > 32 {
+		hashed = sha256.Sum256([]byte(mid))
+	} else {
+		copy(hashed[:], mid)
```
Stebalien: What happens if we get this wrong? I ask because, given this function:

1. `[1, 2, 3]` has the same checksum as `[1, 2, 3, 0, 0, 0]`.
2. If an attacker can exactly specify their message ID somehow, they can pick a message ID that's the hash of some other, longer message ID.

In practice, I _kind_ of doubt either of these cases matter. But my original suggestion had an extra byte to distinguish between these cases.

rpc_queue.go
Stebalien: These are two different things:

1. Keeping the memory of the slice around is one thing and is probably ok (the queues are expected to be short).
2. Keeping the memory of the slice elements around (heap allocated) is a waste and easy to resolve.

These elements are RPC structs, so they're not small.

rpc_queue.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	done := make(chan struct{})
+
+	unregisterAfterFunc := context.AfterFunc(ctx, func() {
+		// Wake up all the waiting routines. The only routine that correponds
+		// to this Pop call will return from the function. Note that this can
+		// be expensive, if there are too many waiting routines.
+		q.dataAvailable.Broadcast()
+		done <- struct{}{}
```
Stebalien: This seems unnecessary and less efficient as it _waits_ until broadcast is called before returning. Any reason not to just use `ctx.Done()`?

rpc_queue.go
haxxpop: What should be the hard-coded initial constant size of the memory then? or should we implement the dynamic allocation of the ring buffer ourselves?, i.e. reallocate the ring buffer if the size exceeds the capacity.

I think this optimization needs a follow up issue for more discussion and probably a new PR.

Stebalien: Oh, no. I agree we shouldn't implement any of that now. I'm just saying we should implement the first suggestion in https://github.com/libp2p/go-libp2p-pubsub/pull/553#discussion_r1705907547.

Specifically, `slice[0] = nil` before re-slicing.

gossipsub.go
haxxpop: Yeah, I will do as you suggested.

rpc_queue.go
haxxpop: I thought the subroutines should finish before the main routine, but never mind, let's use `ctx.Done()`.

haxxpop:
@Stebalien @vyzo I think it's all done

rpc_queue.go
Stebalien: > I thought the subroutines should finish before the main routine, but never mind.

I generally agree but... eh, it probably doesn't matter here.

Stebalien (approved):

vyzo:
running the CI, will merge once it greens.

haxxpop:
should I squash the fix-up commits?

vyzo (approved):
Looks good! Thank you!

vyzo:
> should I squash the fix-up commits?

no need, I will squash merge anyway

PR #555, opened 2024-04-17
Fix: Own our CertifiedAddrBook

MarcoPolo:
This is blocked until the next go-libp2p release that includes https://github.com/libp2p/go-libp2p/pull/2759, but please still review.

Fixes GossipSub's Peer Exchange (PX) after go-libp2p's change to stop consuming signed peer records into its Certified Address book.

I'll briefly summarize the problem and this solution here, but for more context please follow the links below.

The problem:

- GossipSub's PX relied on go-libp2p's Identify consuming the signed peer record and inserting it into the host's peerstore.
- go-libp2p's Identify was [updated](https://github.com/libp2p/go-libp2p/pull/2325) to avoid saving the signed peer records because:
  - They were not consumed anywhere. (by go-libp2p)
  - We need to filter the addresses to avoid polluting the address book with local addresses.
- GossipSub's PX silently broke as it depended on an up to date CertifiedAddrBook on the host, but it was effectively unused.

The proposed solution in this PR:

- Use the emitted events from go-libp2p to maintain gossipsub's own Certified Address Book.

A slightly more long term solution would be for go-libp2p to support services https://github.com/libp2p/go-libp2p/issues/1993 that can provide resources that can be shared amongst other services/protocols. Imagine if two services needed a certified address book, then it would make sense to have a separate service that could provide an up to date address book.

For now potentially duplicating data if multiple services require a certified addr book seems like an okay solution.

Related though, is anyone aware of other protocols that depend on the host's certified address book?  

Fixes https://github.com/libp2p/go-libp2p/issues/2754
Related to: https://github.com/libp2p/go-libp2p/issues/2355

PR #558, opened 2024-05-20
chores: bump go-libp2p

Wondertan:
Bumps libp2p and resolves Transient -> Limited breaking change


MarcoPolo:
Thanks @Wondertan. I'd like to get #555 in since it'll fix this and fix the gossipsub PX. I'm just blocked on debugging a failing test. Should be ready soon

Wondertan:
Ah, I didn't realize that #555 contains the bump. I would still split the bump and bug fix in separate PRs, but that's up to you. Close the PR as necessary.

MarcoPolo:
Generally agree. The reason for the exception here is that the fix requires the version bump. That said, we could still do this one first. 

I'll do that. Thanks again for the PR :)

MarcoPolo (approved):

PR #564, opened 2024-07-11
Switch to the new peer notify mechanism

### LINKED ISSUE #546: NewStream failure right after startup
This took me forever debug, but here's the issue. If you start up two nodes and immediately join a topic they will fail discover each other in the pubsub. Here's why:

Node1
   - Create host
   - Create pubsub
   
Node2
   - Create host 
       -  This causes an immediate connection to Node1 but the pubsub protocol is not yet registered. 
       -  Node1's pubsub detects the new connection, tries to open a stream, and NewStream fails because Node2 doesn't yet speak pubsub.
   - Create pubsub
      - Node2's pubsub doesn't detect the connection since it was established before the pubsub was created.

Nodes 1 and 2 never connect :(

Fortunately I was able to come up with a fix that doesn't require any changes in libp2p. Basically detect the protocolupdate event and force it think it received a new connection:

```go
        protocolUpdatedSub, err := host.EventBus().Subscribe(new(event.EvtPeerProtocolsUpdated))
	if err != nil {
		return nil, err
	}
	go func(sub event.Subscription) {
		for {
			select {
			case evt, ok := <-sub.Out():
				if !ok {
					return
				}
				var updated bool
				for _, proto := range evt.(event.EvtPeerProtocolsUpdated).Added {
					if proto == cfg.params.ProtocolPrefix+pubsub.GossipSubID_v11 {
						updated = true
						break
					}
				}
				if updated {
					for _, c := range host.Network().ConnsToPeer(evt.(event.EvtPeerProtocolsUpdated).Peer) {
						(*pubsub.PubSubNotif)(ps).Connected(host.Network(), c)
					}
				}
			}
		}
	}(protocolUpdatedSub)
```

But it would be nice if this was built in as it's definitely not expected behavior.

Stebalien:
1. Only listen for peers added and identify events.
2. Remove the old "Limited" check. Peers only show up as "Connected" if they have non-limited connections.
3. Don't bother listening for new connections directly and/or connectivity changes. We'll get a new identify event per new connection regardless.

fixes #546


peer_notify.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/libp2p/go-libp2p/core/event"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
+)
+
+func (ps *PubSub) watchForNewPeers(ctx context.Context) {
+	// We don't bother subscribing to "connectivity" events because we always run identify after
+	// every new connection.
```
Stebalien, line 14: I _think_ this is right but I'm not 100% certain. I remember remember running into some issues with this with respect to bitswap...

cc @sukunrt do you think the code here is sufficient?

vyzo (approved):
Looks reasonable.
In fact, I am quite happy we got rid of that ancient notify code!

Stebalien:
Apparently pubsub uses the "blank" host in tests, which doesn't support identify.

Stebalien:
Tests should be fixed by https://github.com/libp2p/go-libp2p-pubsub/pull/565.

Stebalien:
This conflicts with https://github.com/libp2p/go-libp2p-pubsub/pull/440. I'd prefer to just remove that feature because it doesn't work reliably with identify.

vyzo:
I agree, but we should check with users of the feature.
If there is none, no problem.

If it is in use, we might have to make some changes; maybe have a user function to test instead of directly comparing with the router protocols?

Stebalien:
Hm. You're right, we _have_ the function and can just use that.

Stebalien:
Ok, it should all work correctly-ish now. Or, well, correct enough.

peer_notify.go
sukunrt: Yeah this looks fine. We run identify after connecting on every connection.

PR #569, opened 2024-07-17
Optimize Garbage Collection For Message Deliveries

nisdas:
In the event a delivery record becomes old and expires, we also clear out the associated peer map tracking the different peers who have propagated the message.


Stebalien:
We're deleting the entire map anyways. If something else is hanging onto a copy of that map (it shouldn't be) we should fix that. Clearing the map explicitly isn't going to clear memory any faster.

Closing for now but please tell me if I'm missing something obvious here.

PR #578, opened 2024-09-05
Update go-libp2p to latest

gammazero:
- Update to go-libp2p v0.36.3
- Remove unnecessary timer channel read

### DIFF
--- a/discovery.go
+++ b/discovery.go
@@ -245,7 +245,9 @@ func (d *discover) Bootstrap(ctx context.Context, topic string, ready RouterRead
 	}
 
 	t := time.NewTimer(time.Hour)
-	t.Stop()
+	if !t.Stop() {
+		<-t.C
+	}
 	defer t.Stop()
 
 	for {
--- a/discovery.go
+++ b/discovery.go
@@ -245,9 +245,7 @@ func (d *discover) Bootstrap(ctx context.Context, topic string, ready RouterRead
 	}
 
 	t := time.NewTimer(time.Hour)
-	if !t.Stop() {
-		<-t.C
-	}
+	t.Stop()
 	defer t.Stop()
 
 	for {
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/ipfs/go-log/v2 v2.5.1
 	github.com/libp2p/go-buffer-pool v0.1.0
-	github.com/libp2p/go-libp2p v0.36.2
+	github.com/libp2p/go-libp2p v0.36.3
 	github.com/libp2p/go-libp2p-testing v0.12.0
 	github.com/libp2p/go-msgio v0.3.0
 	github.com/multiformats/go-multiaddr v0.13.0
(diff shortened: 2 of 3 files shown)


discovery.go
```diff
@@ -245,9 +245,7 @@ func (d *discover) Bootstrap(ctx context.Context, topic string, ready RouterRead
 	}
 
 	t := time.NewTimer(time.Hour)
-	if !t.Stop() {
-		<-t.C
-	}
+	t.Stop()
```
vyzo: please dont change this

vyzo (changes requested):
please dont introduce bugs.

discovery.go
gammazero: Out of curiosity, I would like to know why this introduces a bug. It seems to me that the purpose of this code is to create a stopped timer. The line immediately above the call to stop creates a timer with a timeout of 1 hour, so it is not conceivable that the timer's channel should ever need to be cleared.

Further, even if it were possible for more than an hour to elapse between these two consecutive lines of code, reading from the timer's channel would be a bug in go 1.23+, since trying to read from a timer's channel following a call to `Stop` will block forever.

vyzo: read more carefully, if Stop returns false you got to drain it.

gammazero: I did try to read carefully, [Here](https://pkg.go.dev/time@go1.23.0#Timer.Stop) it says this:
> For a chan-based timer created with NewTimer(d), as of Go 1.23, **any receive from t.C after Stop has returned is guaranteed to block rather than receive a stale time value from before the Stop**; if the program has not received from t.C already and the timer is running, Stop is guaranteed to return true. Before Go 1.23, the only safe way to use Stop was insert an extra <-t.C if Stop returned false to drain a potential stale value. See the [NewTimer](https://pkg.go.dev/time@go1.23.0#NewTimer) documentation for more details.

Sorry if I am missing something.

FYI: Here is a way to handle the situation if multiple versions of go need to Stop and reuse timers:
https://github.com/quic-go/quic-go/pull/4659/files

> if Stop returns false you got to drain it

That one hour timeout does a good job of ensuring that the immediately following call to `Stop` will never return `false`, so no need to drain even with go < 1.23.

vyzo: so they changed semantics for go 1.23 so as to _break_ existing code that was following their recommemdations? Gross.

Ok fine, but still remove it from this pr and open a new one separately to have visibility and set the minimum version to 1.23; i dont like the 1hr workaround.

sukunrt: I am trying to get a small repro to see if it's actually a golang bug. Ideally `timer.Stop` shouldn't return false if we haven't read anything. 

The state of the draft PR https://github.com/quic-go/quic-go/pull/4659/files is correct though. You can rely on the capacity of the channel to handle this case.

vyzo: it can (at least before 1.23) return false if the timer has fired.

At any rate, lets make a separate pr for this and limit this one to the go-libp2p upgrade.

sukunrt: >it can (at least before 1.23) return false if the timer has fired.

Yes and it does. 

From go1.23, it shouldn't.

vyzo (approved):

PR #579, opened 2024-09-09
fix: stop timer in a way that works before and after go1.23

gammazero:
In this case, a stopped timer is created. Since the timer duration is set to 1 hour and is immediately stopped, the timer should never fire and the channel should never need to be cleared (as is generally needed for go1.22). For go-1.23 the timer semantics have changed so that reading the timer channel following Stop will always block. Therefore, for both these cases, only calling Stop is sufficient to create the stopped timer.

### DIFF
--- a/discovery.go
+++ b/discovery.go
@@ -245,9 +245,7 @@ func (d *discover) Bootstrap(ctx context.Context, topic string, ready RouterRead
 	}
 
 	t := time.NewTimer(time.Hour)
-	if !t.Stop() {
-		<-t.C
-	}
+	t.Stop()
 	defer t.Stop()
 
 	for {


vyzo (commented):
You know, I think `t.Stop()` might always return `true` in which case we don't need to do anything.

Otherwise, we will have to also add a go 1.23 reuirement to go.md

vyzo:
By not doing anything, it means drop the patch altogether.

Can you please test the behavior of `timer.Stop()` in `go1.23`? If it always returns true, then we can close this pr.

gammazero:
In go1.23 `timer.Stop`, following `time.NewTimer` always returns `true` unless:
- `Stop` has been called
- The timer channel has been read

Once `Stop` has been called, another call to Stop always returns false. Since the timer channel is unbuffered, there is no chance for a value to have been written to it.

vyzo:
ok, please add a requirement for go1.23 to go.mod then.

sukunrt (changes requested):
The existing code is fine. Let's wait for the upstream fix: https://github.com/golang/go/issues/69312

Alternatively, you can change the time.Hour to `math.MaxInt64` and keep the rest of the code same.

gammazero:
Let's just leave it for now. Changing it to require go 1.23 may cause problems with things that depend on this package.

PR #580, opened 2024-09-24
Do not call Debugf for non-debug levels in doDropRPC

algorandskiy:
In high load scenarios when consumer is slow, `doDropRPC` is called often and makes extra unnecessary allocations invoking `log.Debug`.

Fixed by checking log level before running expensive formatting.

Before:
```
BenchmarkAllocDoDropRPC-10    	13684732	        76.28 ns/op	     144 B/op	       3 allocs/op
```

After:
```
BenchmarkAllocDoDropRPC-10    	28140273	        42.88 ns/op	     112 B/op	       1 allocs/op
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/libp2p/go-msgio v0.3.0
 	github.com/multiformats/go-multiaddr v0.13.0
 	github.com/multiformats/go-varint v0.0.7
+	go.uber.org/zap v1.27.0
 )
 
 require (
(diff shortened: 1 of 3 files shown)


vyzo (approved):
nice one.

vyzo:
Can you `go mod tidy` please?

algorandskiy:
done.

There is also extra allocation in `Message.GetFrom()` on `[]byte` to `peer.ID` conversion (`runtime.slicebytetostring`) but I'm not sure how to fix it properly.  One approach could be caching `[]byte` -> `peer.ID` on a first call and reuse on subsequent calls. Another one is to use `[]byte` instead of `peer.ID` everywhere but it does not look right (although I'm not sure why `peer.ID` is a string and not `[]byte` as it contains binary data).

vyzo:
peer.ID is a string so that you can easily compare and use as a key in maps, there is a lot of convenience to that.

algorandskiy:
I updated the PR title/description to properly denote the source of extra allocations: it is strings escaping b/c of `interface{}` in `Debugf` signature, and not the message formatting itself that happens later inside `Debugf` after checking the log level (I've confused it with `fmt.Sprintf()` in `doDropRPC` callers.

PR #585, opened 2024-12-06
Bump go-libp2p

jimmykarily:
Trying to fix tests from this PR: https://github.com/libp2p/go-libp2p-pubsub/pull/584


mudler:
:wave: @vyzo @marten-seemann maybe you guys can help here in the reviews? context: go1.22 is affected by CVEs (https://osv.dev/vulnerability/GO-2024-3302) https://github.com/quic-go/quic-go/pull/4729 

Thank you in advance ! :bow:

go.mod
```diff
@@ -1,13 +1,15 @@
 module github.com/libp2p/go-libp2p-pubsub
 
-go 1.22
+go 1.22.0
+
+toolchain go1.23.3
```
marten-seemann: Not necessary.

go.mod
```diff
@@ -1,13 +1,15 @@
 module github.com/libp2p/go-libp2p-pubsub
 
-go 1.22
+go 1.22.0
```
marten-seemann: Don't change this.

marten-seemann (changes requested):
Misleading PR title. This is NOT fixing a test. It's simply bumping some deps.

jimmykarily:
> Misleading PR title. This is NOT fixing a test. It's simply bumping some deps.

The title refers to the tests failing on this PR: https://github.com/libp2p/go-libp2p-pubsub/pull/584

go.mod
jimmykarily: It was automatically changed because of my local versions used. Removed now.

go.mod
jimmykarily: removed

jimmykarily:
> > Misleading PR title. This is NOT fixing a test. It's simply bumping some deps.
> 
> The title refers to the tests failing on this PR: #584

Reworded the commit message to make more sense after merging

marten-seemann (approved):

vyzo:
What exactly is the issue?

We usually bump go-libp2p only when there are breaking changes, the idea is not to force any particular version and make sure latest works. We do that as matter of policy because upgrading libp2p in upstream projects is kind of a big deal.

Having said that, if there is a good reason for the bump, sure lets do it.

vyzo:
the right way to do this is with a pr directly to master, dependabot can rebase.

marten-seemann:
The quic-go vulnerability only shows up because quic-go is imported by go-libp2p.

jimmykarily:
> the right way to do this is with a pr directly to master, dependabot can rebase.

I don't see any difference but I can change it if you prefer that

go.mod
jimmykarily: For some reason, without this change, if fails with:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```
strange go...

go.mod
jimmykarily: https://github.com/golang/go/issues/65847

jimmykarily:
> > the right way to do this is with a pr directly to master, dependabot can rebase.
> 
> I don't see any difference but I can change it if you prefer that

done: https://github.com/libp2p/go-libp2p-pubsub/pull/586

PR #586, opened 2024-12-09
Bump go-libp2p

jimmykarily:
because otherwise, security scanners complain about this:

https://osv.dev/vulnerability/GO-2024-3302

Replaces  [this PR](https://github.com/libp2p/go-libp2p-pubsub/pull/585) as per [the comment](https://github.com/libp2p/go-libp2p-pubsub/pull/585#issuecomment-2522824629)


marten-seemann:
Security scanners can be silenced if they produce nonsensical results. And if they can't, they're worthless and should not be taken seriously in the first place.

jimmykarily:
> Security scanners can be silenced if they produce nonsensical results. And if they can't, they're worthless and should not be taken seriously in the first place.

Correct. But what makes you say it's a nonsensical result in this case? Actually, you were the one to provide the fix:

- https://github.com/quic-go/quic-go/pull/4729/files
- https://github.com/quic-go/quic-go/security/advisories/GHSA-px8v-pp82-rcvr

is it not needed?

marten-seemann:
We're going in circles: https://github.com/libp2p/go-libp2p-pubsub/pull/585#issuecomment-2522812014.

jimmykarily:
> We're going in circles: [#585 (comment)](https://github.com/libp2p/go-libp2p-pubsub/pull/585#issuecomment-2522812014).

"the idea is not to force any particular version"

How exactly is go-libp2p-pubsub doing this? . The way go dependencies work, this line here makes sure a specific version is being used: https://github.com/libp2p/go-libp2p-pubsub/blob/c06df2f9a38e9382e644b241adf0e96e5ca00955/go.mod#L10

marten-seemann:
It is NOT necessary to bump dependencies **in every single library** that is only ever used transitively (like go-libp2p-pubsub, which is only used in conjunction with go-libp2p). Same reason we don't update to the quic-go release in masque-go or webtransport-go: **Applications** can just update their dependencies. This is the place where security scanners make sense.

I'm going to disengage from this discussion now, and I'll leave it to @vyzo to deal with PR.

jimmykarily:
You got me reading docs :D (which is a good thing). I guess it's misunderstanding on my side on what [go.sum](https://go.dev/ref/mod#go-sum-files) entries mean. Probably the security scan tool makes the same mistake. Feel free to ignore and close this PR.

marten-seemann:
Nope, the scan is _technically_ correct. **If** you were running go-libp2p-pubsub as a standalone program (which you don't, it's a library!), then you'd include a vulnerable quic-go version. Since go-libp2p-pubsub is **always** used with go-libp2p, it doesn't really matter if go-libp2p-pubsub specifies an old version, since the version defined by go-libp2p is used.

PR #587, opened 2024-12-17
Gossipsub v2.0

haxxpop:
The gossipsub v2.0 is defined in the spec here https://github.com/libp2p/specs/pull/653.

I made this PR as a draft first because the spec is still a draft and to get some visibility.

I need to implement this early even if the spec is still in an early stage because I have to use it to do simulations.

- [x] Handling receiving messages and send out `IANNOUNCE`
- [x] Adding test cases for sending out `IANNOUNCE`
- [x] Handling `IANNOUNCE` and send out `INEED`
- [x] Adding test cases for handling `IANNOUNCE` and sending out `INEED`
- [x] Handling `INEED` and sending out the message
- [x] Adding test cases for handling `INEED`

Optional:
- [ ] Adding `WithIgnoreIneed` option to make it easier to create a malicious node that never replies to `INEED`


haxxpop:
I'm not sure if `WithIgnoreIneed` should be added to the upstream repo, but I will implement it in my local repo anyway because I need it for my simulations in malicious cases.

vyzo (commented):
let me know when you think it is ready for review.

PR #590, opened 2024-12-27
Improve IDONTWANT Flood Protection

nisdas:
In this PR we add in a new config parameter called `MaxIDontWantLength`  which would be very similarly used as `MaxIHaveLength` has been used of `IHAVE` messgaes . This parameter has been set as the value of `10` now. 

The main purpose is to bring how IDONTWANT messages are handled in line with how IHAVE have been handled. We add the relevant changes to the `handleIDontWant` method along with adding in a new regression test for this check.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -68,7 +68,8 @@ var (
 	GossipSubGraftFloodThreshold              = 10 * time.Second
 	GossipSubMaxIHaveLength                   = 5000
 	GossipSubMaxIHaveMessages                 = 10
-	GossipSubMaxIDontWantMessages             = 1000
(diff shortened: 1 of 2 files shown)


vyzo (approved):
LGTM

ppopth:
I think you have to make sure that you don't send more than `MaxIDontWantLength` on the sending side as well.

I guess you have to make sure that other implementations have to send fewer than `MaxIDontWantLength` ids as well, which can be a problem?

nisdas:
@vyzo Is it possible to tag a new release with this ?

vyzo:
yes, can do

PR #591, opened 2024-12-29
Allow cancelling IWANT using IDONTWANT

haxxpop:
As specified in the Gossipsub v1.2 spec, we should allow cancelling IWANT by IDONTWANT.

That is if IDONTWANT already arrived, we should not process IWANT.

However due to the code structure, we can cancel IWANT only in handleIWant.

https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#cancelling-iwant

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -839,6 +839,11 @@ func (gs *GossipSubRouter) handleIWant(p peer.ID, ctl *pb.ControlMessage) []*pb.
 	ihave := make(map[string]*pb.Message)
 	for _, iwant := range ctl.GetIwant() {
 		for _, mid := range iwant.GetMessageIDs() {
+			// Check if that peer has sent IDONTWANT before, if so don't send them the message
+			if _, ok := gs.unwanted[p][computeChecksum(mid)]; ok {
+				continue
+			}
+
 			msg, count, ok := gs.mcache.GetForPeer(mid, p)
 			if !ok {
 				continue
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -3079,6 +3079,110 @@ func TestGossipsubIdontwantSmallMessage(t *testing.T) {
 	<-ctx.Done()
 }
 
+// Test that IWANT will have no effect after IDONTWANT is sent
+func TestGossipsubIdontwantBeforeIwant(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	hosts := getDefaultHosts(t, 3)
+
+	msgID := func(pmsg *pb.Message) string {
+		// silly content-based test message-ID: just use the data as whole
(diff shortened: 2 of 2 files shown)


gossipsub.go
```diff
@@ -839,6 +839,11 @@ func (gs *GossipSubRouter) handleIWant(p peer.ID, ctl *pb.ControlMessage) []*pb.
 	ihave := make(map[string]*pb.Message)
 	for _, iwant := range ctl.GetIwant() {
 		for _, mid := range iwant.GetMessageIDs() {
+			// Check if that peer has sent IDONTWANT before, if so don't send them the message
+			if _, ok := gs.unwanted[p][computeChecksum(mid)]; ok {
```
vyzo, line 843: how expensive is this?

vyzo (commented):
Not so sure we need this to be honest, i'd like some more opinions.

vyzo:
@nisdas want to chime in?

nisdas:
Is it common for non-mesh peers to send us IDONTWANTs ? The way i see it both control messages target different peers. 

IDONTWANT - mesh , IWANT - non mesh

gossipsub.go
haxxpop: You meant `computeChecksum`? no, it's just one hash function.

haxxpop:
> Is it common for non-mesh peers to send us IDONTWANTs ?

I don't think it's quite useful either, but it's specified in the spec and Lighthouse is already sending IDONTWANT after it sends IWANT. See https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs#L2733

While we don't send IDONTWANT to non-mesh peers ourselves, I think it's useful to handle IDONTWANT for Lighthouse, otherwise their IDONTWANT will be wasteful.

So yes, it's common.

vyzo (approved):
ok, lets merge

nisdas:
@ppopth One thing that is important to confirm is if other implementations correctly ignore this for scoring. As of gossipsub v1.1 , unfulfilled promises lead to the downscoring of peers. So if you advertise an `IHAVE` and I reply back with an `IWANT` . I will normally expect you to send me the full message. 

However with `IDONTWANT` we are now skipping sending the full message. Peers which send both `IWANT` and `IDONTWANT` for the same message should also not penalize peers for not fulfilling the promise.

haxxpop:
> So if you advertise an `IHAVE` and I reply back with an `IWANT` . I will normally expect you to send me the full message.

Yeah, it's specified in the spec that if I don't reply to `IWANT`, I would be downscored. But I don't worry about it because if you downscore me, but you sent both `IWANT` and `IDONTWANT`, you seem more misbehaving.

https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#spam-protection-measures

PR #601, opened 2025-02-26
feat: include size of RPC messages in Traces

arajasek:


arajasek:
Sorry, meant to open this against my fork :D

PR #607, opened 2025-04-15
feat(gossipsub): Add MessageBatch

MarcoPolo:
to support batch publishing messages

Replaces #602.

Batch publishing lets the system know there are multiple related messages to be published so it can prioritize sending different messages before sending copies of messages. For example, with the default API, when you publish two messages A and B, under the hood A gets sent to D=8 peers first, before B gets sent out. With this MessageBatch api we can now send one copy of A _and then_ one copy of B before sending multiple copies.

When a node has bandwidth constraints relative to the messages it is publishing this improves dissemination time.

For more context see this post: https://ethresear.ch/t/improving-das-performance-with-gossipsub-batch-publishing/21713

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1142,15 +1142,16 @@ func (gs *GossipSubRouter) connector() {
 	}
 }
 
-func (gs *GossipSubRouter) PublishBatch(batch *MessageBatch) {
-	for _, msg := range batch.messages {
+func (gs *GossipSubRouter) PublishBatch(messages []*Message, opts *BatchPublishOptions) {
+	strategy := opts.Strategy
+	for _, msg := range messages {
 		msgID := gs.p.idGen.ID(msg)
 		for p, rpc := range gs.rpcs(msg) {
-			batch.Scheduler.AddRPC(p, msgID, rpc)
+			strategy.AddRPC(p, msgID, rpc)
 		}
 	}
 
-	for p, rpc := range batch.Scheduler.All() {
+	for p, rpc := range strategy.All() {
 		gs.sendRPC(p, rpc, false)
 	}
 }
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1146,11 +1146,11 @@ func (gs *GossipSubRouter) PublishBatch(batch *MessageBatch) {
 	for _, msg := range batch.messages {
 		msgID := gs.p.idGen.ID(msg)
 		for p, rpc := range gs.rpcs(msg) {
-			batch.Strategy.AddRPC(p, msgID, rpc)
+			batch.Scheduler.AddRPC(p, msgID, rpc)
 		}
 	}
 
-	for p, rpc := range batch.Strategy.All() {
+	for p, rpc := range batch.Scheduler.All() {
 		gs.sendRPC(p, rpc, false)
 	}
 }
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -3,10 +3,12 @@ package pubsub
 import (
 	"context"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
 	"sort"
+	"sync"
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
@@ -1140,6 +1142,10 @@ func (gs *GossipSubRouter) connector() {
 }
 
 func (gs *GossipSubRouter) Publish(msg *Message) {
+	if msg.messageBatch != nil {
+		defer msg.messageBatch.doneWithMsg()
+	}
+
 	gs.mcache.Put(msg)
 
 	from := msg.ReceivedFrom
@@ -1213,6 +1219,10 @@ func (gs *GossipSubRouter) Publish(msg *Message) {
 			continue
 		}
 
+		if msg.messageBatch != nil {
+			msg.messageBatch.queueRPC(pid, gs.p.idGen.ID(msg), out)
+			continue
+		}
 		gs.sendRPC(pid, out, false)
 	}
 }
@@ -2203,4 +2213,81 @@ func computeChecksum(mid string) checksum {
 		cs.length = uint8(copy(cs.payload[:], mid))
 	}
 	return cs
+}
+
+type pendingRPC struct {
+	peer peer.ID
+	rpc  *RPC
+}
+
+// MessageBatch allows a user to batch related messages and then publish them
+// at once. This allows the system to prioritize sending a single copy of each
+// message before sending more copies. This helps bandwidth constrained peers.
+type MessageBatch struct {
+	sync.Mutex
+	// PendingRPCsToAdd is a waitgroup that is used to wait for all the RPCs to
+	// be added to the batch. This library's publish is async, so we need to be
+	// careful to not publish the batch before all the RPCs are added.
+	pendingRPCsToAdd sync.WaitGroup
+	sendRPC          func(peer peer.ID, rpc *RPC, urgent bool)
+	rpcs             map[string][]pendingRPC
+}
+
+// NewMessageBatch creates a new MessageBatch. This only works for GossipSub.
+func NewMessageBatch(ps *PubSub) (*MessageBatch, error) {
+	if ps == nil {
+		return nil, errors.New("pubsub is nil")
+	}
+	if gs, ok := ps.rt.(*GossipSubRouter); ok {
+		return &MessageBatch{
+			sendRPC: gs.sendRPC,
+		}, nil
+	}
+	return nil, errors.New("pubsub is not a GossipSubRouter")
+}
+
+// Add adds a message to the batch.
+func (p *MessageBatch) Add(ctx context.Context, topic *Topic, data []byte, opts ...PubOpt) error {
+	p.pendingRPCsToAdd.Add(1)
+	opts = append(opts, func(o *PublishOptions) error {
+		o.messageBatch = p
+		return nil
+	})
+	return topic.Publish(ctx, data, opts...)
+}
+
+// Publish publishes the messages in the batch.
+//
+// Users should make sure there is enough space in the Peer's outbound queue to
+// ensure messages are not dropped. WithPeerOutboundQueueSize should be set to
+// at least the expected number of batched messages per peer plus some slack to
+// account for gossip messages.
+func (p *MessageBatch) Publish() {
+	p.pendingRPCsToAdd.Wait()
+	p.Lock()
+	defer p.Unlock()
+
+	for len(p.rpcs) > 0 {
+		for msgID, rpcs := range p.rpcs {
+			if len(rpcs) == 0 {
+				delete(p.rpcs, msgID)
+				continue
+			}
+			p.sendRPC(rpcs[0].peer, rpcs[0].rpc, false)
+			p.rpcs[msgID] = rpcs[1:]
+		}
+	}
+}
+
+func (p *MessageBatch) doneWithMsg() {
+	p.pendingRPCsToAdd.Done()
+}
+
+func (p *MessageBatch) queueRPC(peer peer.ID, msgID string, rpc *RPC) {
+	p.Lock()
+	defer p.Unlock()
+	if p.rpcs == nil {
+		p.rpcs = make(map[string][]pendingRPC)
+	}
+	p.rpcs[msgID] = append(p.rpcs[msgID], pendingRPC{peer: peer, rpc: rpc})
 }
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -3,7 +3,6 @@ package pubsub
 import (
 	"context"
 	"crypto/sha256"
-	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -523,6 +522,8 @@ type GossipSubRouter struct {
 	// clean up -- eg backoff clean up.
 	heartbeatTicks uint64
 }
+
+var _ BatchPublisher = &GossipSubRouter{}
 
 type connectInfo struct {
 	p   peer.ID
@@ -1141,11 +1142,30 @@ func (gs *GossipSubRouter) connector() {
 	}
 }
 
+func (gs *GossipSubRouter) PublishBatch(batch *MessageBatch) {
+	batch.pendingMsgs.Wait()
+	batch.Strategy.PublishRPCsWithFn(func(peer peer.ID, rpc *RPC) {
+		gs.sendRPC(peer, rpc, false)
+	})
+}
+
 func (gs *GossipSubRouter) Publish(msg *Message) {
 	if msg.messageBatch != nil {
-		defer msg.messageBatch.doneWithMsg()
+		defer msg.messageBatch.pendingMsgs.Done()
+		// This message is part of a batch. Instead of sending the related RPCs
+		// immediately, we queue them.
+		gs.publish(msg, func(p peer.ID, msgID string, rpc *RPC, urgent bool) {
+			msg.messageBatch.Strategy.QueueRPC(p, msgID, rpc)
+		})
+		return
 	}
 
+	gs.publish(msg, func(p peer.ID, msgID string, rpc *RPC, urgent bool) {
+		gs.sendRPC(p, rpc, urgent)
+	})
+}
+
+func (gs *GossipSubRouter) publish(msg *Message, sendRPC func(p peer.ID, msgID string, rpc *RPC, urgent bool)) {
 	gs.mcache.Put(msg)
 
 	from := msg.ReceivedFrom
@@ -1219,11 +1239,7 @@ func (gs *GossipSubRouter) Publish(msg *Message) {
 			continue
 		}
 
-		if msg.messageBatch != nil {
-			msg.messageBatch.queueRPC(pid, gs.p.idGen.ID(msg), out)
-			continue
-		}
-		gs.sendRPC(pid, out, false)
+		sendRPC(pid, msg.ID, out, false)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


raulk (changes requested):
There's quite a bit of indirection here.

1. There's a top-level `NewMessageBatch` API that accepts a gossipsub router. This seems inverted. Intuitively, I'd expect to initiate a message batch _from_ a router (similar to [badger's `NewWriteBatch`](https://pkg.go.dev/github.com/dgraph-io/badger/v3#DB.NewWriteBatch)). It seems more idiomatic to me.
2. The `MessageBatch` keeps track of pending RPCs, which are added by the `GossipSubRouter`. To propagate itself through the call stack, it sets an anonymous `PublishOption` wrapping itself.
3. There's special casing in several spots in the `GossipSubRouter` to identify if this is a batch, unwrap the object, and do special things on it like queue the RPCs instead of actually sending them. This results in the paradox that a call to the router's gossipsub `Publish` doesn't actually publish anything under this mode.
4. The fact that a `Message` now embeds a `messageBatch` is semantically counterintuitive.

All in all, I think this code is (a) hard to maintain, and (b) it exposes a confusing public API. I'm curious what alternative APIs you considered. I'd imagine this dance is a last resort to make it work under the current design/implementation constraints.

Did you consider:

1. Extending the PubSub, PubSubRouter, GossipSubRouter, etc. type hierarchy with `NewBatch` and `PublishBatch` methods?
2. Reorganizing the reusable code under `GossipSubRouter#Publish` so the relevant parts can be reused from `PublishBatch`?

I think this would simplify the whole thing.

For what it's worth, in @cskiraly's [the EthResearch post](https://ethresear.ch/t/improving-das-performance-with-gossipsub-batch-publishing/21713), he hinted at a distinction between message batches and queuing/diffusion disciplines. I think that distinction was lost a bit later, but I'd like to recover it here.

In my head:
-  A message batch is no more than an organizational unit to bundle a set of messages that the router learns about at once (without atomicity guarantees, hence not a Transaction). It does not imply a concrete queuing/diffusion/scheduling discipline.
-  The user should provide a queuing/diffusion/scheduling discipline when calling `PublishBatch`, in the form of a function or an interface implementation. This discipline encapsulates the conversion to RPCs and handles their subsequent dispatch. In other words, the code that currently lives under `MessageBatch#Publish` should be modular in itself.

MarcoPolo:
> All in all, I think this code is (a) hard to maintain, and (b) it exposes a confusing public API. I'm curious what alternative APIs you considered. I'd imagine this dance is a last resort to make it work under the current design/implementation constraints.

On (a), you're the reviewer so I'll defer to you. I agree there are subtleties here that may introduce footguns in the future [example](https://github.com/libp2p/go-libp2p-pubsub/pull/602/files#r2017616683). 

On (b), I disagree about this being a confusing public API. To be clear, this is the api users use:
```go
batch, err := NewMessageBatch(pubsub)
// Handle err

for _, msg := range msgs
  err := batch.Add(ctx, topic, msg)
  // Handle err
}
batch.Publish()
```

Whether this is ` NewMessageBatch(pubsub)` or ` pubsub.NewMessageBatch()` is a fairly minor point. Happy to change.

Whether this is `batch.Publish` or `pubsub.PublishBatch` also seems like a fairly minor point, but I'd prefer the former. It's also consistent with your example of `WriteBatch`.

> Did you consider:
> 
> 1. Extending the PubSub, PubSubRouter, GossipSubRouter, etc. type hierarchy with `NewBatch` and `PublishBatch` methods?
> 2. Reorganizing the reusable code under `GossipSubRouter#Publish` so the relevant parts can be reused from `PublishBatch`?

I did, but I feared it would end up as a larger refactor. It may be worth it anyways as it may generally improve the codebase and remove future footguns. I'll open a new PR along these lines.

---
 
> For what it's worth, in @cskiraly's [the EthResearch post](https://ethresear.ch/t/improving-das-performance-with-gossipsub-batch-publishing/21713), he hinted at a distinction between message batches and queuing/diffusion disciplines. I think that distinction was lost a bit later, but I'd like to recover it here.
> 
> In my head:
> 
> * A message batch is no more than an organizational unit to bundle a set of messages that the router learns about at once (without atomicity guarantees, hence not a Transaction). It does not imply a concrete queuing/diffusion/scheduling discipline.
> * The user should provide a queuing/diffusion/scheduling discipline when calling `PublishBatch`, in the form of a function or an interface implementation. This discipline encapsulates the conversion to RPCs and handles their subsequent dispatch. In other words, the code that currently lives under `MessageBatch#Publish` should be modular in itself.

An earlier draft of this PR allowed users to define the publish strategy. I tested via simulation the rarest message first strategy and the "shuffle" strategy from #602. The rarest first performed better (which intuitively makes sense). To that end, I chose to keep the API simpler by only using the rarest-first strategy. We can always make this configurable, but I think it would be a mistake to prematurely add this extension point now. This is not to say that the current rarest-first implementation is optimal, just that the inputs to the optimal solution may be non-obvious and defining an extension point now when we only have n=1 options is premature.

vyzo:
Please hold your horses in the larger refactor.
This something that is being actively discussed as part of the v2.0 initiative, and we shouldnt rush on it.

MarcoPolo:
Consider my horses held. I'll explore a small refactor that hopes to remove some indirection here.

raulk:
@MarcoPolo I'd love to see a version of this PR with the refactor. Happy to pair on it if you'd like!

Re: `batch.Publish()` vs. `pubsub.PublishBatch()`, I suspect the latter can reduce the "spooky action at a distance" effect. It accomplishes that by encapsulating the RPC planning, scheduling, and dispatch all in a single place vs. spread across the indirection. Makes it easier to follow; and I'm all for reducing complexity ;-) But hard to tell without taking a stab.

Re: queuing/dispatch disciplines, even if we _only_ support the "prioritize rarest" one at the moment, it still makes sense to introduce the abstraction as long as we have some confidence that it can withstand future disciplines going forward. Deliberate API design signals intentionality and makes a difference in shaping how APIs evolve. However, if you feel strongly against this, I can live without it.

I agree with @vyzo that we don't want a major refactor here, but introducing this feature cleanly (in an already complex and organic codebase) is a win.

MarcoPolo:
I made significant changes to the design. Thanks @sukunrt and @raulk for the feedback. I think this approach is clearer. I'd recommend initially reviewing with whitespace changes hidden (`?w=1`).

Care was taken to ensure batched messages and normal messages go through as much as the same code flows as possible.

Some refactors along the way worth highlighting:
- Introduce a new `validation.ValidateLocal` method that does not send a message. This lets us validate a message when adding to a batch without also running the send message logic.
- ⚠️ Breaking change: `validation.validate` returns a `ValidationError{Reason: RejectValidationIgnoredDuplicate}` on duplicate instead of a nil error. The only time you would get this error is if you are publishing two duplicate messages, and it's probably better that you get this error instead of silently doing nothing.

messagebatch.go
```diff
@@ -0,0 +1,64 @@
...
+	// All returns an ordered iterator of RPCs.
+	All() iter.Seq2[peer.ID, *RPC]
+}
+
+type pendingRPC struct {
+	peer peer.ID
+	rpc  *RPC
+}
+
+type RarestFirstRPCScheduler struct {
```
raulk: nit: godocs would be great!

messagebatch.go
```diff
@@ -0,0 +1,64 @@
...
+	sync.Mutex
+	rpcs map[string][]pendingRPC
+}
+
+func (s *RarestFirstRPCScheduler) AddRPC(peer peer.ID, msgID string, rpc *RPC) {
+	s.Lock()
+	defer s.Unlock()
+	if s.rpcs == nil {
+		s.rpcs = make(map[string][]pendingRPC)
+	}
```
raulk, line 42: Worth having a constructor for `RarestFirstRPCScheduler` that initializes the map.

gossipsub.go
```diff
@@ -1139,81 +1142,104 @@ func (gs *GossipSubRouter) connector() {
...
+	for p, rpc := range gs.rpcs(msg) {
+		gs.sendRPC(p, rpc, false)
+	}
+}
 
-	from := msg.ReceivedFrom
-	topic := msg.GetTopic()
+func (gs *GossipSubRouter) rpcs(msg *Message) iter.Seq2[peer.ID, *RPC] {
+	return func(yield func(peer.ID, *RPC) bool) {
+		gs.mcache.Put(msg)
```
raulk, line 1167: note to self: MessageCache is not guarded by a lock, but this is fine because it's never used concurrently from the reactor loop. And despite `router.Publish` and `router.PublishBatch` being public, it's misuse if the user calls those directly.

messagebatch.go
```diff
@@ -0,0 +1,64 @@
...
+				}
+				if !yield(rpcs[0].peer, rpcs[0].rpc) {
+					return
+				}
+
+				s.rpcs[msgID] = rpcs[1:]
+			}
+		}
+	}
+}
```
raulk: The RarestFirst name sounds unintuitive to me. "Rare" implies relative scarcity, but relative to what? How about we call it `PrioritizeFirstCopy`? Sounds more descriptive to me.

pubsub.go
```diff
@@ -1358,6 +1378,34 @@ func (p *PubSub) Publish(topic string, data []byte, opts ...PubOpt) error {
...
+// implement the BatchPublisher interface.
+//
+// Users should make sure there is enough space in the Peer's outbound queue to
+// ensure messages are not dropped. WithPeerOutboundQueueSize should be set to
+// at least the expected number of batched messages per peer plus some slack to
+// account for gossip messages.
+func (p *PubSub) PublishBatch(batch *MessageBatch) error {
+	if _, ok := p.rt.(BatchPublisher); !ok {
+		return fmt.Errorf("pubsub router is not a BatchPublisher")
+	}
```
raulk, line 1393: This case evokes panic semantics to me, but happy keeping it as-is if we feel more comfortable this way.

pubsub.go
```diff
@@ -1358,6 +1378,34 @@ func (p *PubSub) Publish(topic string, data []byte, opts ...PubOpt) error {
...
+	var copy MessageBatch
+	copy.Scheduler = batch.Scheduler
+	copy.messages = batch.messages
+	batch.Scheduler = nil
+	batch.messages = nil
+
+	if copy.Scheduler == nil {
+		// Default to Rarest first
+		copy.Scheduler = &RarestFirstRPCScheduler{}
+	}
```
raulk: Let's document this in the godocs for this method.

pubsub.go
```diff
@@ -1358,6 +1378,34 @@ func (p *PubSub) Publish(topic string, data []byte, opts ...PubOpt) error {
...
+// Users should make sure there is enough space in the Peer's outbound queue to
+// ensure messages are not dropped. WithPeerOutboundQueueSize should be set to
+// at least the expected number of batched messages per peer plus some slack to
+// account for gossip messages.
+func (p *PubSub) PublishBatch(batch *MessageBatch) error {
+	if _, ok := p.rt.(BatchPublisher); !ok {
+		return fmt.Errorf("pubsub router is not a BatchPublisher")
+	}
+
+	// Copy the batch to avoid the footgun of reusing scheduler state across batches
```
raulk: Why would this happen? Taking the scheduler as a parameter would avoid this weirdness.

pubsub.go
```diff
@@ -1358,6 +1378,34 @@ func (p *PubSub) Publish(topic string, data []byte, opts ...PubOpt) error {
...
 }
 
+// PublishBatch publishes a batch of messages. This only works for routers that
+// implement the BatchPublisher interface.
+//
+// Users should make sure there is enough space in the Peer's outbound queue to
+// ensure messages are not dropped. WithPeerOutboundQueueSize should be set to
+// at least the expected number of batched messages per peer plus some slack to
+// account for gossip messages.
+func (p *PubSub) PublishBatch(batch *MessageBatch) error {
```
raulk: In terms of API design, taking the `Scheduler` as a second parameter makes this API more self-descriptive. It's kinda unergonomic for the user to remember to set an obscure struct field to supply the scheduler. Also, from a conceptual perspective, the scheduler does not belong in the batch, as the batch is purely an organizational unit to group messages disclosed to pubsub all at once. (And taking the scheduler here would get rid of the copy weirdness below).

```go
// Current approach
var batch MessageBatch
batch.Scheduler = make(PrioritizeFirstCopyScheduler)
topic.AddToBatch(ctx, &batch, []byte("foo"))
topic.AddToBatch(ctx, &batch, []byte("bar"))
pubsub.PublishBatch(&batch)

// Preferred approach
var batch MessageBatch
topic.AddToBatch(ctx, &batch, []byte("foo"))
topic.AddToBatch(ctx, &batch, []byte("bar"))
pubsub.PublishBatch(&batch, make(PrioritizeFirstCopyScheduler))
```

topic.go
```diff
@@ -303,13 +320,18 @@ func (t *Topic) Publish(ctx context.Context, data []byte, opts ...PubOpt) error
...
 				case <-ticker.C:
 				case <-ctx.Done():
-					return fmt.Errorf("router is not ready: %w", ctx.Err())
+					return nil, fmt.Errorf("router is not ready: %w", ctx.Err())
 				}
 			}
 		}
 	}
 
-	return t.p.val.PushLocal(&Message{m, "", t.p.host.ID(), pub.validatorData, pub.local})
```
raulk, line 312: PushLocal looks unused now? Can we remove?

validation.go
```diff
@@ -306,7 +330,7 @@ func (v *validation) validate(vals []*validatorImpl, src peer.ID, msg *Message,
 	id := v.p.idGen.ID(msg)
 	if !v.p.markSeen(id) {
 		v.tracer.DuplicateMessage(msg)
-		return nil
+		return ValidationError{Reason: RejectValidationIgnoredDuplicate}
```
raulk: 🚩 I don't think we can do this. On surface, this appears like a sane thing to do but I suspect it's going to cause notable breakage downstream. The reason being that this will now error if the application attempts to publish a message that it already received from peers, which **certainly** happens in Ethereum when the CL successfully fetches blobs from the EL and broadcasts them. See the description here: https://github.com/libp2p/go-libp2p-pubsub/issues/610.

raulk (commented):
I think we're almost here, barring one major remark, and a few minor comments.

topic.go
vyzo: wait, this is used for local publishing to bypass the pipeline. Are you sure this makes sense to remove?

validation.go
vyzo: ya, this is not so safe to change.

messagebatch.go
MarcoPolo: This way a zero value is valid, which is nice. It's a cheap check.

messagebatch.go
MarcoPolo: I'm not the biggest fan of the name either, but I'm not sure prioritize _first_ copy is right. Rarest first in the sense that the next rarest message in the network will be published next, and between equally rare messages the choice is arbitrary.

pubsub.go
MarcoPolo: A small correction, the current api is:
```go
var batch MessageBatch
topic.AddToBatch(ctx, &batch, []byte("foo"))
topic.AddToBatch(ctx, &batch, []byte("bar"))
pubsub.PublishBatch(&batch)
```

What's nice about this is that the scheduler parameter is optional and has a sane default that we can change easily. The proposed alternative does not have this property.

> taking the scheduler here would get rid of the copy weirdness below

We'd still want to have a copy to clear the messages slice I think

messagebatch.go
```diff
@@ -0,0 +1,64 @@
...
+	All() iter.Seq2[peer.ID, *RPC]
+}
+
+type pendingRPC struct {
+	peer peer.ID
+	rpc  *RPC
+}
+
+type RarestFirstRPCScheduler struct {
+	sync.Mutex
```
MarcoPolo: I don't think we need this mutex

pubsub.go
MarcoPolo: Panic should be reserved for "we have no idea what to do next, the safest thing is to exit the program". Returning an error here seems like a reasonable next step to do, so we have some idea what to do next.

validation.go
MarcoPolo: Yeah, that's a good callout. Here's what we can do instead:

- Introduce a new duplicate error type (private)
- return this new duplicate error here
- callers check against this error and preserve all behavior
  - Topic.Publish no error, do nothing.
  - topic.AddToBatch no error, do not add the message to the batch.

topic.go
MarcoPolo: It is unused. I think we can remove it :+1:

@vyzo, yes I changed this to to `ValidateLocal` which does a similar thing, except does not _always_ push the message to the channel. See the comment in that method.

pubsub.go
MarcoPolo: > Taking the scheduler as a parameter would avoid this weirdness.

It wouldn't. Example:
```go
var batch MessageBatch topic.AddToBatch(ctx, &batch, []byte("foo")) topic.AddToBatch(ctx, &batch, []byte("bar")) 
var strategy WhateverStrategy
pubsub.PublishBatch(&batch, &strategy)
pubsub.PublishBatch(&batch, &strategy) // error!
```

messagebatch.go
MarcoPolo: Maybe `NextRarestRPCScheduler` is better?

validation.go
MarcoPolo: I made this change. Thinking about it more, it's actually reasonable to not return an error in `Publish`/`AddToBatch` for a duplicate case. Semantically we still published the message, even if it was in the past.

pubsub.go
raulk: This seems bad usage -- retaining and reusing an object beyond its intended lifetime. Sadly, Go doesn't have semantics in the type system to handle this; Rust does.

pubsub.go
raulk: Yeah and what I'm saying is that the `Scheduler` is hidden, which is a poor design choice because it's intended to be configurable. And it should be associated with the Publish action, not with the Batch, because it affects how messages are published, not the batch itself. (i.e. the data model is off here)

raulk: If we really wanted to have the ability to elide the strategy and fall back to a default, the idiomatic way in Go is to use the Option paradigm. Idiomatic Go favours explicitness (as evidenced by the pedantic error handling 🙃).

```go
pubsub.PublishBatch(&batch, WithScheduler(&PrioritizeFirstCopyScheduler{})
// or
pubsub.PublishBatch(&batch)
```

validation.go
raulk: Yep, I prefer restoring this and keeping this unchanged.
1. If this is deemed a duplicate because we observed an inbound copy, we already propagated it to peers in the past, so we can discard it silently.
2. If this is deemed a duplicate originating here (second Publish for same message ID), we should swallow it silently because our ability to sense duplicates is time-boxed anyway, causing a non-deterministic error which is hard to deal with. The max I would do here is log a WARN, to flag to the user that there may be some inconsistency in the app layer leading to duplicates. But even that is likely to be noisier than most users want tbh (it's simplier if they can push data to pubsub and trust that the layer will dedup where possible -- this being a feature and not an exception).

pubsub.go
raulk: Not just, panics are often used for more circumstances, amongst which is programmer error, especially in application design (not runtime), which this is.

raulk: But as I said, even if this invalidates all possibility of batch publishing and it's the result of the user wiring the stack incorrectly, I'm happy to go with a runtime error here if it makes us feel safer 👍

messagebatch.go
raulk: Ahhhhh naming. I think "Rare" is a misnomer because it implies that this node is measuring the rate of presence/circulation of concrete messages, to enough degree that it can determine which messages in the batch are Rare in the network, to prioritise the dissemination of those. But that's not what's happening; this is simply a breadth-first / round-robin strategy for dispatch.

pubsub.go
MarcoPolo: Yes, this is a lifetime issue. The best we can do is make a shallow copy and clear the fields of the batch.

pubsub.go
MarcoPolo: I don't have a strong opinion here so I'll add the suggested optional publish batch options.

messagebatch.go
MarcoPolo: `RoundRobinMessageIDScheduler` ?

pubsub.go
MarcoPolo: added

messagebatch.go
raulk: Yeah, let's call it that and merge this baby. It's experimental anyway.

raulk (approved):
@MarcoPolo LGTM; feel free to merge when ready.

MarcoPolo:
Merging this now. It would be nice to get https://github.com/libp2p/go-libp2p-pubsub/pull/612 in before we make a release.

Issue #610, opened 2025-05-06
Experiment with `IDONTWANT` on `Publish()`


raulk:
The `IDONTWANT` control signal was recently introduced to instruct peers to suppress the propagation of messages that a node has already received from other sources (or that it is otherwise aware of). This mechanism aims to reduce unnecessary message duplication within the network.

Currently, go-libp2p-pubsub emits ⁠`IDONTWANT` signals only while processing incoming messages. These signals are dispatched early in the inbound pipeline (specifically, during RPC prevalidation and before messages are handed off to the application). Tests performed in controlled environments by various groups have confirmed measurable reductions in redundant traffic from this new feature.

However, there is an additional opportunity to further decrease superfluous network traffic: sending an ⁠`IDONTWANT` signal to mesh peers when the app invokes ⁠`Publish`. In this scenario, the ⁠`IDONTWANT` would precede the corresponding `Publish` RPC (in fact, the `Publish` RPC itself may even be discarded if the peer had previously signalled disinterest via an ⁠`IDONTWANT`).

This optimization is particularly useful in highly synchronized environments, such as the Ethereum network. When a block is produced, CL nodes typically race to source blobs locally (from the EL), construct sidecars, and disseminate them to peers. In 60–70% of cases the required data is already present in the local EL. In these situations, the application issues a ⁠`Publish` and does not need to receive the same message from its peers, so triggering an `IDONTWANT` would correctly signal that.

raulk:
Apparently rust-libp2p already behaves this way (cc @jxs) 🚀

jxs:
Hi Raúl. Rust-libp2p allows has implemented sending `IDONTWANT` messages when publishing a message: https://github.com/libp2p/rust-libp2p/pull/5773/. 
But now that I re-read your description, it has a small difference from it, in rust-libp2p the `IDONTWANT` is sent after the message to be published, we can easily send it before the published message, it probably makes more sense.
cc @AgeManning

MarcoPolo:
This makes sense. I'm wondering if we could do a bit better.

The problem we're trying to solve here is to reduce duplicates in the case that a large percentage (most?) of the network already has some message that will be propagated on a topic.

In this case, you want to avoid eagerly pushing the message to your peers that likely already have it. Maybe by first sending an IDONTWANT, add the message to our local message cache (so it gets gossiped), and delay pushing the message by a small amount (rtt? rtt/2?) to allow neighbors to give us their IDONTWANT.

[...]

PR #612, opened 2025-05-06
Send IDONTWANT before first publish

MarcoPolo:
See #610 

We previously send IDONTWANT only when forwarding. This has us send IDONTWANT on our initial publish as well. Helps in the case that one or more peers may also publish the same thing at around the same time (see #610 for a longer explanation) and prevents "boomerang" duplicates where a peer sends you back the message you sent before you get a chance to send it to them.

This also serves as a hint to a peer that you are about to send them a certain message.

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -349,6 +349,7 @@ func (t *Topic) validate(ctx context.Context, data []byte, opts ...PubOpt) (*Mes
 	}
 
 	msg := &Message{m, "", t.p.host.ID(), pub.validatorData, pub.local}
+	t.p.rt.PreValidation(t.p.host.ID(), []*Message{msg})
 	err := t.p.val.ValidateLocal(msg)
 	if err != nil {
 		return nil, err
--- a/topic.go
+++ b/topic.go
@@ -349,7 +349,7 @@ func (t *Topic) validate(ctx context.Context, data []byte, opts ...PubOpt) (*Mes
 	}
 
 	msg := &Message{m, "", t.p.host.ID(), pub.validatorData, pub.local}
-	t.p.rt.PreValidation(t.p.host.ID(), []*Message{msg})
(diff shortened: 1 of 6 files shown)


topic.go
```diff
@@ -349,6 +349,7 @@ func (t *Topic) validate(ctx context.Context, data []byte, opts ...PubOpt) (*Mes
 	}
 
 	msg := &Message{m, "", t.p.host.ID(), pub.validatorData, pub.local}
+	t.p.rt.PreValidation(t.p.host.ID(), []*Message{msg})
```
raulk: While we're working on this area, how about we rename `PreValidation()` to something more accurate/obvious, e.g. `DispatchIDONTWANT()` or even something more generic like `Preprocess()` or `Intercept()`? The spirit of the method just isn't to validate (returns nothing), and it's confusing when juxtaposed with `ValidateLocal()` ("what validation happens in each?").

raulk (approved):

topic.go
vyzo: names, names, names 🤣

But yes, preprocessing is what it does.

MarcoPolo: No strong opinion. `Preprocess` seems fine.

haxxpop: `Preprocess` is good for me too and no strong preference about it.

> The spirit of the method just isn't to validate (returns nothing), and it's confusing when juxtaposed with `ValidateLocal()` ("what validation happens in each?").

`PreValidation` isn't supposed to validate anything. It's supposed to do something before the validation. Maybe it was because of my poor English. I checked the dictionary and it said.

"To prevalidate means to validate or verify something in advance, before it is actually used or implemented"

So you're right. Prevalidate means validate.

haxxpop (approved):
LGTM after all

Issue #611, opened 2025-05-06
Cancel queued `Publish` RPCs if `IDONTWANT` arrives in the meantime


raulk:
See https://github.com/libp2p/rust-libp2p/issues/5751 and https://github.com/sigp/rust-libp2p/pull/570.

MarcoPolo:
We effectively do this already [here](https://github.com/MarcoPolo/go-libp2p-pubsub/blob/master/gossipsub.go#L1199-L1207) by checking if the message is unwanted right before sending it.

raulk:
`sendRPC` only _queues_ the RPC for sending. The actual sending happens async via a dispatch goroutine a few layers down the stack. Between now and then we could've received an `IDONTWANT`. That's the cancellation I was referring to here. Some useful breadcrumbs:

- [`doSendRPC`](https://github.com/libp2p/go-libp2p-pubsub/blob/ff109a6a4240dec8dd9b81f1f310cd85b6afd30e/gossipsub.go#L1396)
- [`rpcQueue#push`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/rpc_queue.go#L67)
- [`push`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/rpc_queue.go#L75)
- [`priorityQueue#NormalPush`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/rpc_queue.go#L25)

One way to approach this is by adding an `RPC#Cancelled() bool` that is backed by a closure supplied at RPC creation time that queries the `gs.unwanted` map, but we'd need to guard that with a lock, break the concurrency model, and introduce contention which ain't good. There are more ways.

MarcoPolo:
This is likely a difference of <1ms for a machine not in high cpu contention. Is the juice worth the squeeze? We can certainly try.

MarcoPolo:
I'd be curious to look at our trace metrics and see the time differences between when we send the rpc (in doSendRPC) and when we receive a IDONTWANT for that rpc, and how many times these are within 1 ms of each other.

raulk:
I don't think this is CPU-bound? I was thinking it would be IO bound by the multiplexer or transport congestion control (i.e. if the Yamux window is full), which can happen more frequently under bandwidth saturation (especially under bursty loads).

MarcoPolo:
ah right, if the transport backpressures on stream writes. That seems more likely 👍

haxxpop:
[...]

PR #613, opened 2025-05-07
Check peer's IDONTWANT state right before sending RPC

MarcoPolo:
Related to #611.

A bit hacky, and there's a couple of footguns with the various RPC copies.

It would be nice to use a context for this, but it doesn't work well because the send is async so we'd have to keep the context around long enough for the send to finish and then clean it up.

Having the rpc check a map behind a lock is one of the cheapest options.

A slight improvement here would be to send the cancelled function separately from the RPC at the expense of more code changes.

Another approach is to change the priority queue to be aware of cancellation, but I think this looks a lot like this approach.

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -199,6 +199,13 @@ func rpcWithSubs(subs ...*pb.RPC_SubOpts) *RPC {
 	}
 }
 
+func rpcWithMessagesAndChecksums(msgs []*pb.Message, checksums []checksum) *RPC {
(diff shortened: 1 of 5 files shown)


MarcoPolo:
This needs to be careful to not drop piggybacked messages (or at least re-queue them)

MarcoPolo:
Before merging, we should add metrics to this so we can measure how effective this is.

PR #582, opened 2024-10-01
Fix appendOrMergeRPC inefficiency in message size recalculation

algorandskiy:
## Summary

As discussed in #581 there an inefficiency in `appendOrMergeRPC` in calling `Size()` more times than needed.

## Fix

Instead of calling `lastRPC.Size()` that iterates over `RPC.Publish`, save the last known size and sum with a current message size and protobuf upper bound overhead.

## Status and Evaluation

Benchmark results
```
[master●●] % go test ./ -run ^$ -bench BenchmarkAppendOrMergeRPC -benchmem -count=10 > old.txt
[pavel/append-or-merge●] % go test ./ -run ^$ -bench BenchmarkAppendOrMergeRPC -benchmem -count=10 > new.txt
[pavel/append-or-merge●] % benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/libp2p/go-libp2p-pubsub
cpu: Apple M1 Pro
                          │   old.txt   │               new.txt               │
                          │   sec/op    │   sec/op     vs base                │
AppendOrMergeRPC/small-10   741.8n ± 0%   280.8n ± 0%  -62.15% (p=0.000 n=10)
AppendOrMergeRPC/large-10   36.88µ ± 0%   10.67µ ± 0%  -71.06% (p=0.000 n=10)
geomean                     5.231µ        1.731µ       -66.90%

                          │   old.txt    │                new.txt                │
                          │     B/op     │     B/op      vs base                 │
AppendOrMergeRPC/small-10     368.0 ± 0%     368.0 ± 0%       ~ (p=1.000 n=10) ¹
AppendOrMergeRPC/large-10   18.54Ki ± 0%   18.54Ki ± 0%       ~ (p=1.000 n=10) ¹
geomean                     2.581Ki        2.581Ki       +0.00%
¹ all samples are equal

                          │  old.txt   │               new.txt               │
                          │ allocs/op  │ allocs/op   vs base                 │
AppendOrMergeRPC/small-10   7.000 ± 0%   7.000 ± 0%       ~ (p=1.000 n=10) ¹
AppendOrMergeRPC/large-10   216.0 ± 0%   216.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                     38.88        38.88       +0.00%
¹ all samples are equal
```

PR #615, opened 2025-05-23
refactor: 10x faster RPC splitting

MarcoPolo:
Builds on #582. 10x faster than current master. 0 allocs.

The basic logic is the same as the old version, except we return an `iter.Seq[RPC]` and yield `RPC` types instead of a slice of `*RPC`. This lets us avoid allocations for heap pointers.

Please review @algorandskiy, and let me know if this improves your use case.


gossipsub.go
```diff
@@ -1412,135 +1412,182 @@ func (gs *GossipSubRouter) doSendRPC(rpc *RPC, p peer.ID, q *rpcQueue, urgent bo
...
-
-	out := slice
-	if len(out) == 0 {
-		out = append(out, &RPC{RPC: pb.RPC{}})
-		out[0].from = elems[0].from
-	}
+// split splits the given RPC If a sub RPC is too large and can't be split
+// further (e.g. Message data is bigger than the RPC limit), then it will be
+// returned as an oversized RPC. The caller should filter out oversized RPCs.
+func (rpc *RPC) split(limit int) iter.Seq[RPC] {
```
MarcoPolo: This method should likely be in the `pubsub.go` file instead. I kept it here to make reviewing easier, but would like to move it after reviewer's approval.

MarcoPolo:
benchstat:

```
goos: darwin
goarch: arm64
pkg: github.com/libp2p/go-libp2p-pubsub
cpu: Apple M1 Max
                                             │   old.txt    │               new.txt               │
                                             │    sec/op    │   sec/op     vs base                │
SplitRPCLargeMessages/Many_large_messages-10   51.751µ ± 0%   4.327µ ± 1%  -91.64% (p=0.000 n=10)
SplitRPCLargeMessages/2_large_messages-10      320.70n ± 2%   34.63n ± 0%  -89.20% (p=0.000 n=10)
geomean                                         4.074µ        387.0n       -90.50%

                                             │   old.txt    │                 new.txt                  │
                                             │     B/op     │    B/op      vs base                     │
SplitRPCLargeMessages/Many_large_messages-10   18.65Ki ± 0%   0.00Ki ± 0%  -100.00% (p=0.000 n=10)
SplitRPCLargeMessages/2_large_messages-10        392.0 ± 0%      0.0 ± 0%  -100.00% (p=0.000 n=10)
geomean                                        2.672Ki                     ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean

                                             │  old.txt   │                 new.txt                 │
                                             │ allocs/op  │ allocs/op   vs base                     │
SplitRPCLargeMessages/Many_large_messages-10   217.0 ± 0%     0.0 ± 0%  -100.00% (p=0.000 n=10)
SplitRPCLargeMessages/2_large_messages-10      8.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
geomean                                        41.67                    ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

gossipsub.go
```diff
@@ -1412,135 +1412,182 @@ func (gs *GossipSubRouter) doSendRPC(rpc *RPC, p peer.ID, q *rpcQueue, urgent bo
...
+
+		if nextRPC.Size() > 0 {
+			if !yield(nextRPC) {
+				return
+			}
+		}
 	}
+}
+
+// pbFieldNumberLT15Size is the number of bytes required to encode a protobuf
```
algorandskiy: maybe it makes sense to add a reference to pb.RPC.Size() to let a reader to find the original `1 (pbFieldNumberLT15Size) + l + sov` formula

gossipsub.go
```diff
@@ -1412,135 +1412,182 @@ func (gs *GossipSubRouter) doSendRPC(rpc *RPC, p peer.ID, q *rpcQueue, urgent bo
...
 
 		// Merge/Append Subscriptions
-		for _, sub := range elem.GetSubscriptions() {
-			if lastRPC.Subscriptions = append(lastRPC.Subscriptions, sub); lastRPC.Size() > limit {
-				lastRPC.Subscriptions = lastRPC.Subscriptions[:len(lastRPC.Subscriptions)-1]
-				lastRPC = &RPC{RPC: pb.RPC{}, from: elem.from}
-				lastRPC.Subscriptions = append(lastRPC.Subscriptions, sub)
-				out = append(out, lastRPC)
+		for _, sub := range rpc.Subscriptions {
+			if nextRPC.Subscriptions = append(nextRPC.Subscriptions, sub); nextRPC.Size() > limit {
```
algorandskiy: `nextRPC.Size()` - this is the same issue with calling Size in a loop over and over again as for Publish but probably does not matter much as (from my understanding) new subscriptions as well as control are relatively rare compare to Publish thing

algorandskiy (approved):
Looks great and impressive +90% perf gain comparing to mine feeb +66% :)

gossipsub.go
MarcoPolo: Yes exactly. Notice that we flush the messages to publish above. So here we aren't calling .Size on Publish messages again.

Splitting due to control messages should be extremely rare (happy to be proven wrong), so I opted to do the easier thing and keep the old code.

Thinking about this just now, we could do a fast path check before these loops that check if the original RPC without the publish messages is small enough to fit and can be yielded.

MarcoPolo: Added this, and it adds a nice small improvement. Many large benchmark is around 3.3 us, and 2 large is around 28 ns.

PR #622, opened 2025-06-24
fix(BatchPublishing): Make topic.AddToBatch threadsafe

MarcoPolo:
topic.Publish is already thread safe. topic.AddToBatch should strive to follow similar semantics.

Looking at how this would integrate with Prysm, they use separate goroutines per message they'd like to batch.

### DIFF
--- a/messagebatch.go
+++ b/messagebatch.go
@@ -2,6 +2,7 @@ package pubsub
 
 import (
 	"iter"
+	"sync"
 
 	"github.com/libp2p/go-libp2p/core/peer"
 )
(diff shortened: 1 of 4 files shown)


messagebatch.go
```diff
@@ -10,9 +11,24 @@ import (
...
+	mb.messages = append(mb.messages, msg)
+}
+
+func (mb *MessageBatch) takeMessages() []*Message {
+	mb.mu.Lock()
+	defer mb.mu.Unlock()
+	messages := mb.messages
+	mb.messages = nil
+	return messages
+}
```
sukunrt: NIT: I prefer just `add` and `take`

sukunrt (approved):

PR #625, opened 2025-07-02
Limit the number of outstanding IWANTs per messageid and per heartbeat.

### LINKED ISSUE #606: Sending only one IWANT at a time
From what I see in the function `handleIhave` in gossipsub.go
https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L757

The logic is that when you receive IHAVE, you check if you have seen the message already or not. If not, just send IWANT right away.

This makes it possible that you will send many IWANTs for a single message which results to receiving many copies of the message.

AFAIU currently rust-libp2p sends only one IWANT per message so that it doesn't receive too many copies. Should go-libp2p-pubsub do the same?

MarcoPolo:
This fixes an issue where a node would end up making an IWANT request each time a peer advertised a message with IHAVE, resulting in many duplicates for the requesting peer.

The change is to limit the number of IWANT requests per message id, and to reset this limit every heartbeat. This allows fewer reqeusts

Fixes #606

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -3860,3 +3860,94 @@ func BenchmarkSplitRPCLargeMessages(b *testing.B) {
 		}
 	})
 }
+
+func TestGossipsubLimitIWANT(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	hosts := getDefaultHosts(t, 3)
+	denseConnect(t, hosts)
+
+	psubs := make([]*PubSub, 3)
+
+	defaultParams := DefaultGossipSubParams()
+	defaultParams.MaxIWantsPerMessageIDPerHeartbeat = 1
+	// Delay the heartbeat so we don't reset our count of allowed IWANTS for the
+	// first part of the test.
+	defaultParams.HeartbeatInitialDelay = 4 * time.Second
+	defaultParams.HeartbeatInterval = time.Second
+
+	psubs[0] = getGossipsub(ctx, hosts[0], WithGossipSubParams(defaultParams))
+
+	var iwantsRecvd atomic.Int32
+
+	copy(psubs[1:], getGossipsubs(ctx, hosts[1:], WithRawTracer(&mockRawTracer{
+		onRecvRPC: func(rpc *RPC) {
+			if len(rpc.GetControl().GetIwant()) > 0 {
+				iwantsRecvd.Add(1)
+			}
+		},
+	})))
+
+	topicString := "foobar"
+	for _, ps := range psubs {
+		topic, err := ps.Join(topicString)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = topic.Subscribe()
+		if err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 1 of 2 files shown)


gossipsub_test.go
```diff
@@ -3860,3 +3860,94 @@ func BenchmarkSplitRPCLargeMessages(b *testing.B) {
...
+	})))
+
+	topicString := "foobar"
+	for _, ps := range psubs {
+		_, err := ps.Join(topicString)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = ps.Subscribe(topicString)
```
haxxpop:
```suggestion
		topic, err := ps.Join(topicString)
		if err != nil {
			t.Fatal(err)
		}

		_, err = topic.Subscribe(topicString)
```
I think PubSub.Subscribe is deprecated. It's probably better to be compliant with the current norm.

haxxpop (approved):
Looking good overall

cskiraly:
This is largely needed based on my measurements (also based on finding the same gap and implementing it in nim-libp2p in the past). However, I would use a slightly different shape, with a refill window instead of a heartbeat based timer. 

Details follow:
We ran a measured comparison of this PR's rule while studying segmented 1 MiB payload diffusion (32 KiB segments as individual messages, announce-and-pull-heavy workloads) on a 500-node simulated network — real go-libp2p-pubsub over simulated QUIC links (`x/simlibp2p` + synctest). Three findings that may be useful here:

1. **The fix matters a lot for large-message workloads.** At limit 1, duplicate segment receptions network-wide dropped from ~36,500 (uncapped) to ~1,500, with completion time unchanged.
2. **The default of 10 was a no-op in our setting** (connectivity degree 20, mesh D=8): 32,479 duplicates vs 36,556 uncapped, latency identical. Worth a lower default, or a note that large-message topics want ~1.
3. **The per-heartbeat wholesale reset leaves a boundary artifact**: a request still in flight when a heartbeat fires gets its budget refreshed, admitting a second request; and an unanswered request cannot retry until the next boundary. Refilling **per entry** instead removed another order of magnitude — ~90–160 duplicates vs ~1,500 at limit 1, across three topology seeds of paired runs (common random numbers). Completion time was equal within realization noise in every configuration, so the win is purely duplicates.

Branch with the per-entry variant on top of this PR: https://github.com/cskiraly/go-libp2p-pubsub/tree/iwant-per-entry-window — it reuses `IWantFollowupTime` as the refill deadline (the existing definition of "this IWANT failed", so no new parameter; our measurements used a 400 ms window, the mechanism is the same) and renames the param since it's no longer per-heartbeat. Happy to adapt however is most useful: fold into this PR (feel free to cherry-pick it), a follow-up PR after this lands, or drop the rename.

sukunrt:
@cskiraly 

>The default of 10 was a no-op in our setting (connectivity degree 20, mesh D=8): 32,479 duplicates vs 36,556 uncapped, latency identical. Worth a lower default, or a note that large-message topics want ~1.

Do you have a suggestion for a lower default?

sukunrt (approved):

PR #630, opened 2025-07-09
Gossipsub Extensions

MarcoPolo:
Before Merge:
- [x] https://github.com/libp2p/specs/issues/687. Come to rough consensus on having Gossipsub v1.3 only add support for the extensions control message. 

An alternative would be to rely on peers ignoring protobuf messages they don't understand. It seems better to explicitly signal support for the Extensions control Message and only send it if you know your peer supports it. I'll create a new specs issue/PR to discuss this.

---

This implements the Gossipsub Extensions Control Message from https://github.com/libp2p/specs/pull/684. It also implements the Test Extension https://github.com/libp2p/specs/pull/686 as an interop test.

There is a rust-libp2p implementation, and using the gossip-interop tester, I've tested interop between these implementations for the Test Extension.

Don't let the diff stats scare you, they are mostly generated protobuf code.

### DIFF
--- a/extensions.go
+++ b/extensions.go
@@ -1,0 +1,145 @@
+package pubsub
+
+import (
+	pubsub_pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type PeerExtensions struct {
+	TestExtension bool
+}
+
+type TestExtensionConfig struct {
+	OnReceiveTestExtension func(from peer.ID)
+}
+
+func WithTestExtension(c TestExtensionConfig) Option {
+	return func(ps *PubSub) error {
+		if rt, ok := ps.rt.(*GossipSubRouter); ok {
+			rt.extensions.testExtension = &testExtension{
+				sendRPC:                rt.extensions.sendRPC,
+				onReceiveTestExtension: c.OnReceiveTestExtension,
+			}
+			rt.extensions.myExtensions.TestExtension = true
+		}
+		return nil
+	}
+}
+
+func hasPeerExtensions(rpc *RPC) bool {
+	if rpc != nil && rpc.Control != nil && rpc.Control.Extensions != nil {
+		return true
+	}
+	return false
+}
+
+func peerExtensionsFromRPC(rpc *RPC) PeerExtensions {
+	out := PeerExtensions{}
+	if hasPeerExtensions(rpc) {
+		out.TestExtension = rpc.Control.Extensions.GetTestExtension()
+	}
+	return out
+}
+
+func (pe *PeerExtensions) ExtendRPC(rpc *RPC) *RPC {
+	if pe.TestExtension {
+		if rpc.Control == nil {
+			rpc.Control = &pubsub_pb.ControlMessage{}
+		}
+		rpc.Control.Extensions = &pubsub_pb.ControlExtensions{
+			TestExtension: &pe.TestExtension,
+		}
+	}
+	return rpc
+}
+
+type extensionsState struct {
+	myExtensions      PeerExtensions
+	peerExtensions    map[peer.ID]PeerExtensions // peer's extensions
+	sentExtensions    map[peer.ID]struct{}
+	reportMisbehavior func(peer.ID)
+	sendRPC           func(p peer.ID, r *RPC, urgent bool)
+
+	testExtension *testExtension
+}
+
+func newExtensionsState(myExtensions PeerExtensions, reportMisbehavior func(peer.ID), sendRPC func(peer.ID, *RPC, bool)) *extensionsState {
+	return &extensionsState{
+		myExtensions:      myExtensions,
+		peerExtensions:    make(map[peer.ID]PeerExtensions),
+		sentExtensions:    make(map[peer.ID]struct{}),
+		reportMisbehavior: reportMisbehavior,
+		sendRPC:           sendRPC,
+		testExtension:     nil,
+	}
+}
+
+func (es *extensionsState) HandleRPC(rpc *RPC) {
+	if _, ok := es.peerExtensions[rpc.from]; !ok {
+		// We know this is the first message because we didn't have extensions
+		// for this peer, and we always set extensions on the first rpc.
+		es.peerExtensions[rpc.from] = peerExtensionsFromRPC(rpc)
+		if _, ok := es.sentExtensions[rpc.from]; ok {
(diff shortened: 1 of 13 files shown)


pubsub.go
```diff
@@ -800,6 +800,7 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			}
 
 			p.rt.AddPeer(pid, s.Protocol())
+			q.Push(p.getHelloPacket(), true)
```
MarcoPolo: Would be nice to consolidate the hello packet and the extensions control message.

gossipsub_feat_test.go
```diff
@@ -50,6 +56,9 @@ func TestDefaultGossipSubFeatures(t *testing.T) {
 	if !GossipSubDefaultFeatures(GossipSubFeatureIdontwant, GossipSubID_v12) {
 		t.Fatal("gossipsub-v1.2 should support IDONTWANT")
 	}
+	if !GossipSubDefaultFeatures(GossipSubFeatureIdontwant, GossipSubID_v13) {
+		t.Fatal("gossipsub-v1.3 should support IDONTWANT")
+	}
```
haxxpop, line 61:
```suggestion
	}

	if GossipSubDefaultFeatures(GossipSubFeatureExtensions, FloodSubID) {
		t.Fatal("floodsub should not support Extensions")
	}
	if GossipSubDefaultFeatures(GossipSubFeatureExtensions, GossipSubID_v10) {
		t.Fatal("gossipsub-v1.0 should not support Extensions")
	}
	if GossipSubDefaultFeatures(GossipSubFeatureExtensions, GossipSubID_v11) {
		t.Fatal("gossipsub-v1.1 should not support Extensions")
	}
	if GossipSubDefaultFeatures(GossipSubFeatureExtensions, GossipSubID_v12) {
		t.Fatal("gossipsub-v1.2 should not support Extensions")
	}
	if !GossipSubDefaultFeatures(GossipSubFeatureExtensions, GossipSubID_v13) {
		t.Fatal("gossipsub-v1.3 should support Extensions")
	}
```

pubsub.go
```diff
@@ -194,8 +194,11 @@ type PubSubRouter interface {
 	// Attach is invoked by the PubSub constructor to attach the router to a
 	// freshly initialized PubSub instance.
 	Attach(*PubSub)
-	// AddPeer notifies the router that a new peer has been connected.
-	AddPeer(peer.ID, protocol.ID)
+	// AddPeer notifies the router that a new peer has been connected. It
+	// includes a reference to the initial RPC that will be sent to the peer.
+	// Routers may add messages to the RPC to try to have them sent in the
+	// initial packet. This is also referred to as the "hello packet."
+	AddPeer(peer.ID, protocol.ID, *RPC)
```
haxxpop: Making a param an output is very weird. What I usually expect from a function is that the params are inputs and the returned value is an output.

What you can do with this function is that you can return the `*pb.ControlExtensions` instead which I think is cleaner and you don't have to worry about mutating the parameter.

```suggestion
	AddPeer(peer.ID, protocol.ID) *pb.ControlExtensions
```

pubsub.go
```diff
@@ -799,7 +802,9 @@ func (p *PubSub) processLoop(ctx context.Context) {
 				continue
 			}
 
-			p.rt.AddPeer(pid, s.Protocol())
+			helloPacket := p.getHelloPacket()
+			p.rt.AddPeer(pid, s.Protocol(), helloPacket)
```
haxxpop: After the change I mentioned in https://github.com/libp2p/go-libp2p-pubsub/pull/630/files#r2306618920, this can be
```suggestion
			if extensions := p.rt.AddPeer(pid, s.Protocol()); extensions != nil {
						if helloPacket.Control == nil {
									helloPacket.Control = &pb.ControlMessage{}
						}
						hellPacket.Control.Extensions = extensions
			}
```

extensions.go
```diff
@@ -0,0 +1,131 @@
...
+		peerExtensions:    make(map[peer.ID]PeerExtensions),
+		sentExtensions:    make(map[peer.ID]struct{}),
+		reportMisbehavior: reportMisbehavior,
+		sendRPC:           sendRPC,
+		testExtension:     testExtension{sendRPC: sendRPC},
+	}
+}
+
+func (es *extensionsState) HandleRPC(rpc *RPC) {
+	if _, ok := es.peerExtensions[rpc.from]; !ok {
```
haxxpop, line 78: `es.peerExtensions` and `es.sentExtensions` can be accessed by both `HandleRPC` and `AddPeer` at the same time. Maybe we should have some mutex?

testextension.go
```diff
@@ -0,0 +1,18 @@
...
+	pubsub_pb "github.com/libp2p/go-libp2p-pubsub/pb"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type testExtension struct {
+	sendRPC func(p peer.ID, r *RPC, urgent bool)
+}
+
+func (e *testExtension) AddPeer(id peer.ID) {
+	e.sendRPC(id, &RPC{
```
haxxpop, line 14: Since you send an RPC here, should you read it from your peer as well?

extensions.go
MarcoPolo: This is guaranteed to only be called by a single goroutine. No mutex needed.

testextension.go
MarcoPolo: Yes. Might be easier to test as well. thanks!

pubsub.go
MarcoPolo: I tried this out, and I think it's worse than the out param. It leads to all pubsub implementations having to know about control extensions in their signatures. Also there's no particular reason to only allow the ControlExtensions besides it's the only thing we bundle in the first message. Passing the out param is more general.

Instead I'll follow the append style where I return a (possibly new) pointer.

MarcoPolo:
Just a friendly reminder that you can view the diffs between versions of these commits at:
https://github.com/libp2p/go-libp2p-pubsub/compare/222428a77e4557f1097e6f0879539ac956760c8f..5299023741423ee771b0db79e902cdf991cd769f#diff-c5386f794d081a791f3a2cbe234dcb5c6452209e916a5869117073c4f8fa8d6c

or (preferred) by locally running:

```
git range-diff origin/master 222428a 5299023
```

The latter is preferred as it also tells you which commit introduced the change.

haxxpop (approved):
Looking good to me

PR #631, opened 2025-07-15
Partial Messages Extension

MarcoPolo:
Implements https://github.com/libp2p/specs/pull/685

Commits are meaningful. Best to review commit by commit.


extensions.go
```diff
@@ -50,6 +56,14 @@ func (pe *PeerExtensions) ExtendRPC(rpc *RPC) *RPC {
 			TestExtension: &pe.TestExtension,
 		}
 	}
+	if pe.PartialMessages {
+		if rpc.Control == nil {
+			rpc.Control = &pubsub_pb.ControlMessage{}
+		}
+		rpc.Control.Extensions = &pubsub_pb.ControlExtensions{
```
haxxpop: I think this removes the previously added `TestExtension` field above because this creates a new instance of `ControlExtensions`.

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,395 @@
...
+		partialMessageBytes []byte,
+	)
+
+	// ValidateRequestMetadata should be a fast function that performs some
+	// basic sanity checks on incoming RPC. For example:
+	//   - Is this a known topic?
+	//   - Is the groupID well formed per application semantics?
+	//   - If this is a PartialIHAVE/PartialIWant, is the request metadata within
+	//     expected bounds?
+	ValidateRPC func(from peer.ID, rpc *pb.PartialMessagesExtension) error
```
haxxpop, line 118: This can be nil and this file assumes that this is not nil. The problem is that the users can create an instance of this struct with this to be nil and don't know that it's wrong.

I think we should have a constructor for this struct (`NewPartialMessageExtension`) and emit an error if this is nil or just assign this to a default value.

extensions.go
MarcoPolo: yes, thank you. It was an incorrect conflict resolution

partialmessages/partialmsgs.go
MarcoPolo: This shouldn't be nil. It would be a user error to set this to nil and be detected as soon as you receive the first message.

However, an explicit error is nice. I don't want a needless constructor when we don't need one. Instead I'll add a .Validate method, and have that called in `WithPartialMessagesExtension`

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,392 @@
...
+		e.statePerTopicPerGroup = make(map[string]map[string]*partialMessageStatePerTopicGroup)
+	}
+	statePerTopic, ok := e.statePerTopicPerGroup[topic]
+	if !ok {
+		statePerTopic = make(map[string]*partialMessageStatePerTopicGroup)
+		e.statePerTopicPerGroup[topic] = statePerTopic
+	}
+	state, ok := statePerTopic[string(groupID)]
+	if !ok {
+		state = newPartialMessageStatePerTopicGroup(max(e.GroupTTLByHeatbeat, minGroupTTL))
```
haxxpop: max is already done in `newPartialMessageStatePerTopicGroup`
```suggestion
		state = newPartialMessageStatePerTopicGroup(e.GroupTTLByHeatbeat)
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,392 @@
...
+
+			log.Debug("Respond to peer's IWant")
+			sendRPC = true
+			rpc.Message = &pb.PartialMessage{
+				Data: pm,
+			}
+			inResponseToIWant = true
+		}
+
+		if !inResponseToIWant && eagerPush != nil && !bytes.Equal(ihave, peerState.has) {
```
haxxpop: I don't quite understand why do you need this condition? `!bytes.Equal(ihave, peerState.has) `. `eagerPush` is not part of the partial message, right?

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,392 @@
...
+	// message contains. Used for PartialIHAVE
+	AvailableParts() ([]byte, error)
+}
+
+type partialMessagePeerState struct {
+	wants    []byte
+	has      []byte
+	sentWant []byte
+	sentHas  []byte
+}
```
haxxpop: Could you please explain all the fields as comments here? I think it would be helpful while I read the code.

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,392 @@
...
+		if !inResponseToIWant && eagerPush != nil && !bytes.Equal(ihave, peerState.has) {
+			sendRPC = true
+			log.Debug("Eager pushing")
+			rpc.Message = eagerPush
+		}
+
+		// Only send IHAVE if they don't have what we have and they haven't sent
+		// us an IDONTWANT
+		if !inResponseToIWant && partialIHAVE != nil {
+			if !peerStateOk || !bytes.Equal(ihave, peerState.sentHas) {
```
haxxpop: As commented here
```
		// Only send IHAVE if they don't have what we have and they haven't sent
		// us an IDONTWANT
```
I think this uses a wrong field. Can you confirm?
```suggestion
			if !peerStateOk || !bytes.Equal(ihave, peerState.has) {
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,392 @@
...
+				sendRPC = true
+				peerState.sentHas = ihave
+				log.Debug("Including IHAVE")
+				rpc.Ihave = partialIHAVE
+			}
+		}
+
+		// Only send an IWANT if it was different then before and we haven't exceeded eager limit
```
haxxpop: "and we haven't exceeded eager limit". I can't see where is the eager limit. Can you point it out please?

partialmessages/partialmsgs.go
MarcoPolo: The comment was wrong. Updated.

partialmessages/partialmsgs.go
MarcoPolo: added a comment to explain

partialmessages/partialmsgs.go
MarcoPolo: stale comment, thanks

MarcoPolo:
thanks for the review @ppopth, friendly reminder you can see the latest changes with `git range-diff master 7fc7eb495bd6 578e12f4afcd`

haxxpop (approved):
Already looking good to me

pubsub.go
```diff
@@ -262,6 +262,35 @@ type RPC struct {
 	from peer.ID
 }
 
+// LogValue implements slog.LogValuer.
+func (rpc *RPC) LogValue() slog.Value {
+	// Messages
+	var msgs []any
```
sukunrt: Can prealloc to len(rpc.Publish)

pubsub.go
```diff
@@ -262,6 +262,35 @@ type RPC struct {
...
+	for _, msg := range rpc.Publish {
+		msgs = append(msgs, slog.Group(
+			"message",
+			slog.Any("topic", msg.Topic),
+			slog.Any("dataPrefix", msg.Data[0:min(len(msg.Data), 32)]),
+			slog.Any("dataLen", len(msg.Data)),
+		))
+	}
+
+	var fields []slog.Attr
```
sukunrt: Can prealloc.

extensions.go
```diff
@@ -132,14 +149,79 @@ func (es *extensionsState) extensionsAddPeer(id peer.ID) {
...
+
+		gs.extensions.myExtensions.PartialMessages = true
+		gs.extensions.partialMessagesExtension = pm
+		return nil
+	}
+}
+
+type routerForPartialMessage struct {
+	gs *GossipSubRouter
+}
```
sukunrt: NIT: I prefer `PartialMessageRouter`

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+	// partial message from a peer. This func MUST be fast and non-blocking.
+	// If you need to do slow work, dispatch the work to your own goroutine.
+	OnIncomingRPC func(
+		from peer.ID,
+		topic string,
+		groupID []byte,
+		iwant []byte,
+		ihave []byte,
+		partialMessageBytes []byte,
+	)
```
sukunrt: Should we just make it `OnIncomingRPC(from, rpc)`. The user anyway needs to deal with the RPC because of `ValidateRPC`. Also seems like any changes to the RPC will be bubbled up here.

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+		var sendRPC bool
+		var inResponseToIWant bool
+
+		peerState, peerStateOk := state.peerState[p]
+		if !peerStateOk {
+			peerState = &partialMessagePeerState{}
+			state.peerState[p] = peerState
+		}
+		// Try to fulfill any outstanding PartialIWants
+		if peerStateOk && len(peerState.wants) > 0 {
```
sukunrt: NIT: `peerStateOk` check is redundant

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+			// This peer has previously asked for a certain part. We'll give
+			// them what we can.
+			pm, rest, err := partial.PartialMessageBytesFromMetadata(peerState.wants)
+			if err != nil {
+				if e.Logger != nil {
+					e.Logger.Warn("partial message extension failed to get partial message bytes", "error", err)
+				}
+				// Possibly a bad request, we'll delete the request as we will likely error next time we try to handle it
```
sukunrt: Should we log this deletion? A hexdump of the metadata maybe?

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+		// Only send the eager push to the peer if:
+		//   - we aren't explicitly replying to an earlier request
+		//   - we have something to eager push
+		//   - we don't have the same parts (implies they already have the thing
+		//     we are eager pushing)
+		if !inResponseToIWant && eagerPush != nil && !bytes.Equal(ihave, peerState.has) {
+			sendRPC = true
+			log.Debug("Eager pushing")
+			rpc.Message = eagerPush
+		}
```
sukunrt: Should this be another callback on PartialMessage? the bytes.Equal check isn't sufficient, we need a subset check. You have part 1 and peer has parts 1, 2, 3, 4. You'll still eager push, right?

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+	// returns a encoded partial message that fulfills as much of the request as
+	// possible. It also returns a opaque request metadata representing the
+	// parts it could not fulfill. This MUST be empty if the implementation could
+	// fulfill the whole request.
+	//
+	// An empty metadata should be treated the same as a request for all parts.
+	//
+	// If the Partial Message is empty, the implementation MUST return:
+	// nil, metadata, nil.
+	PartialMessageBytesFromMetadata(metadata []byte) ([]byte, []byte, error)
```
sukunrt: I think we should just name this IWant: Makes it clearer. 
```
PartialMessageBytesFromIWantRepr(iwantrepr []byte) (partialMsgs []byte, newIWantRepr []byte, err error)
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+	// Return an empty slice if you have all available parts.
+	MissingParts() ([]byte, error)
+
+	// ShouldRequest returns true if the given metadata has parts that this
+	// partial message is missing. Used when handling PartialIHAVE.
+	ShouldRequest(from peer.ID, peerHasMetadata []byte) bool
+
+	// AvailableParts returns the opaque metadata representing the parts this
+	// message contains. Used for PartialIHAVE
+	AvailableParts() ([]byte, error)
```
```suggestion
	// AvailableParts returns the opaque metadata representing the parts this
	// message contains. Used for PartialIHAVE
	AvailableParts() (iHaveRepr []byte, error)
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+
+// partialMessagePeerState is the state we keep per peer. Used to make Publish
+// Idempotent.
+type partialMessagePeerState struct {
+	// What parts the peer wants
+	wants []byte
+	// What parts the peer has
+	has []byte
+	// The wants this endpoint has sent to the peer
+	sentWant []byte
```
```suggestion
	// The wants we have sent to the peer
	sentWant []byte
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+	// The has this endpoint has sent to the peer
+	sentHas []byte
```
```suggestion
	// The has we have sent to the peer
	sentHas []byte
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,402 @@
...
+	// An empty metadata should be treated the same as a request for all parts.
+	//
+	// If the Partial Message is empty, the implementation MUST return:
+	// nil, metadata, nil.
+	PartialMessageBytesFromMetadata(metadata []byte) ([]byte, []byte, error)
+
+	// MissingParts returns the opaque request metadata representing the missing
+	// parts. Used for PartialIWANT
+	// Return an empty slice if you have all available parts.
+	MissingParts() ([]byte, error)
```
```suggestion
	// MissingParts returns the opaque request metadata representing the missing
	// parts. Used for PartialIWANT
	// Return an empty slice if you have all available parts.
	MissingParts() (iWantRepr []byte, err error)
```

gossipsub_test.go
```diff
@@ -4394,3 +4397,396 @@ func TestTestExtension(t *testing.T) {
...
+	}
+	checker := &minimalTestPartialMessageChecker{
+		fullMessage: full,
+	}
+	partialmessages.TestPartialMessageInvariants(t, checker)
+}
+
+func TestPartialMessages(t *testing.T) {
+	topic := "test-topic"
+	const hostCount = 2
```
sukunrt: Should we make this a bit higher, say 5?

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+	// returns a encoded partial message that fulfills as much of the request as
+	// possible. It also returns a opaque request metadata representing the
+	// parts it could not fulfill. This MUST be empty if the implementation could
+	// fulfill the whole request.
+	//
+	// An empty metadata should be treated the same as a request for all parts.
+	//
+	// If the Partial Message is empty, the implementation MUST return:
+	// nil, metadata, nil.
+	PartialMessageBytes(metadata []byte) ([]byte, []byte, error)
```
```suggestion
	PartialMessageBytes(hasPartsMetadata []byte) (msg []byte, newHasPartsMetadata []byte, _ error)
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+// PartialMessage is a message that can be broken up into parts. It can be
+// complete, partially complete, or empty. It is up to the application to define
+// how a message is split into parts and recombined, as well as how missing and
+// available parts are represented.
+//
+// It is passed to Gossipsub with a PublishPartialMessage method call. Gossipsub
+// keeps a reference to this object, so the implementation should either not
+// mutate this object in a separate goroutine after handing it to Gossipsub, or
+// take care to make the object thread safe.
+type PartialMessage interface {
```
sukunrt: NIT: partialmessages is the package name, so we can remove Partial from the type names:

read by users as `partialmessages.Message`

```suggestion
type Message interface {
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+	PartialMessageBytes(metadata []byte) ([]byte, []byte, error)
+
+	// PartsMetadata returns metadata about the parts this partial message
+	// contains and, possibly implicitly, the parts it wants.
+	PartsMetadata() []byte
+}
+
+// partialMessagePeerState is the state we keep per peer. Used to make Publish
+// Idempotent.
+type partialMessagePeerState struct {
```
sukunrt: NIT: for the same reasons as the renaming for PartialMessage. 
```suggestion
type peerState struct {
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+	// The parts metadata the peer has sent us
+	partsMetadata []byte
```
sukunrt: Is this the parts the peer has? If so
partsMetadata -> hasMetadata.

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+
+type PartialMessageExtension struct {
+	Logger *slog.Logger
+
+	// OnIncomingRPC is called whenever we receive an encoded
+	// partial message from a peer. This func MUST be fast and non-blocking.
+	// If you need to do slow work, dispatch the work to your own goroutine.
+	OnIncomingRPC func(from peer.ID, rpc *pb.PartialMessagesExtension) error
+
+	// ValidateRequestMetadata should be a fast function that performs some
```
```suggestion
	// ValidateRPC should be a fast function that performs some
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+
+type Router interface {
+	SendRPC(p peer.ID, r *pb.PartialMessagesExtension, urgent bool)
+	MeshPeers(topic string) iter.Seq[peer.ID]
+}
+
+func (e *PartialMessageExtension) groupState(topic string, groupID []byte) (*partialMessageStatePerTopicGroup, error) {
+	if e.statePerTopicPerGroup == nil {
+		e.statePerTopicPerGroup = make(map[string]map[string]*partialMessageStatePerTopicGroup)
+	}
```
sukunrt: NIT: we can initialize this in Init

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+	// PublishToPeers limits the publishing to only the specified peers.
+	// If nil, will use the topic's mesh peers.
+	PublishToPeers []peer.ID
+	// EagerPush is data that will be eagerly pushed to peers in a PartialMessage
+	EagerPush []byte
+}
+
+type Router interface {
+	SendRPC(p peer.ID, r *pb.PartialMessagesExtension, urgent bool)
+	MeshPeers(topic string) iter.Seq[peer.ID]
```
sukunrt, line 153: We collect the iter to a slice when using this. Maybe just change this to return a slice?

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+
+	state.groupTTL = max(e.GroupTTLByHeatbeat, minGroupTTL)
+
+	peers := opts.PublishToPeers
+	if peers == nil {
+		peers = slices.Collect(e.router.MeshPeers(topic))
+	}
+	e.Logger.Debug("Sending partial messages to peers", "peers", peers, "topic", topic, "group", groupID)
+	for _, p := range peers {
+		log := e.Logger.With("peer", p)
```
sukunrt, line 233: Will this allocate? It is fine as is, just curious.

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+			state.peerState[p] = peerState
+		}
+
+		// Try to fulfill any wants from the peer
+		if peerState.partsMetadata != nil {
+			// This peer has previously asked for a certain part. We'll give
+			// them what we can.
+			pm, rest, err := partial.PartialMessageBytes(peerState.partsMetadata)
+			if err != nil {
+				e.Logger.Warn("partial message extension failed to get partial message bytes", "error", err)
```
```suggestion
				log.Warn("partial message extension failed to get partial message bytes", "error", err)
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+	// ValidateRequestMetadata should be a fast function that performs some
+	// basic sanity checks on incoming RPC. For example:
+	//   - Is this a known topic?
+	//   - Is the groupID well formed per application semantics?
+	//   - If this is a PartialIHAVE/PartialIWant, is the request metadata within
+	//     expected bounds?
+	ValidateRPC func(from peer.ID, rpc *pb.PartialMessagesExtension) error
+
+	// GroupTTLByHeatbeat is how many heartbeats we store Group state for.
+	GroupTTLByHeatbeat int
```
```suggestion
	// GroupTTLByHeatbeat is how many heartbeats we store Group state for after publishing a partial message for the group.
	GroupTTLByHeatbeat int
```

partialmessages/partialmsgs.go
```diff
@@ -0,0 +1,290 @@
...
+		return err
+	}
+
+	if rpc.PartsMetadata != nil {
+		peerState, ok := state.peerState[from]
+		if !ok {
+			peerState = &partialMessagePeerState{}
+			state.peerState[from] = peerState
+		}
+		peerState.partsMetadata = rpc.PartsMetadata
```
sukunrt: This naming is confusing. Can we make it either `HasMetadata` or `WantsMetadata`

sukunrt: Note: The state we store is what the peer *HAS* but some of the code reads like this is what the peer *WANTS*

sukunrt: Maybe we should let the application set this. 

This update looks like a timing issue. 

Say: We have parts 1, 2, 5, 6 
1. Peer sends a message saying I have part 1, 2
2. We send parts 5, 6 and update the peerstate in `PublishMessage` to (1, 2, 5, 6)
3. Concurrently with 2, we receive message from the peer saying I now have parts (1, 2, 3, 4)

Now the new state may override the previous state. This will result in duplicate sending of parts 5, 6 on next publishPartial if we have any new part to send.

sukunrt: Another way I can think of is to have another field for the peer which says 
`sentParts` and change this 
```
PartialMessageBytes(metadata []byte) ([]byte, []byte, error)
```
to
```
PartialMessageBytes(peerHasMetadata []byte, weSentMetadata []byte) ([]byte, newWeSentMetadata []byte, error)
```

We can then remove the second return value and just set it to `myPartsMeta`.

partialmessages/partialmsgs.go
MarcoPolo: I'll make this change. Though it's just partsMetadata as the meaning is left to the application.

partialmessages/partialmsgs.go
MarcoPolo: Not necessarily. It can be used that way, but it could also be the inverse, a bloom filter or something else.

partialmessages/partialmsgs.go
MarcoPolo: the slices collect was for debugging. I can remove that.

partialmessages/partialmsgs.go
MarcoPolo: Yes, it does. Although I don't think it will stringify p until necessary.

partialmessages/partialmsgs.go
MarcoPolo: The code here assigns no meaning to this. The application decides if this is has or wants.

Open to a better name, but has/wantsMetadata isn't right.

MarcoPolo: maybe we need a application provided `mergePartsMetadata`?

Right now the merge function when we receive a partsMetadata from the peer is a last write wins. But in your example a `binary or` would be more appropriate.

We could also replace

```
PartialMessageBytes(metadata []byte) ([]byte, []byte, error)
```

With
```
PartialMessageBytes(metadata []byte) ([]byte, error)
```
And update the peers partsMetatdata with `merge(peerParts, ourParts)`

MarcoPolo: done in: eb2f04a3205d

gossipsub.go
```diff
@@ -1375,6 +1375,10 @@ func (gs *GossipSubRouter) rpcs(msg *Message) iter.Seq2[peer.ID, *RPC] {
 			if pid == from || pid == peer.ID(msg.GetFrom()) {
 				continue
 			}
+			if peerStates, ok := gs.p.topics[topic]; ok && peerStates[pid].requestsPartial {
+				// The peer requested partial messages. We'll skip sending them full messages
+				continue
+			}
```
sukunrt: Do we need to check whether we support partialMessages?
Overall, can you add some documentation regarding this?

1. What happens when they support and we don't.
1. What happens when we support and they don't.

gossipsub.go
```diff
@@ -1375,6 +1375,10 @@ func (gs *GossipSubRouter) rpcs(msg *Message) iter.Seq2[peer.ID, *RPC] {
 			if pid == from || pid == peer.ID(msg.GetFrom()) {
 				continue
 			}
+			if peerStates, ok := gs.p.topics[topic]; ok && peerStates[pid].requestsPartial {
```
sukunrt: On line 871, there's IDONTWANT logic for incoming messages. We should skip sending IDONTWANT for partial message peers.

pubsub.go
```diff
@@ -1192,9 +1241,16 @@ func (p *PubSub) doAnnounceRetry(pid peer.ID, topic string, sub bool) {
...
+			requestPartialMessages = t.requestPartialMessages
+		}
+	}
 	subopt := &pb.RPC_SubOpts{
 		Topicid:   &topic,
 		Subscribe: &sub,
+		Partial:   &requestPartialMessages,
 	}
 
 	out := rpcWithSubs(subopt)
```
sukunrt, line 1269: This needs to be sent with our extensions / Hello packet. Otherwise the first RPC we send the peer won't have the partial messages extension.

sukunrt: fixed in: fe1f68da7d0119bdc0c568cd2e1c9307682a4087

I could see the impact in: https://github.com/libp2p/test-plans/pull/691

sukunrt: Here's an alternative implementation: a638f7f040960b48975ca0c401552e7d701eff6d

gossipsub.go
MarcoPolo: done in 237d4182

pubsub.go
MarcoPolo: nice! I merged a slight variant of https://github.com/libp2p/go-libp2p-pubsub/commit/fe1f68da7d0119bdc0c568cd2e1c9307682a4087

sukunrt (approved):
LGTM :tada:

PR #633, opened 2025-08-09
Migrate to `log/slog`

MarcoPolo:
related to https://github.com/libp2p/go-libp2p/issues/3362.

- Changes all logging to be structured.
- Removes the global logger. This is really nice for tests since you can now see which ID a log line belongs to.
- Adds a sub-logger just for RPC messages. In my experience the RPC messages are the only thing I need to debug issues. It's nice to have the structured log there I can use easily without having to futz around with tracers.
- Implements the LogValuer interface for the RPC pb type. In practice it's much easier to update this implementation than it is to maintain a separate trace protobuf system.

Currently inlines an internal `gologshim.go` package to maintain ipfs/go-log environment variable compatibility. Once https://github.com/libp2p/go-libp2p/pull/3364 is merged, we can use the gologshim package in go-libp2p.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"iter"
+	"log/slog"
 	"math/rand"
 	"sort"
 	"time"
@@ -20,8 +21,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/record"
 	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoremem"
-
-	"go.uber.org/zap/zapcore"
 )
 
 const (
@@ -338,7 +337,7 @@ func WithPeerScore(params *PeerScoreParams, thresholds *PeerScoreThresholds) Opt
 			return err
 		}
 
-		gs.score = newPeerScore(params)
+		gs.score = newPeerScore(params, ps.logger)
 		gs.gossipThreshold = thresholds.GossipThreshold
 		gs.publishThreshold = thresholds.PublishThreshold
 		gs.graylistThreshold = thresholds.GraylistThreshold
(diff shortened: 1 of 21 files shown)


gossipsub.go
```diff
@@ -839,7 +837,7 @@ func (gs *GossipSubRouter) handleIWant(p peer.ID, ctl *pb.ControlMessage) []*pb.
 	// we don't respond to IWANT requests from any peer whose score is below the gossip threshold
 	score := gs.score.Score(p)
 	if score < gs.gossipThreshold {
-		log.Debugf("IWANT: ignoring peer %s with score below threshold [score = %f]", p, score)
+		gs.p.logger.Debug("IWANT: ignoring peer with score below threshold", "peer", p, "score", score)
```
sukunrt: NIT: I'd just have the logger reference on gs. So we can do `gs.logger.Debug`

internal/gologshim/gologshim.go
```diff
@@ -0,0 +1,138 @@
...
+		logFmt = os.Getenv("GOLOG_LOG_FMT")
+	}
+	if logFmt == "json" {
+		c.format = logFormatJSON
+	}
+
+	logFmt = os.Getenv("GOLOG_LOG_ADD_SOURCE")
+	if logFmt != "" && logFmt != "0" {
+		c.addSource = true
+	}
```
sukunrt: Same remark as on the go-libp2p PR. This should be default True. It's really nice to have the full file name.

pubsub.go
```diff
@@ -457,10 +459,13 @@ type Option func(*PubSub) error
 
 // NewPubSub returns a new PubSub management object.
 func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option) (*PubSub, error) {
+	log := gologshim.Logger("pubsub")
+
```
sukunrt: Should we add a `self` key set equal to the host PeerID to this logger. Then we can identify which pubsub instance is doing the logging.

pubsub.go
```diff
@@ -511,6 +516,10 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		}
 	}
 
+	if ps.rpcLogger == nil {
+		ps.rpcLogger = slog.New(ps.logger.Handler().WithAttrs([]slog.Attr{slog.String("system", "pubsub/rpc")}))
+	}
```
sukunrt, line 521: Is there a way to enable only these rpc logs?

sukunrt (approved):

pubsub.go
MarcoPolo: yes. You pass a custom rpcLogger

PR #635, opened 2025-08-18
fix: Select ctx.Done() when preprocessing to avoid blocking on cancel

### LINKED ISSUE #636: Publish blocks indefinitely when context is canceled
### Issue
When calling `Publish`, a `Preprocess` function is pushed into the `eval` channel.  If the consumer is not actively reading from the channel, this send operation can block indefinitely.  When the context is canceled, the operation still waits instead of exiting, leading to a deadlock.  

### Fix
The PR updates the send logic to use a `select` with `ctx.Done()` and `t.p.ctx.Done()`, ensuring the operation terminates gracefully.  A test case reproducing the issue is included in the PR for verification.

infrmtcs:
Close #636 

The PR updates the send logic to use a select with ctx.Done() and t.p.ctx.Done(), ensuring the operation terminates gracefully. A test case reproducing the issue is included in the PR for verification.

### DIFF
--- a/topic.go
+++ b/topic.go
@@ -349,8 +349,14 @@ func (t *Topic) validate(ctx context.Context, data []byte, opts ...PubOpt) (*Mes
 	}
 
 	msg := &Message{m, "", t.p.host.ID(), pub.validatorData, pub.local}
-	t.p.eval <- func() {
+	select {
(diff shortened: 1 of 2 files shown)


MarcoPolo (approved):
Thank you!

Issue #617, opened 2025-06-04
GossipSub PX shares unreachable peers


dennis-tra:
A few weeks ago we added GossipSub "crawling" support to Nebula [[link](https://github.com/dennis-tra/nebula/pull/92/)]. This works as follows:

1. Connect to a peer that supports GossipSub
2. Wait until the other peer opens the GossipSub (`/meshsub/...`) stream to us and send the hello RPC. This RPC contains the active subscriptions of the remote peer.
3. Send the same hello message back to the remote peer (containing the same subscriptions)
4. Send a graft message to the remote peer that contains all subscriptions
5. Wait for the other peer to send a prune message that then contains additional peers which we can dial.

We are only running this routine for the Filecoin network so far because that's the only network we know of that makes use of the PX feature.

With the above technique we were able to identify ~50% additional peers in the Filecoin network:

```
PX Disabled: crawlDuration=31.395986409s crawledPeers=684
PX Enabled:  crawlDuration=51.537761043s crawledPeers=1015
```

However, virtually all of them are unreachable (>97%):

<img width="774" alt="Image" src="https://github.com/user-attachments/assets/c099c0c1-5b4a-4c1c-8e51-2007971da6e5" />

This graph shows the error distribution when trying to dial the peers that we have identified via GossipSub [[source](https://probelab.io/filecoin_mainnet/dht/2025-22/#dial_errors_over_7d_plot-plot)]. There are two error cases that stand out:

- `no_public_address` means that the set of multiaddresses which we have learned through PX does only contain addresses from private IP ranges, e.g., 127.0.0.1. Thus Nebula does not try to dial these peers.
- `io_timeout` means that connecting to the peer timed out (5s dial timeout)

I'm not super familiar with the internals of the GossipSub PX but I would propose to add:

1. Some configuration similar to the various DHT routing table/query/diversity filter [here](https://github.com/libp2p/go-libp2p-kad-dht/blob/2a7d79e9618a5b4148beaa41eb2d846e88a217d5/dht.go#L122) which checks that a peer satisfies some requirements before they are added to the PX pool (like: has at least one public IP address).
2. Periodically checks the pool of PX peers for reachability and removes them from the list (maybe that's already there?)

If both things would solve the problems, according to the above numbers, this would render PX in Filecoin basically useless because there would only be a handful peers being shared around.

I just wanted to start the discussion here and get some feedback 👍 

References:

[...]

PR #640, opened 2025-08-26
Add filter for public IPs on GossipSub peer exchange

cortze:
# Description
Following up on the issue raised in #617, this PR adds a new argument to the `WithPeerExchange` GossipSub option to restrict peer exchange to only include peer signed records with public IPs.

I can see a scenario where, despite a peer having a public IP in the PeerStore, we might not have a signed record for that IP. However, I think sharing just the PeerID (without private records) is still sufficient or at least better than sharing unsigned or private addresses.

I’m happy to discuss any of these points.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -8,6 +8,7 @@ import (
 	"iter"
 	"log/slog"
 	"math/rand"
+	"slices"
 	"sort"
 	"time"
 
@@ -21,6 +22,8 @@ import (
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/record"
 	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoremem"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 const (
@@ -286,25 +289,26 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
 func DefaultGossipSubRouter(h host.Host) *GossipSubRouter {
 	params := DefaultGossipSubParams()
 	return &GossipSubRouter{
-		peers:        make(map[peer.ID]protocol.ID),
-		mesh:         make(map[string]map[peer.ID]struct{}),
-		fanout:       make(map[string]map[peer.ID]struct{}),
-		lastpub:      make(map[string]int64),
-		gossip:       make(map[peer.ID][]*pb.ControlIHave),
-		control:      make(map[peer.ID]*pb.ControlMessage),
-		backoff:      make(map[string]map[peer.ID]time.Time),
-		peerhave:     make(map[peer.ID]int),
-		peerdontwant: make(map[peer.ID]int),
-		unwanted:     make(map[peer.ID]map[checksum]int),
-		iasked:       make(map[peer.ID]int),
-		outbound:     make(map[peer.ID]bool),
-		connect:      make(chan connectInfo, params.MaxPendingConnections),
-		cab:          pstoremem.NewAddrBook(),
-		mcache:       NewMessageCache(params.HistoryGossip, params.HistoryLength),
-		protos:       GossipSubDefaultProtocols,
-		feature:      GossipSubDefaultFeatures,
-		tagTracer:    newTagTracer(h.ConnManager()),
-		params:       params,
+		peers:           make(map[peer.ID]protocol.ID),
+		mesh:            make(map[string]map[peer.ID]struct{}),
+		fanout:          make(map[string]map[peer.ID]struct{}),
+		lastpub:         make(map[string]int64),
+		gossip:          make(map[peer.ID][]*pb.ControlIHave),
+		control:         make(map[peer.ID]*pb.ControlMessage),
+		backoff:         make(map[string]map[peer.ID]time.Time),
+		peerhave:        make(map[peer.ID]int),
+		peerdontwant:    make(map[peer.ID]int),
+		unwanted:        make(map[peer.ID]map[checksum]int),
+		iasked:          make(map[peer.ID]int),
+		outbound:        make(map[peer.ID]bool),
+		connect:         make(chan connectInfo, params.MaxPendingConnections),
+		cab:             pstoremem.NewAddrBook(),
+		mcache:          NewMessageCache(params.HistoryGossip, params.HistoryLength),
+		protos:          GossipSubDefaultProtocols,
+		feature:         GossipSubDefaultFeatures,
+		tagTracer:       newTagTracer(h.ConnManager()),
+		params:          params,
+		reducePXRecords: defaultPXRecordReducer,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gossipsub.go
```diff
@@ -381,14 +382,15 @@ func WithFloodPublish(floodPublish bool) Option {
 // WithPeerExchange is a gossipsub router option that enables Peer eXchange on PRUNE.
 // This should generally be enabled in bootstrappers and well connected/trusted nodes
 // used for bootstrapping.
-func WithPeerExchange(doPX bool) Option {
+func WithPeerExchange(doPX bool, pubMaddrsOnly bool) Option {
```
Kubuxu: I would make this a separate option. This way it won't break the interface for everyone already using `WithPerrExchange`

cortze: That makes sense. Updated on [69d3262](https://github.com/libp2p/go-libp2p-pubsub/pull/640/commits/69d326205c712ba810396ce02fb4e871d536c6c1)

gossipsub.go
```diff
@@ -389,6 +390,24 @@ func WithPeerExchange(doPX bool) Option {
...
+		gs.pubPX = false
+
+		return nil
+	}
+}
+
+// WithPeerExchange is a gossipsub router option that enables Peer eXchange on PRUNE.
+// This should generally be enabled in bootstrappers and well connected/trusted nodes
+// used for bootstrapping but using only Signer PeerRecords that include public addreses.
+func WithPublicAddrsOnPeerExchange(doPX bool, pubMaddrsOnly bool) Option {
```
Kubuxu: I would make this manipulate only `pubPX`, and `WithPeerExchange` shouldn't manipulate `pubPX`. This way there is no order dependency between these options.

Kubuxu:
It would be great if someone with more current libp2p experience also looked at it.

MarcoPolo (changes requested):
Hey @cortze thanks for the changes. Overall looks good, just a couple notes. you can find my comments in the last commit.

I've squashed all your commits into a single commit and made some changes that you can see here: https://github.com/libp2p/go-libp2p-pubsub/compare/7cbe1cbee004afa9ef43f7a3369ca1cd9d751ef7..83e3c602debcb09f5f73b5a2a7ae2fa8be0ee45f

I don't think we should send empty peer records. Happy to discuss further if you feel strongly.

Please squash your future changes and drop my review commit, thanks!

cortze:
Hey @MarcoPolo, thanks a lot for the review and the suggestions. 

I've addressed all the points except for the one regarding the empty peer records. My original thought was to follow what is described in the comment in [line 2015](https://github.com/libp2p/go-libp2p-pubsub/blob/d53e38b98dc291c355671e19400abb478ecb6e52/gossipsub.go#L2015) - where we would also ignore the signed records that are not good up to our public IP check:
> // see if we have a signed peer record to send back; if we don't, just send
> // the peer ID and let the pruned peer find them in the DHT -- we can't trust
> // unsigned address records through px anyway.

However, I don't have any strong opinion either. My only comment on this would be that if we were to send the SignedRecords without a public address anyway, we might be at a similar point as when the PR was started. 
The case that comes to my mind, and that I would like to avoid, is: 
- A node behind a NAT that connects our node
- We know it has a public IP and thus is in our Peerstore, but we didn't receive that address in the SignedPeerRecords.
- When checking the address we know about the peer, we see that the peer has a public IP
- However, that same IP is not within the SignedPeerRecord
I think it would still be okay and useful to share only the ID of the peer. The rest of the peers can use the ID to resolve the IP somehow without trying to dial the private addresses.

Either way, this is a good moment to choose the behaviour we think is best for the existing cases.

Also, and somehow out of the discussion, after talking with @dennis-tra, he raised a good point on making the GossipSub option open to any filtering function, not only the check for public addresses. I would make things easier to configure, but I would also like to hear what you think about this.

MarcoPolo:
> I’ve addressed all the points except for the one regarding the empty peer records. My original thought was to follow what is described in the comment in [line 2015](https://github.com/libp2p/go-libp2p-pubsub/blob/d53e38b98dc291c355671e19400abb478ecb6e52/gossipsub.go#L2015) - where we would also ignore the signed records that are not good up to our public IP check:

Ah, that’s a good point. Although the difference might be that the peer *didn’t* give us a signed peer record versus having given us one and it not containing a public address. The latter reveals a bit more about the peer, namely that the peer was capable of creating a signed peer record and chose not to include the public address.

> > // see if we have a signed peer record to send back; if we don’t, just send // the peer ID and let the pruned peer find them in the DHT – we can’t trust // unsigned address records through px anyway.
>
> However, I don’t have any strong opinion either. My only comment on this would be that if we were to send the SignedRecords without a public address anyway, we might be at a similar point as when the PR was started.

Agreed. We should never send signed peer records that contain no public addresses.

> The case that comes to my mind, and that I would like to avoid, is:
>
> - A node behind a NAT that connects our node
> - We know it has a public IP and thus is in our Peerstore, but we didn’t receive that address in the SignedPeerRecords.
> - When checking the address we know about the peer, we see that the peer has a public IP
> - However, that same IP is not within the SignedPeerRecord

In this case there may be no point in sharing the peer ID at all. The public IP address that you see is a NAT’ed IP address that won’t work for inbound connections. The only case it would be useful is if the node had registered a relay address that it shared on the DHT but not in its signed peer record.

> I think it would still be okay and useful to share only the ID of the peer. The rest of the peers can use the ID to resolve the IP somehow without trying to dial the private addresses.
>
> Either way, this is a good moment to choose the behaviour we think is best for the existing cases.

Maybe the safest/most-consistent-with-previous-behavior option is to include the peer id without the signed peer record as you initially did.

> Also, and somehow out of the discussion, after talking with @dennis-tra, he raised a good point on making the GossipSub option open to any filtering function, not only the check for public addresses. I would make things easier to configure, but I would also like to hear what you think about this.

I considered this as well. We recently merged a similar change in go-libp2p <https://github.com/libp2p/go-libp2p/commit/95cad5091fa1ee8af8b3cdf45c21630ca6b25dab>.

I think it’s a good idea. And it gives users an alternative from whatever we decide here. Let’s make it configurable with an easy option to enable what you propose in this PR (including just sending the PeerID in case of a signed peer record without public addresses).

Instead of `.pubPX` we’d have a `reducePXRecords` that takes a func of type: `func(pxRecords []*pb.PeerInfo, p peer.ID, addrs []multiaddr.Multiaddr, record *record.Envelope) []*pb.PeerInfo`

The pxPeers loop becomes:

``` go
for _, p := range pxPeers {
  var record *record.Envelope
  if cabOK {
    record = cab.GetPeerRecord(p)
  }
  px = reducePXRecords(px, p, gs.p.host.Peerstore().Addrs(p), record)
}
```

We’d have the following options:

``` go
func OnlyPublicAddrsOnPeerExchange() Option

type PXRecordReducer func(pxRecords []*pb.PeerInfo, p peer.ID, addrs []multiaddr.Multiaddr, record *record.Envelope) []*pb.PeerInfo

// WithCustomPXRecordReducer enables custom filtering of PX Records. See
// the implementation of OnlyPublicAddrsOnPeerExchange for an example.
func WithCustomPXRecordReducer(f PXRecordReducer) Option
```

The `OnlyPublicAddrsOnPeerExchange` option would use something like the following reducer:

``` go
func publicAddrsPXRecordReducer(pxRecords []*pb.PeerInfo, p peer.ID, addrs []multiaddr.Multiaddr, record *record.Envelope) []*pb.PeerInfo {
  if !slices.ContainsFunc(addrs, manet.IsPublicAddr) {
    return pxRecords
  }

  var recordBytes []byte
  if record != nil {
    err := checkPubAddresOnSignedPeerRecord(p, record);
    if err != nil {
      log.Warnf("no public address on signed peer record for %s: %s", p, err)
      // This peer has a public address, it just doesn't
      // have it in its signed peer record.
      // We'll share an empty record to at least inform the peer id.

    } else {
      var err error
      recordBytes, err = record.Marshal()
      if err != nil {
        log.Warnf("error marshaling signed peer record for %s: %s", p, err)
        return pxRecords
      }
    }
  }
  return append(pxRecords, &pb.PeerInfo{PeerID: []byte(p), SignedPeerRecord: recordBytes})
}
```

This probably also cleans up some complexity in the current makePrune method which is appreciated.

What do you think @cortze? If you agree, do you think you could update the PR?

cortze:
Sure! Let me make a second iteration over the PR over the day, and I'll squash the changes again.

MarcoPolo:
Looks good @cortze! I pushed a small change. Can you add a test to ensure we exchange the correct peer exchange records?

After that I think we're good here and can merge

cortze:
I've spotted a small issue on the `publicAddrsPXRecordReducer()` that was breaking the forwarding of the `peer.ID` as we agreed:
 
> Maybe the safest/most-consistent-with-previous-behavior option is to include the peer id without the signed peer record as you initially did.

Plus an extra check on `checkPubAddrsOnSignedPeerRecord()` that prevents panics with `nil` `*record.Envelop`.

Posting here the `git diff` of the `gossipsub.go` to make it easier to track:
```git
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -444,7 +444,7 @@ func OnlyPublicAddrsOnPeerExchange() Option {
 func publicAddrsPXRecordReducer(pxRecords []*pb.PeerInfo, p peer.ID, addrs []multiaddr.Multiaddr, record *record.Envelope) []*pb.PeerInfo {
        if !slices.ContainsFunc(addrs, manet.IsPublicAddr) {
-               return pxRecords
+               return defaultPXRecordReducer(pxRecords, p, addrs, nil)
        }
        if !checkPubAddrsOnSignedPeerRecord(p, record) {
                log.Warnf("no public address on signed peer record for %s", p)

@@ -454,6 +454,9 @@ func publicAddrsPXRecordReducer(pxRecords []*pb.PeerInfo, p peer.ID, addrs []mul
 }

 func checkPubAddrsOnSignedPeerRecord(p peer.ID, env *record.Envelope) bool {
+       if env == nil {
+               return false
+       }
        rec, err := env.Record()
        if err != nil {
                log.Warnf("error getting peer record for %s from signed envelop: %v", p, err)
```

Thanks again @MarcoPolo for the feedback ✌🏽

MarcoPolo:
> I've spotted a small issue on the `publicAddrsPXRecordReducer()` that was breaking the forwarding of the `peer.ID` as we agreed:
> 
> > Maybe the safest/most-consistent-with-previous-behavior option is to include the peer id without the signed peer record as you initially did.
> 
> Plus an extra check on `checkPubAddrsOnSignedPeerRecord()` that prevents panics with `nil` `*record.Envelop`

Nice, thanks!

> Posting here the `git diff` of the `gossipsub.go` to make it easier to track:

git range-diff is great for this: https://git-scm.com/docs/git-range-diff. e.g. `git range-diff master 68fb2e5dafa4812a924a95869df35ae19a51c88c 1e59ebd7b399fe2bbd765dbf4e18509599c8185d`

MarcoPolo:
> Looks good @cortze! I pushed a small change. Can you add a test to ensure we exchange the correct peer exchange records?
> 
> After that I think we're good here and can merge

Sorry I wasn't totally clear. I meant a end to end test with gossipsub and libp2p nodes to make sure the whole setup works as expected. If you don't get around to this, I'll add a test soon.

Thanks!

MarcoPolo:
Actually, we already have a test for PX in `TestGossipsubStarTopology`. So we're good here! I'll merge this after tests pass.

MarcoPolo (approved):

PR #643, opened 2025-09-07
chore: add check for dscore < dhi and bootstrap params

dozyio:
Was getting an out-of-bounds panic when running as a bootstrapper (D=D_lo=D_hi=D_out=0) https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#recommendations-for-network-operators.

Also added an early return to the new gossipsub validation to bootstrapper case.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -242,6 +242,20 @@ type GossipSubParams struct {
 }
 
 func (params *GossipSubParams) validate() error {
+	if !(params.HistoryGossip <= params.HistoryLength) {
+		return fmt.Errorf("param HistoryGossip=%d must be less than or equal to HistoryLength=%d", params.HistoryGossip, params.HistoryLength)
+	}
+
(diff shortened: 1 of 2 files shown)


MarcoPolo:
thanks!

MarcoPolo:
Thinking about this more, it's actually an error to set Dscore > Dhi. Dhi is our high watermark for peers in the mesh, and Dscore specifies how many high scoring peers are kept. By setting Dscore > Dhi you try to place keep more peers in your mesh than your hi watermark. 

You probably want to set Dscore to 0 for the bootstrapper case.

Do you mind changing the validation to error if Dscore > Dhi?

dozyio:
Thanks Marco - yeah that makes sense. Removed the fix, added a few bits the validation.

MarcoPolo (approved):

PR #648, opened 2025-10-13
cancel enqueued rpcs on receiving IDontWant

### LINKED ISSUE #611: Cancel queued `Publish` RPCs if `IDONTWANT` arrives in the meantime

sukunrt:
fixes: #611


log.txt
MarcoPolo, line 1: mistake?

sukunrt: whoops. Sorry that was for debugging.

comm.go
```diff
@@ -215,6 +218,10 @@ func rpcWithMessages(msgs ...*pb.Message) *RPC {
 	return &RPC{RPC: pb.RPC{Publish: msgs}}
 }
 
+func rpcWithMessageAndMsgID(msg *pb.Message, msgID string) *RPC {
```
MarcoPolo: This should go next to the RPC type in pubsub.go. (I know the previous similar function is in here, it should also move)

pubsub.go
```diff
@@ -274,6 +276,16 @@ func (rpc *RPC) split(limit int) iter.Seq[RPC] {
 
 			messagesInNextRPC := 0
 			messageSlice := rpc.Publish
+			var msgIDSlice []string
```
MarcoPolo: reminder to run the fuzzer after this change

pubsub.go
```diff
@@ -274,6 +276,16 @@ func (rpc *RPC) split(limit int) iter.Seq[RPC] {
...
+			if len(rpc.MsgIDs) != len(rpc.Publish) {
+				if len(rpc.MsgIDs) == 0 {
+					msgIDSlice = make([]string, len(rpc.Publish))
+				} else {
+					panic("MsgIDs and Publish have different lengths")
```
MarcoPolo: Is panic the best option? Is there a reason it's unsafe to use rpc.Publish as the source of truth?

rpc_queue.go
```diff
@@ -50,20 +51,41 @@ type rpcQueue struct {
 	dataAvailable  sync.Cond
 	spaceAvailable sync.Cond
 	// Mutex used to access queue
-	queueMu sync.Mutex
-	queue   priorityQueue
+	queueMu      sync.Mutex
+	queue        priorityQueue
+	queuedMsgIDs map[string]int      // message ids in queue
```
MarcoPolo: It seems an error to have multiple identical message ids in the queue, this could probably just be a `map[string]struct{}`

rpc_queue.go
```diff
@@ -133,10 +161,48 @@ func (q *rpcQueue) Pop(ctx context.Context) (*RPC, error) {
...
+			hasCancellations = true
+		}
+	}
+	if hasCancellations {
+		// clone the RPC parts that we'll modify. It may be shared with other queues.
+		newRPC := *rpc
+		newRPC.RPC.Publish = slices.Clone(rpc.RPC.Publish)
+		newRPC.MsgIDs = slices.Clone(rpc.MsgIDs)
+		rpc = &newRPC
+		for i, msgID := range slices.Backward(newRPC.MsgIDs) {
```
MarcoPolo: It's probably simpler (and a bit faster) to manually do the copy.

e.g.:
```
newRPC.Publish = make(...,0, N)
for _, msg := rpc.Publish {
  if notCancelled {
    newRpc.Publish =  append(...)
}
```

rpc_queue.go
```diff
@@ -133,10 +161,48 @@ func (q *rpcQueue) Pop(ctx context.Context) (*RPC, error) {
...
+			if msgID == "" {
+				continue
+			}
+			_, ok := q.cancelledIDs[msgID]
```
MarcoPolo: when do you `delete(q.cancelledIDs, msgID)`?

MarcoPolo (commented):
some small changes, but I think we should merge this.

pubsub.go
sukunrt: This effectively says: panic if the MessageIDs are present but the length is < rpc.Publish length. I'm panicing here because it'd have paniced in the append later in the loop. 

I'm changing this to work for any len of MessageIDs. Slightly ugly but also avoids allocation.

rpc_queue.go
sukunrt: I didn't want to rely on that property here. But I think that simplifies it a lot. So I'll make this change.

rpc_queue.go
sukunrt: 5 lines below. It's confusing because it's controlled by queuedMsgID counts. I've changed this to use a struct{}

rpc_queue.go
sukunrt: Doing the whole thing manually now.

PR #650, opened 2025-10-24
partialmessages: add per peer bounds on peer initiated groups

MarcoPolo:
A peer initiated group state is a state that is initiatlized due to receiving a partial message RPC for which we have not (yet) published to.

It's useful to have this state so we can remember what parts a peer has if we do publish to that group.

Previously we limited the total number of peer initiated group states in a topic. This was agnostic to how many groups a single peer initiated, which meant that a single peer could exhaust the whole limit.

This change tracks how many group states each peer has initialized and limits each peer to a configurable number of these states. The default total and per peer limits of 255 8 are likely good enough for most use cases, but the options exist for users to change this if they need to.

cc @kasey who suggested the idea.

### DIFF
--- a/partialmessages/partialmsgs.go
+++ b/partialmessages/partialmsgs.go
@@ -67,24 +67,24 @@ func (ps *peerState) IsZero() bool {
 	return ps.partsMetadata == nil && ps.sentPartsMetadata == nil
 }
 
-type partialMessageStatePerTopicGroup struct {
+type partialMessageStatePerGroupPerTopic struct {
 	peerState   map[peer.ID]*peerState
 	groupTTL    int
 	initiatedBy peer.ID // zero value if we initiated the group
 }
 
-func newPartialMessageStatePerTopicGroup(groupTTL int) *partialMessageStatePerTopicGroup {
-	return &partialMessageStatePerTopicGroup{
+func newPartialMessageStatePerTopicGroup(groupTTL int) *partialMessageStatePerGroupPerTopic {
+	return &partialMessageStatePerGroupPerTopic{
 		peerState: make(map[peer.ID]*peerState),
 		groupTTL:  max(groupTTL, minGroupTTL),
 	}
 }
 
-func (s *partialMessageStatePerTopicGroup) remotePeerInitiated() bool {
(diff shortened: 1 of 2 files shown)


partialmessages/partialmsgs.go
```diff
@@ -111,13 +117,26 @@ type PartialMessageExtension struct {
...
 	// publishing a partial message for the group.
 	GroupTTLByHeatbeat int
 
+	// map topic -> map[group]partialMessageStatePerTopicGroup
+	// TODO rename this to ...PerGroupPerTopic
 	statePerTopicPerGroup map[string]map[string]*partialMessageStatePerTopicGroup
 
-	peerInitiatedGroupCountPerTopic map[string]int
+	// map[topic]counter
+	peerInitiatedGroupCounter map[string]*peerInitiatedGroupCounterState
```
sukunrt: Reminder to fix the commented out code.

partialmessages/partialmsgs.go
```diff
@@ -111,13 +117,26 @@ type PartialMessageExtension struct {
...
 
+	// PeerInitiatedGroupLimitPerTopic limits the number of Group states all
+	// peers can initialize per topic. A group state is initialized by a peer if
+	// the peer's message marks the first time we've seen a group id.
+	PeerInitiatedGroupLimitPerTopic int
+
+	// PeerInitiatedGroupLimitPerTopicPerPeer limits the number of Group states
+	// a single peer can initialize per topic. A group state is initialized by a
+	// peer if the peer's message marks the first time we've seen a group id.
+	PeerInitiatedGroupLimitPerTopicPerPeer int
```
sukunrt, line 128: I'd recommend rephrasing the comments to talk of messages with group IDs as opposed to group state. The state is private to the package so the comment might be confusing. The clarifying statement helps, but if it's directly related to the number of new messages that'd be clearer.

sukunrt (approved):

partialmessages/partialmsgs.go
MarcoPolo: I'm not sure phrasing this in terms of messages is clearer. The thing this is limiting is the number of group ID related state for a group ID we don't know about.

Messages sounds like we are persisting application data, which we aren't.

Do you have a another suggestion?

PR #652, opened 2025-10-25
add `Announce` API for GossipSub message announcement

raulk:
Implements a new `Announce()` API that allows advertising messages via IHAVE without pushing them to the mesh. These messages are retained in the mcache until, at least, the specified deadline.

We send the `IHAVE` immediately to all connected topic subscribers, whether in mesh, in gossip, in a cached fanout, or none of these active pubsub states.

This enables pull-based message distribution, useful for scenarios outside of the app's critical path, such as backup availability. Subscribers can pull messages on-demand via `IWANT` requests.

## Details

- Add `Topic.Announce()` method that sends IHAVE gossip to topic subscribers with expiry-based message retention
- Refactor `MessageCache` to support dual storage model:
  - Sliding window for regular published messages (current)
  - Time wheel for announced messages with TTL-based expiry
- Add `GossipSubAnnouncementMaxTTL` parameter (default 60s) for sizing announcement storage
- Rename `MessageCache` methods for clarity (Put→AppendWindow, Shift→ShiftWindow, GetGossipIDs→GossipForTopic). Add missing godocs.
- Implement unified message storage with reference counting to handle messages in both window and announcement wheel
- Add heartbeat cleanup for expired announcements via `PruneAnns()`
- Add comprehensive test coverage for announcement functionality including storage, delivery, expiry, duplicates, and edge cases

## Open points

- [ ] Figure out the threat model. Any scoring implications? Spamming possible?
- [ ] Decide whether we need `Announce` handlers, so that the receiver can decide if they want to react by pulling the message or not (e.g. in the case of Ethereum, a node knows whether it has received the correct column at a particular slot, and can ignore any further announcements during that slot window)

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -62,6 +62,7 @@ var (
 	GossipSubHeartbeatInitialDelay            = 100 * time.Millisecond
 	GossipSubHeartbeatInterval                = 1 * time.Second
 	GossipSubFanoutTTL                        = 60 * time.Second
+	GossipSubAnnouncementMaxTTL               = 60 * time.Second
 	GossipSubPrunePeers                       = 16
 	GossipSubPruneBackoff                     = time.Minute
 	GossipSubUnsubscribeBackoff               = 10 * time.Second
@@ -167,6 +168,11 @@ type GossipSubParams struct {
 	// FanoutTTL since we've published to a topic that we're not subscribed to,
 	// we'll delete the fanout map for that topic.
 	FanoutTTL time.Duration
+
+	// AnnouncementMaxTTL is the maximum possible time-to-live for a message announced
+	// via Announce. This is used to size internal data structures. Deadlines passed to
+	// Announce exceeding this value will be clamped, and a warning will be logged.
+	AnnouncementMaxTTL time.Duration
 
 	// PrunePeers controls the number of peers to include in prune Peer eXchange.
 	// When we prune a peer that's eligible for PX (has a good score, etc), we will try to
@@ -292,6 +298,7 @@ func NewGossipSubWithRouter(ctx context.Context, h host.Host, rt PubSubRouter, o
 // DefaultGossipSubRouter returns a new GossipSubRouter with default parameters.
 func DefaultGossipSubRouter(h host.Host) *GossipSubRouter {
 	params := DefaultGossipSubParams()
+	mcache := NewMessageCache(params.HistoryGossip, params.HistoryLength, params.HeartbeatInterval, params.AnnouncementMaxTTL)
 	rt := &GossipSubRouter{
 		peers:           make(map[peer.ID]protocol.ID),
 		mesh:            make(map[string]map[peer.ID]struct{}),
@@ -307,7 +314,7 @@ func DefaultGossipSubRouter(h host.Host) *GossipSubRouter {
 		outbound:        make(map[peer.ID]bool),
 		connect:         make(chan connectInfo, params.MaxPendingConnections),
 		cab:             pstoremem.NewAddrBook(),
-		mcache:          NewMessageCache(params.HistoryGossip, params.HistoryLength),
+		mcache:          mcache,
 		protos:          GossipSubDefaultProtocols,
 		feature:         GossipSubDefaultFeatures,
 		tagTracer:       newTagTracer(h.ConnManager()),
@@ -341,6 +348,7 @@ func DefaultGossipSubParams() GossipSubParams {
 		HeartbeatInitialDelay:     GossipSubHeartbeatInitialDelay,
 		HeartbeatInterval:         GossipSubHeartbeatInterval,
 		FanoutTTL:                 GossipSubFanoutTTL,
+		AnnouncementMaxTTL:        GossipSubAnnouncementMaxTTL,
 		PrunePeers:                GossipSubPrunePeers,
 		PruneBackoff:              GossipSubPruneBackoff,
 		UnsubscribeBackoff:        GossipSubUnsubscribeBackoff,
@@ -569,7 +577,7 @@ func WithGossipSubParams(cfg GossipSubParams) Option {
 		// Overwrite current config and associated variables in the router.
 		gs.params = cfg
 		gs.connect = make(chan connectInfo, cfg.MaxPendingConnections)
-		gs.mcache = NewMessageCache(cfg.HistoryGossip, cfg.HistoryLength)
+		gs.mcache = NewMessageCache(cfg.HistoryGossip, cfg.HistoryLength, cfg.HeartbeatInterval, cfg.AnnouncementMaxTTL)
 
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


gossipsub.go
```diff
@@ -168,6 +169,11 @@ type GossipSubParams struct {
 	// we'll delete the fanout map for that topic.
 	FanoutTTL time.Duration
 
+	// AnnouncementMaxTTL is the maximum possible time-to-live for a message announced
+	// via Announce. This is used to size internal data structures. Deadlines passed to
+	// Announce exceeding this value will be clamped, and a warning will be logged.
```
raulk, line 174: note: I haven't implemented the clamping yet, just realised.

gossipsub.go
```diff
@@ -2033,6 +2048,44 @@ func (gs *GossipSubRouter) enqueueGossip(p peer.ID, ihave *pb.ControlIHave) {
...
+		})
+		gossipQueued = true
+	}
+
+	if !gossipQueued {
+		return
+	}
+
+	// Track announcement in message cache for IWANT retrieval
+	gs.mcache.TrackAnn(msg, expiry)
```
raulk, line 2083: I wonder if we can just remove the msgID from the MessageCache? We seem to be calculating the message ID over and over again, which adds up unnecessary compute time. We can calculate once and memoize.

More importantly the msg ID function can be overriden per topic, so having a global msgID in the MessageCache is definitely a design flaw.

raulk: WDYT @sukunrt?

mcache.go
```diff
@@ -2,69 +2,108 @@ package pubsub
...
 )
 
+type historyEntry struct {
+	mid   string
+	topic string
+}
+
+type messageRef struct {
+	*Message
+	refs int
```
raulk, line 17: I didn't like this refcounting, but it seemed like unified message storage was best. Although we can choose to be stricter (e.g. do not accept an announcement for a message that's sitting in the sliding window, and viceversa), in which case we can drop the refcount because a Message can only be tracked by one cache region at a time.

This has some problems though, namely dealing with a single message sent on multiple topics.

gossipsub.go
MarcoPolo: We could avoid another parameter here and adjust the wheel size dynamically from the largest TTL seen.

mcache.go
MarcoPolo: It would be surprising to me if I did an Announce for 60s, and the message stopped being available after 3 heartbeats because the message was actually held in the sliding window cache already. Likewise if I Announce multiple times.

The refcount seems fine in order to avoid that behavior.

>This has some problems though, namely dealing with a single message sent on multiple topics.

Does anyone send the same exact message (same message id) on multiple topics? It might break how a peer responds to an IWANT (as it only includes the message id. What topic should the peer send back?)

MarcoPolo:
> We send the IHAVE immediately to all connected topic subscribers, whether in mesh, in gossip, in a cached fanout, or none of these active pubsub states.

As a callout, this would be the first time we send an IHAVE to our mesh peers. It might be okay (at least the Go implementation would behave fine), but we could consider putting this specific behavior behind an extension.


---

Skimmed the diff, seems reasonable. When you move it out of draft I'll do a deeper review.

gossipsub.go
```diff
@@ -2033,6 +2048,44 @@ func (gs *GossipSubRouter) enqueueGossip(p peer.ID, ihave *pb.ControlIHave) {
...
+	tmap, ok := gs.p.topics[topic]
+	if !ok {
+		return
+	}
+
+	// Store message for IWANT retrieval in the message cache
+	msgID := gs.p.idGen.ID(msg)
+
+	// Send IHAVE to all topic peers (excluding direct peers, applying score threshold)
+	// Match the filtering logic from emitGossip
```
sukunrt, line 2062: Why do we exclude `directPeers` here? Alternatively, how do they discover that we have this message?

sukunrt: The code doesn't actually exclude, so just a stale comment?

gossipsub.go
sukunrt: I thought this was cached: 
https://github.com/libp2p/go-libp2p-pubsub/blob/raulk/announce/midgen.go#L41

Is this not the function that's ultimately used? 

I agree this code is unclear and we should rely on some conspicuously cached message id as opposed to calling this function again and again.

mcache.go
```diff
@@ -91,14 +131,86 @@ func (mc *MessageCache) GetGossipIDs(topic string) []string {
...
-func (mc *MessageCache) Shift() {
+// ShiftWindow advances the sliding window by one slot.
+// Messages that fall out of the window have their reference counts decremented
+// and are removed from the cache if they are no longer referenced.
+func (mc *MessageCache) ShiftWindow() {
 	last := mc.history[len(mc.history)-1]
 	for _, entry := range last {
-		delete(mc.msgs, entry.mid)
+		mc.tryDropMessage(entry.mid)
 		delete(mc.peertx, entry.mid)
```
sukunrt, line 141: how does the `mc.peertx` counter work? we should clear it when the message is actually dropped, no? 
```
if !mc.tryDropMessage(entry.mid) {
    delete(mc.peertx, entry.mid)
}
```

mcache.go
```diff
@@ -2,69 +2,108 @@ package pubsub
...
+// The slack between `gossipLen` and `historyLen` accounts for the reaction time
 // between when a message is advertised via IHAVE gossip, and the peer pulls it
 // via an IWANT command.
-func NewMessageCache(gossip, history int) *MessageCache {
-	if gossip > history {
+func NewMessageCache(gossipLen, historyLen int, heartbeatInterval, maxTTL time.Duration) *MessageCache {
+	if gossipLen > historyLen {
 		err := fmt.Errorf("invalid parameters for message cache; gossip slots (%d) cannot be larger than history slots (%d)",
-			gossip, history)
+			gossipLen, historyLen)
```
sukunrt, line 57: Not introduced in this PR, so we can let this be. 

Do we ever use entries from `[gossipsubLen: historyLen]`?

mcache.go
```diff
@@ -2,69 +2,108 @@ package pubsub
...
-
-func (mc *MessageCache) Put(msg *Message) {
-	mid := mc.msgID(msg)
-	mc.msgs[mid] = msg
-	mc.history[0] = append(mc.history[0], CacheEntry{mid: mid, topic: msg.GetTopic()})
+// AppendWindow adds a message to the sliding window cache.
+// The message will be retained for the duration of the window.
+// If the message already exists in the cache, its reference count is incremented.
+func (mc *MessageCache) AppendWindow(msg *Message) {
+	mid := mc.upsertMessage(msg)
```
sukunrt, line 86: I think we've subtly changed the definition of `AppendWindow` / `Put`

Previously a message appended multiple times would be removed from the cache at the expiry of the first put. 

Now it'll be removed from the cache at the expiry of the last put. 

Arguably, the previous definition is wrong. I haven't thought this through, but we should consider the impact this would have. 
https://github.com/libp2p/go-libp2p-pubsub/blob/master/mcache.go#L97

sukunrt (commented):
The structure seems fine to me. I'll review once again when it's ready for review.

PR #657, opened 2025-12-05
fix bug of not sending full messages to peers that requested partial

MarcoPolo:
When we support partial extensions.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -864,7 +864,7 @@ func (gs *GossipSubRouter) Preprocess(from peer.ID, msgs []*Message) {
 				// We don't send IDONTWANT to the peer that sent us the messages
 				continue
 			}
-			if gs.peerRequestsPartial(p, topic) {
+			if gs.iSupportSendingPartial(topic) && gs.peerRequestsPartial(p, topic) {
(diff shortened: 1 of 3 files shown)


gossipsub.go
```diff
@@ -1398,7 +1398,14 @@ func (gs *GossipSubRouter) peerSupportsPartial(p peer.ID, topic string) bool {
 	return ok && gs.extensions.myExtensions.PartialMessages && peerStates[p].supportsPartial
 }
 
-func (gs *GossipSubRouter) peerRequestsPartial(p peer.ID, topic string) bool {
+func (gs *GossipSubRouter) sendPeerPartialMessage(p peer.ID, topic string) bool {
+	myTopicState := gs.p.myTopics[topic]
```
sukunrt: I'd suggest structuring this as 4 methods. 

1. PeerSupportsReceivingPartial(p, topic)
2. PeerSupportsSendingPartial(p, topic)
3. ICanReceivePartial(topic)
4. ICanSendPartial(topic)

sukunrt (approved):
I think the approach suggested in the comment would make this cleaner. But this looks correct :)

PR #661, opened 2025-12-09
optimize parts metadata msgs

MarcoPolo:
This is a small optimization we can do to reduce the total number of messages sent. The commit contains more details.

The first commit is a refactor to make the optimization easier, the second is the optimization.


partialmessages/partialmsgs.go
```diff
@@ -140,8 +140,14 @@ type PublishOptions struct {
...
-	// EagerPush is data that will be eagerly pushed to peers in a PartialMessage
-	EagerPush []byte
+	// EagerPushWithPartsMetadata is used to eagerly push data to peers if we have
+	// not yet received their parts metadata. The encoded partial message sent
+	// is generated by calling `PartialMessageBytes` with the provided
+	// PartsMetadata.
+	//
+	// If nil, no partial message is sent to a peer unless we have received
+	// their partsMetadata first.
+	EagerPushWithPartsMetadata PartsMetadata
```
sukunrt, line 150: NIT: we want to eager push *these* parts not eager push and parts 
```suggestion
	EagerPushParts PartsMetadata
```

partialmessages/partialmsgs.go
```diff
@@ -262,10 +277,13 @@ func (e *PartialMessagesExtension) PublishPartial(topic string, partial Message,
...
 		//   - we have something to eager push
-		if requestedPartial && !inResponseToIWant && len(opts.EagerPush) > 0 {
+		if requestedPartial && !inResponseToIWant && len(eagerData) > 0 {
 			log.Debug("Eager pushing")
 			sendRPC = true
-			rpc.PartialMessage = opts.EagerPush
+			rpc.PartialMessage = eagerData
+			// Merge the peer's empty partsMetadata with the parts we eagerly pushed.
+			// This tracks what has been sent to the peer and avoids sending duplicates.
+			pState.partsMetadata = e.MergePartsMetadata(topic, pState.partsMetadata, opts.EagerPushWithPartsMetadata)
```
sukunrt: I feel the `inResponseToIWant` variable should be `havePeersPartsMetadata`. and rather than `inResponseToIWant=true` when `len(pm) > 0` we should always set `havePeerPartsMetadata=true` in the above block. That way if the peer has everything we have, we won't eager push to them. I think the current implementation would end up pushing to this peer.

partialmessages/partialmsgs.go
MarcoPolo: This is removed in https://github.com/libp2p/go-libp2p-pubsub/pull/662

partialmessages/partialmsgs.go
MarcoPolo: You're right, thank you

sukunrt (approved):

PR #662, opened 2026-01-13
partialmsgs: Add explicit eager push method

MarcoPolo:
Allows applications to send an extra header in the eager push case


partialmessages/partialmsgs.go
```diff
@@ -31,22 +31,20 @@ type PartsMetadata []byte
...
 
-	// PartialMessageBytes takes in the opaque request metadata and
-	// returns a encoded partial message that fulfills as much of the request as
-	// possible.
-	//
-	// If the Partial Message is empty, the implementation MUST return:
-	// nil, nil.
+	// PartialMessageBytes takes the opaque parts metadata and returns an
+	// encoded partial message that fulfills as much of the request as possible.
 	PartialMessageBytes(partsMetadata PartsMetadata) (msg []byte, _ error)
```
sukunrt, line 39: why not structure this as: 

```
	PartialMessageBytes(partsMetadata PartsMetadata, havePeersParts bool) (msg []byte, msgParts PartsMetadata, _ error)
```

This would allow us to skip all the eager pushing business here and let the application handle all that. eager / non eager etc is better handled completely by the application.

sukunrt (approved):

partialmessages/partialmsgs.go
MarcoPolo: discussed sync. I pointed out that you don't need to return the msgParts in the non eager case if you assume the application is responding to as much data as possible. Therefore the function signature looks different and we should keep the different interface methods.

PR #663, opened 2026-01-13
partialmsgs: Implement Partial Message gossip

MarcoPolo:
More efficient than IHAVE/IWANT as it includes more context (groupID) and allows peers to request subsets of the full message.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -2010,6 +2010,7 @@ func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}
 	for p := range gs.p.topics[topic] {
 		_, inExclude := exclude[p]
 		_, direct := gs.direct[p]
+
 		if !inExclude && !direct && gs.feature(GossipSubFeatureMesh, gs.peers[p]) && gs.score.Score(p) >= gs.gossipThreshold {
 			peers = append(peers, p)
 		}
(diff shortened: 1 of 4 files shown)


partialmessages/partialmsgs.go
```diff
@@ -342,6 +345,35 @@ func (e *PartialMessagesExtension) Heartbeat() {
...
+			GroupID:       []byte(group),
+			PartsMetadata: s.myLastPartsMetadata,
+		}
+		pState, peerStateOk := s.peerState[peer]
+		if !peerStateOk {
+			pState = &peerState{}
+			s.peerState[peer] = pState
+		}
+
+		if pState.sentPartsMetadata == nil || !bytes.Equal(rpc.PartsMetadata, pState.sentPartsMetadata) {
```
```suggestion
		if !bytes.Equal(rpc.PartsMetadata, pState.sentPartsMetadata) {
```

sukunrt: NIT: It'd be cleaner to make the rpc after this check

gossipsub.go
```diff
@@ -2039,6 +2040,13 @@ func (gs *GossipSubRouter) emitGossip(topic string, exclude map[peer.ID]struct{}
 			shuffleStrings(mids)
 			copy(peerMids, mids)
 		}
+
+		if gs.iSupportSendingPartial(topic) && gs.peerRequestsPartial(p, topic) {
+			// We should send the peer partial messages for gossip instead
+			gs.extensions.partialMessagesExtension.EmitGossip(topic, p)
+			continue
+		}
```
sukunrt, line 2051: you can batch these and enqueue gossip in batch for all peers in the same topic. That would allow us to reuse the rpc for all the relevant peers we will send the gossip to.

sukunrt (approved):

gossipsub.go
MarcoPolo: done in latest push: https://github.com/libp2p/go-libp2p-pubsub/compare/ce9c663c2c4d94b1f45392ce22d494461977cda6..208237add15f0ad2876f5fac2c638b17f67f8e45

PR #664, opened 2026-01-28
Include peer we are sending to for EagerPartialMessageBytes

MarcoPolo:
Allows applications to control the eager data depending on the destination peer.

Cell-Level Deltas in Ethereum for example, only need to send the PartialDataColumn header once per peer across any topics. This change enables that.


MarcoPolo:
Replaced by #669

PR #666, opened 2026-01-29
Add Rpc.From()

MarcoPolo:
Alternative to #665 that doesn't break existing tracers. I think I have a preference for this one.


sukunrt (approved):
I prefer this to the other one.

PR #667, opened 2026-01-30
partialmessages: Copy partsMetadata to avoid data races

MarcoPolo:
The code (essentially one line) was written by me, the test was added by Claude (reviewed by me).

### DIFF
--- a/partialmessages/partialmsgs.go
+++ b/partialmessages/partialmsgs.go
@@ -208,7 +208,6 @@ func (e *PartialMessagesExtension) Init(router Router) error {
 
 func (e *PartialMessagesExtension) PublishPartial(topic string, partial Message, opts PublishOptions) error {
 	groupID := partial.GroupID()
-	myPartsMeta := partial.PartsMetadata()
 
 	state, err := e.groupState(topic, groupID, false, "")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


partialmessages/partialmsgs.go
```diff
@@ -208,15 +208,18 @@ func (e *PartialMessagesExtension) Init(router Router) error {
...
 	if err != nil {
 		return err
 	}
 
 	state.groupTTL = max(e.GroupTTLByHeatbeat, minGroupTTL)
-	state.myLastPartsMetadata = slices.Clone(myPartsMeta)
+
+	// Copy this as the owner of the partial message if free to mutate the
+	// returned metadata after this function returns.
+	myPartsMeta := slices.Clone(partial.PartsMetadata())
```
sukunrt, line 221: Should it be upto the method `PartsMetadat` to clone / not share this?

sukunrt (approved):

partialmessages/partialmsgs.go
MarcoPolo: Yes, that's an alternative. This is defensive, but prevents optimization in the case you know you won't mutate the data, I think that's an okay tradeoff for now, but worth revisiting if this becomes a hotspot.

MarcoPolo:
Replaced by #669

PR #669, opened 2026-01-31
Refactor parts metadata

MarcoPolo:
To optimize when we send PartsMetadata we need the metadata to do two
things:

- Get the union of two parts metadata.
- Return if one parts metadata is a subset of another.

This allows us to avoid sending our updated partsMetadata to the peer
that gave us new data, as they can infer our updatedPartsMetadata.

### DIFF
--- a/partialmessages/partialmsgs.go
+++ b/partialmessages/partialmsgs.go
@@ -10,8 +10,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
-// TODO: Add gossip fallback (pick random connected peers and send ihave/iwants)
-
 const minGroupTTL = 3
 
 // defaultPeerInitiatedGroupLimitPerTopic limits the total number (per topic) of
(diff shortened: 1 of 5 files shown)


sukunrt (approved):

partialmessages/partialmsgs.go
```diff
@@ -25,7 +21,26 @@ const defaultPeerInitiatedGroupLimitPerTopicPerPeer = 8
...
+// so IsSubset is not expected to return true.
+type PartsMetadata interface {
+	Encode() []byte
+	// Clone returns a deep clone
+	Clone() PartsMetadata
+	// Merge creates a union with the provided PartsMetadata.
+	// Merge(nil) is a no-op.
+	Merge(PartsMetadata)
+	// Returns true if the object is a subset of B.
+	IsSubset(B PartsMetadata) bool
```
sukunrt, line 42: NIT: 
`IsSubsetOf` is clearer, I think.

PR #671, opened 2026-02-13
Simplify partial messages

MarcoPolo:
PeerState is now only used by the application, but its lifetime is
managed by gossipsub. This attempts to strike a reasonable tradeoff
between giving the application complete responsibility of handling every
peer's state, and the library from having to know too much about the
structure of PartsMetadata.

Applications can use the PeerState to store objects representing the
node's view of a remote's state, and the node's view of the remote's
view of our state. Applications SHOULD use this to avoid sending
redundant or duplicate data.

---

A future PR may use generics here instead of `any`, but otherwise very similar.


partialmessages/invariants.go
```diff
@@ -1,290 +0,0 @@
-package partialmessages
-
```
aarshkshah1992, line 2: @MarcoPolo Why have completely removed this ?

partialmessages/partialmsgs.go
```diff
@@ -313,9 +295,6 @@ func (e *PartialMessagesExtension) peersToPublishTo(topic string, state *partial
 	}
 }
 
-func (e *PartialMessagesExtension) AddPeer(id peer.ID) {
```
aarshkshah1992, line 316: @MarcoPolo The tests fail because we've removed this.

partialmessages/invariants.go
MarcoPolo: This was used when the Message interface had more methods and the library side expected more from the application. With these changes, the library side expects very little from the Message implementation, so these invariants are no longer required by the library. An application should have similar tests though, and they can be more specific.

partialmessages/partialmsgs.go
```diff
@@ -225,61 +223,38 @@ func (e *PartialMessagesExtension) PublishPartial(topic string, partial Message,
...
 
-	eagerData, eagerPartsMeta, err := partial.EagerPartialMessageBytes()
-	if err != nil {
-		return err
-	}
 	for p := range peers {
 		log := e.Logger.With("peer", p)
 		requestedPartial := e.router.PeerRequestsPartial(p, topic)
 
+		pState := state.peerState[p]
```
sukunrt: why is it not an error if requestedPartial is `false`?
As I understand, if the app doesn't provide a set of peers in the function call, the internal `peersToPublishTo` method will only have peers who request partial messages.

partialmessages/partialmsgs.go
```diff
@@ -34,44 +30,57 @@ type PartsMetadata []byte
...
-	partsMetadata PartsMetadata
-	// The parts metadata this node has sent to the peer
-	sentPartsMetadata PartsMetadata
+// PeerState is the state we keep per peer. Gossipsub manages the lifecycle of
+// this state, but does not use it. Applications SHOULD use this to avoid
+// sending redundant data.
+type PeerState struct {
+	// Opaque (to gossipsub) state used by the application to track a peer's
+	// state (e.g. PartsMetadata)
+	RecvdState any
```
sukunrt: NIT:
```suggestion
	ReceivedState any
```

I also like `RecvState` the 4 letters in `Recv` line up nicely with `Sent`

partialmessages/partialmsgs.go
```diff
@@ -80,38 +89,32 @@ func (s *partialMessageStatePerGroupPerTopic) remotePeerInitiated() bool {
...
+	//
+	// In particular it SHOULD use the rpc.PartialMessage contents to:
+	//  - update PeerState.RecvdState (we can infer they have a part if they have given it us)
+	//  - update PeerState.SentState (they can infer an update if they have given us a part)
+	//
+	// If PublishPartial is called within this callback, the ForPeer methods
+	// will not see the latest PeerState.
+	//
+	// If the rpc should be ignored, the application can return an empty PeerState.
+	OnIncomingRPC func(from peer.ID, peerState PeerState, rpc *pb.PartialMessagesExtension) (PeerState, error)
```
```suggestion
	OnIncomingRPC func(from peer.ID, previousPeerState PeerState, rpc *pb.PartialMessagesExtension) (newPeerState PeerState, _ error)
```

partialmessages/partialmsgs.go
```diff
@@ -34,44 +30,57 @@ type PartsMetadata []byte
...
-	EagerPartialMessageBytes() (msg []byte, partsMetadata PartsMetadata, _ error)
-
-	PartsMetadata() PartsMetadata
+	// ForPeer takes the remote's peer ID, if the peer requested partial
+	// message data, and its peerState and returns an encoded partial message that
+	// fulfills some or all of the request.
+	//
+	// If requestedMessage is false, the returned encoded message MUST be empty,
+	// as the peer does not want encoded partial messages. The implementation
+	// SHOULD still return relevant partsMetadataToSend.
```
sukunrt: why? Can you elaborate on this?

partialmessages/partialmsgs.go
```diff
@@ -34,44 +30,57 @@ type PartsMetadata []byte
...
+	//
+	// The implementation should update:
+	// - peerState.RecvdState to remember what parts we've sent to this peer.
```
sukunrt: I think this should be

```suggestion
	// - peerState.RecvdState to remember what parts the peer has sent to us.
```

partialmessages/partialmsgs.go
```diff
@@ -386,39 +362,52 @@ func (e *PartialMessagesExtension) HandleRPC(from peer.ID, rpc *pb.PartialMessag
...
+
+	if !groupStateExists {
+		// If we don't have the groupState, only create it if there is something
+		// worth storing.
+		pState, err := e.OnIncomingRPC(from, PeerState{}, rpc)
+		if err != nil {
+			return err
+		}
+		if !pState.IsZero() {
+			state, err := e.groupState(topic, groupID, true, from)
```
sukunrt: NIT: I think this will read better if we rename `state` -> `gState`

partialmessages/partialmsgs.go
```diff
@@ -225,61 +223,38 @@ func (e *PartialMessagesExtension) PublishPartial(topic string, partial Message,
...
-		if pState.sentPartsMetadata == nil || !bytes.Equal(myPartsMeta, pState.sentPartsMetadata) {
+		if len(partsMetadata) > 0 {
 			log.Debug("Including parts metadata")
 			sendRPC = true
-			pState.sentPartsMetadata = myPartsMeta
-			rpc.PartsMetadata = myPartsMeta
+			rpc.PartsMetadata = partsMetadata
 		}
 
 		if sendRPC {
```
sukunrt: This block in `gossipForPeer` looks cleaner: 

```
	if len(msg) > 0 || len(partsMetadataToSend) > 0 {
				rpc.TopicID = &topic
				rpc.GroupID = []byte(group)
				rpc.PartialMessage = msg
				rpc.PartsMetadata = partsMetadataToSend
				e.sendRPC(peer, &rpc)
			}
```

sukunrt (approved):

partialmessages/partialmsgs.go
MarcoPolo: Remember each peer tells us two things:
- They Support partial messages, meaning they can provide you partial messages, but they prefer full messages.
- They Request partial messages, meaning they both support partial messages as above and they would like to receive partial messages.

The library will provide peers that support partial messages but not request them as that's how you request the parts from them.

partialmessages/partialmsgs.go
MarcoPolo: https://github.com/libp2p/go-libp2p-pubsub/pull/671#discussion_r2824155118

The peer needs to get your partsMetadata in order to know what to send you.

partialmessages/partialmsgs.go
MarcoPolo: agreed. This is what it kind of what it looks like in #674

partialmessages/partialmsgs.go
MarcoPolo: hm, this can probably be rephrased. The original is closer to what I want because we are updating their recvdState to remember that we sent them part X. That way, next time we are publishing, we don't resend them the part we've previously sent.

partialmessages/partialmsgs.go
sukunrt: This makes sense. Thanks!

partialmessages/partialmsgs.go
sukunrt: I'm confused. This is the peer's partsmetadata that **we've received** from this peer, right? 

The very next line: 
```
	// peerState.RecvdState is nil if the node has not seen its peer's
	// partsMetadata, applications can still eagerly push data in this case.

```

partialmessages/partialmsgs.go
```diff
@@ -225,61 +223,38 @@ func (e *PartialMessagesExtension) PublishPartial(topic string, partial Message,
...
+		pState := state.peerState[p]
 		var rpc pb.PartialMessagesExtension
 		var sendRPC bool
 
-		pState, peerStateOk := state.peerState[p]
-		if !peerStateOk {
-			pState = &peerState{}
-			state.peerState[p] = pState
+		// Ask the application for message data
+		pState, pm, partsMetadata, err := partial.ForPeer(p, requestedPartial, pState)
```
aarshkshah1992: @MarcoPolo @sukunrt 

There is one problem here. Either we solve it here or document it well.
Basically, we pass a copy of the peerState to the "ForPeer", "GossipForPeer" and "OnIncomingRPC" functions. Then, if these functions do not return an error, we update the peerState with what these functions return.

The problem is that if "RecvdState" and "SentState" are reference types(
for eg: a pointer), the app layer can shoot itself in the foot by updating them in-place and then return an error anyways.  So either we do a deep copy of the PeerState here before passing it to the app layer and restore the state with the deep copy if there's an error or the app does it.

aarshkshah1992: 1) We can mandate that it is the app layer's responsibility to not mutate the PeerState in place if it ultimately errors out (by operating on a deep copy of the passed PeerState and then returning the modified deep copy here) but we should document it well here 

OR 

2) Add a deep copy interface that the app layer must implement for PeerState -> use that to create a deep copy here before passing the PeerState to the app callbacks and then update the mapping with the returned PeerState only if there is no error (which we already do).

aarshkshah1992: For now, I've handled this in prysm  by doing a deep copy of the PeerState before mutating it.
See https://github.com/OffchainLabs/prysm/pull/16364/changes/eda07c01989468113d919f754fcaa7d7be58060e.

sukunrt: So the latest version here makes it very clear that the app is responsible for state updates. Does that work?

 
https://github.com/libp2p/go-libp2p-pubsub/pull/674#pullrequestreview-3866365116

PR #673, opened 2026-02-13
Add threadsafe dynamic direct peer handling to GossipSub

cortze:
# Description
The current direct peer handling is a bit clunky, it only allows direct peers to be set at the start of the service. This makes the application layer unable to modify the direct peer list, or to create new services as we want to add/remove peers to that list.

The PR adds two new self-descriptive methods to the `GossipSubRouter` struct:
- `AddDirectPeer()`
- `RemoveDirectPeer()`
Both handling the underlaying `Peerstore` and `TagTracer` changes as well. Ensuring that we have all in place to ensure the connection protected with the remote peer.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -841,7 +841,7 @@ func (gs *GossipSubRouter) directPeers() []peer.ID {
 	defer gs.directRWM.RUnlock()
 	directPeers := make([]peer.ID, len(gs.direct))
 	i := 0
-	for p, _ := range gs.direct {
+	for p := range gs.direct {
 		directPeers[i] = p
 		i++
 	}
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -834,11 +834,7 @@ func (gs *GossipSubRouter) AddDirectPeer(pi peer.AddrInfo) {
 		gs.p.host.Peerstore().AddAddrs(pi.ID, pi.Addrs, peerstore.PermanentAddrTTL)
 		gs.tagTracer.addDirectPeer(pi.ID)
 	}
-	select {
-	case gs.p.eval <- addPeerFn:
-	case <-gs.p.ctx.Done():
-		return
-	}
+	gs.runSyncEvalFn(addPeerFn)
 }
 
 func (gs *GossipSubRouter) RemoveDirectPeer(p peer.ID) {
@@ -846,10 +842,23 @@ func (gs *GossipSubRouter) RemoveDirectPeer(p peer.ID) {
 		delete(gs.direct, p)
 		gs.tagTracer.removeDirectPeer(p)
 	}
+	gs.runSyncEvalFn(rmPeerFn)
+}
+
+// evalSync is generic helper to run custom functions using the pubsub eval channel
+func (gs *GossipSubRouter) runSyncEvalFn(fn func()) {
+	doneC := make(chan struct{})
+	syncFn := func() {
+		fn()
+		close(doneC)
+	}
 	select {
-	case gs.p.eval <- rmPeerFn:
+	case gs.p.eval <- syncFn:
+		select {
+		case <-doneC:
+		case <-gs.p.ctx.Done():
+		}
 	case <-gs.p.ctx.Done():
-		return
 	}
 }
 
@@ -1466,8 +1475,8 @@ func (gs *GossipSubRouter) Join(topic string) {
 				// filter our current peers, direct peers, peers we are backing off, and
 				// peers with negative scores
 				_, inMesh := gmap[p]
-				_, doBackOff := backoff[p]
 				_, direct := gs.direct[p]
+				_, doBackOff := backoff[p]
 				return !inMesh && !direct && !doBackOff && gs.score.Score(p) >= 0
 			})
 			for _, p := range more {
@@ -1481,8 +1490,8 @@ func (gs *GossipSubRouter) Join(topic string) {
 		backoff := gs.backoff[topic]
 		peers := gs.getPeers(topic, gs.params.D, func(p peer.ID) bool {
 			// filter direct peers, peers we are backing off and peers with negative score
-			_, doBackOff := backoff[p]
 			_, direct := gs.direct[p]
+			_, doBackOff := backoff[p]
 			return !direct && !doBackOff && gs.score.Score(p) >= 0
 		})
 		gmap = peerListToMap(peers)
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -10,7 +10,6 @@ import (
 	"math/rand"
 	"slices"
 	"sort"
-	"sync"
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
@@ -590,7 +589,6 @@ type GossipSubRouter struct {
 	peers      map[peer.ID]protocol.ID // peer protocols
 	extensions *extensionsState
 
-	directRWM    sync.RWMutex
 	direct       map[peer.ID]struct{}             // direct peers
 	mesh         map[string]map[peer.ID]struct{}  // topic meshes
 	fanout       map[string]map[peer.ID]struct{}  // topic fanout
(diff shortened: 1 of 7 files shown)


gossipsub.go
```diff
@@ -827,9 +830,48 @@ func (gs *GossipSubRouter) EnoughPeers(topic string, suggested int) bool {
 	return false
 }
 
+func (gs *GossipSubRouter) directPeerLen() int {
+	gs.directRWM.RLock()
+	defer gs.directRWM.RUnlock()
+	return len(gs.direct)
+}
+
+func (gs *GossipSubRouter) directPeers() []peer.ID {
```
vyzo: is this copy in any hot paths?

if so please take the read lock and iterate directly.

gossipsub_test.go
```diff
@@ -1340,6 +1340,114 @@ func TestGossipsubDirectPeers(t *testing.T) {
...
+
+	h := getDefaultHosts(t, 3)
+	psubs := []*PubSub{
+		getGossipsub(ctx, h[0], WithDirectConnectTicks(2)),
+		getGossipsub(ctx, h[1], WithDirectConnectTicks(2)),
+		getGossipsub(ctx, h[2], WithDirectConnectTicks(2)),
+	}
+
+	// test dinamic addition of direct-peers to h[1]
+	gs1, ok := psubs[1].rt.(*GossipSubRouter)
```
MarcoPolo: Applications can't access this field, how are they supposed to call these new methods?

gossipsub.go
```diff
@@ -588,6 +590,7 @@ type GossipSubRouter struct {
 	peers      map[peer.ID]protocol.ID // peer protocols
 	extensions *extensionsState
 
+	directRWM    sync.RWMutex
```
MarcoPolo: You've introduced the only mutex in this file, that's suspicious.

The existing pattern is that only the owner goroutine mutates these fields. I'd try to keep that pattern here. With that, the change becomes a bit simpler, you remove the need of the mutex and the need for the `directPeers` method.

You'll also need to add public methods to the PubSub type to expose these methods. You may be able to use the `eval chan func` as an escape hatch to avoid having to add a new channel to select on.

MarcoPolo (changes requested):

gossipsub_test.go
cortze: I thought I could create first the `GossipSubRouter`, to then call the `NewGossipSubWithRouter` option.

gossipsub.go
cortze: Thanks for the review Marco, that is a very good point let me reiterate the PR

gossipsub.go
```diff
@@ -827,6 +828,31 @@ func (gs *GossipSubRouter) EnoughPeers(topic string, suggested int) bool {
 	return false
 }
 
+func (gs *GossipSubRouter) AddDirectPeer(pi peer.AddrInfo) {
```
MarcoPolo, line 833: This should be a method on the PubSub type as that's the handle that most applications have access to. Every other method here is not expected to be called by the application, only by the pubsub goroutine.

MarcoPolo: Same comment for RemoveDirectPeer.

gossipsub.go
```diff
@@ -827,6 +828,31 @@ func (gs *GossipSubRouter) EnoughPeers(topic string, suggested int) bool {
...
+	addPeerFn := func() {
+		gs.direct[pi.ID] = struct{}{}
```
MarcoPolo: Does the caller want to wait until the function is complete? If so you should close a channel in this function and block the caller until the channel is closed.

In general, I think that's preferred.

gossipsub.go
```diff
@@ -1440,8 +1466,8 @@ func (gs *GossipSubRouter) Join(topic string) {
 				// filter our current peers, direct peers, peers we are backing off, and
 				// peers with negative scores
 				_, inMesh := gmap[p]
-				_, direct := gs.direct[p]
 				_, doBackOff := backoff[p]
+				_, direct := gs.direct[p]
```
MarcoPolo, line 1460: Why this diff?

gossipsub_test.go
MarcoPolo: Right, but every other method on GossipSubRouter only expects to be called from the pubsub goroutine. It's not intended for the user to keep a handle to the gossipsub router. It's not ideal, I know, but it's how the code is today.

gossipsub_test.go
```diff
@@ -1340,6 +1340,105 @@ func TestGossipsubDirectPeers(t *testing.T) {
 	}
 }
 
+func TestGossipsubDynamicDirectPeers(t *testing.T) {
```
MarcoPolo, line 1343: This test can be a bit simpler if you just check the direct map after adding a direct peer and after removing the direct peer.

Use the eval chan func trick to access that state without race issues.

MarcoPolo (changes requested):

gossipsub.go
cortze: I've added it as part of the PubSub Interface at [0a81e02](https://github.com/libp2p/go-libp2p-pubsub/pull/673/commits/0a81e02c4168629b47bd053ba79205c3eed7f9b0)

gossipsub.go
cortze: added it at [0a81e02](https://github.com/libp2p/go-libp2p-pubsub/pull/673/commits/0a81e02c4168629b47bd053ba79205c3eed7f9b0)

gossipsub_test.go
cortze: updated at [0a81e02](https://github.com/libp2p/go-libp2p-pubsub/pull/673/commits/0a81e02c4168629b47bd053ba79205c3eed7f9b0)

MarcoPolo:
I pushed the latest commit @cortze, please take a look

cortze:
I just fixed the test and covered a possible nil pointer given to the `tag_tracer` when the direct peer list was empty at the `GossipSubRouter` startup.

This should be ready to merge, but let me know if I should address anything else ✌🏽

tag_tracer.go
```diff
@@ -83,7 +83,9 @@ func (t *tagTracer) Start(gs *GossipSubRouter, logger *slog.Logger) {
 	t.logger = logger
 
 	t.idGen = gs.p.idGen
-	t.direct = gs.direct
+	if gs.direct != nil {
+		t.direct = gs.direct
```
MarcoPolo: Does this rely on having a stable pointer to t.direct?

MarcoPolo: is it intentional that this aliases to the gossipsub map?

tag_tracer.go
```diff
@@ -181,6 +183,20 @@ func (t *tagTracer) nearFirstPeers(msg *Message) []peer.ID {
 	return peers
 }
 
+func (t *tagTracer) addDirectPeer(p peer.ID) {
+	t.Lock()
+	defer t.Unlock()
+	t.direct[p] = struct{}{}
```
MarcoPolo: Do we need to do this if this is aliased to the gossipsub map?

tag_tracer.go
```diff
@@ -181,6 +183,20 @@ func (t *tagTracer) nearFirstPeers(msg *Message) []peer.ID {
 	return peers
 }
 
+func (t *tagTracer) addDirectPeer(p peer.ID) {
+	t.Lock()
```
MarcoPolo: I don't think we need the lock here.

tag_tracer.go
MarcoPolo: I don't like that it aliases to the gossipsub map. It means that if gs does not initialize the map this will be a nil pointer that will be dereferenced later. I've changed it to use a function instead.

cortze: I guess this made sense when the direct peer map was only initialized on GossipSub's startup. Much cleaner now with the direct check over the gs.pointer directly

MarcoPolo (approved):

MarcoPolo:
Thanks @cortze

PR #674, opened 2026-02-17
Partial Messages: Generic PeerState and Iterator style publish

MarcoPolo:
Worth reviewing commit by commit and with a focus on partialmsgs.go

### DIFF
--- a/partialmessages/partialmsgs.go
+++ b/partialmessages/partialmsgs.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"iter"
 	"log/slog"
+	"maps"
 
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -99,7 +100,7 @@ type PartialMessagesExtension[PeerState any] struct {
 	GroupTTLByHeatbeat int
 
 	// map topic -> TopicState
-	statePerTopicPerGroup map[string]*topicState[PeerState]
+	statePerTopicPerGroup map[string]map[string]*partialMessageStatePerGroupPerTopic[PeerState]
 
 	// map[topic]counter
 	peerInitiatedGroupCounter map[string]*peerInitiatedGroupCounterState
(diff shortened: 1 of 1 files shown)


pubsub.go
```diff
@@ -558,6 +553,9 @@ func NewPubSub(ctx context.Context, h host.Host, rt PubSubRouter, opts ...Option
 		seenMsgStrategy:       TimeCacheStrategy,
 		idGen:                 newMsgIdGenerator(),
 		counter:               uint64(time.Now().UnixNano()),
+		// eval is buffered to allow a single thunk to be queued within the
+		// pubsub goroutine.
+		eval: make(chan func(), 1),
```
sukunrt: Is this safe? 
Can you add more details as to why we needed this?

extensions.go
```diff
@@ -191,6 +203,45 @@ func WithPartialMessagesExtension(pm *partialmessages.PartialMessagesExtension)
 	}
 }
```
sukunrt, line 204: Should this be per topic? 
Here we require all topics with partial messages to use `PeerState` as the state type.

extensions.go
```diff
@@ -191,6 +203,45 @@ func WithPartialMessagesExtension(pm *partialmessages.PartialMessagesExtension)
...
+			resp <- errors.New("partial publishing is not enabled")
+			return
+		}
+		pme, ok := rt.extensions.partialMessagesExtension.(*partialmessages.PartialMessagesExtension[PeerState])
+		if !ok {
+			resp <- errors.New("incompatible partial messages extension type")
+			return
+		}
+
+		resp <- pme.PublishPartial(topic, message, opts)
```
sukunrt: I don't like the control flow. 

app => pubsub loop => app => sendRPC

But I see why you had to do this given the current gossipsub single goroutine system. IIUC, you can't give the app an iterator for `(peer, peerstate)` without copying **all** the state.

sukunrt (approved):

pubsub.go
MarcoPolo: backed this out, my original thinking was wrong.

extensions.go
MarcoPolo: Possibly, but I think that's worth doing separately. For now we have the same PeerState across all topics.

extensions.go
MarcoPolo: It's correct that I would have to copy the state if I wanted to give the list of peers and peerstate to the application.

But I don't think this pattern is that uncommon. Specifically the pattern of calling a function with a function parameter that you expect to be used by the function you are calling. slices.SortFunc comes to mind, but I'm sure there are many other examples.

The latest commit hopefully makes this clearer by making the PublishActions a function type that is passed into to PublishPartial.

sukunrt: I agree with the callback argument being a common pattern.

I think this will be much better to read at the call site:  
```
func (e *PartialMessagesExtension[PeerState]) PublishPartial(topic string, groupID []byte, publishActionsFn PublishActionsFn[PeerState]) error
```

partialmessages/partialmsgs.go
```diff
@@ -200,165 +169,155 @@ func (e *PartialMessagesExtension) Init(router Router) error {
...
+// The function SHOULD update the peer's peerState in the map to track the
+// parts we've sent to this peer, along with any other peer-specific application
+// data.
+//
+// The function should use the peerRequestsPartial to avoid encoding a partial
+// message if the peer did not request one.
+//
+// A peer's peerState will be the zero value if this is the first time
+// interacting with this peer. Applications can still eagerly push data in
+// this case by returning the appropriate PublishAction.
```
sukunrt: We should call out pointer types here. A sentence like. 
`For pointer types this will be nil.`

partialmessages/partialmsgs.go
```diff
@@ -200,165 +169,155 @@ func (e *PartialMessagesExtension) Init(router Router) error {
...
-			var rpc pb.PartialMessagesExtension
-			pState := s.peerState[peer]
-			pState, msg, partsMetadataToSend, err := e.GossipForPeer(topic, group, peer, pState)
-			if err != nil {
-				e.Logger.Error("Failed to encode partial message", "error", err)
-				continue
+		var newGossipPeerCount int
+		for _, p := range peers {
+			if _, ok := gState.peerState[p]; !ok {
+				newGossipPeerCount++
```
sukunrt, line 290: You can break after this statement.

sukunrt (approved):
This looks great!

partialmessages/partialmsgs.go
MarcoPolo: I'm using this to size the peerState map correctly

MarcoPolo:
merged into simplify-partial-messages branch.

PR #675, opened 2026-02-24
Use fanout for publish if mesh is empty even if we are subscribed to a topic

aarshkshah1992:
If a peer subscribes to a topic when it has not yet seen any other peers for the topic, it will set the mesh to an empty map.

Now say another peer comes in and subscribes to the topic. But, it has not been grafted yet and has yet to be added to the mesh. However, it is now added to the list of peers subscribed to a given topic.

Now, if the first peer does a Publish -> we see that the mesh exists and so don't do a fanout. However, the mesh is empty which means that the peer above never sees the message. 

Later, on a heartbeat, the second peer gets grafted but it's too late.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -1364,7 +1364,7 @@ func (gs *GossipSubRouter) rpcs(msg *Message) iter.Seq2[peer.ID, *RPC] {
 
 			// gossipsub peers
 			gmap, ok := gs.mesh[topic]
-			if !ok {
+			if !ok || len(gmap) == 0 {
 				// we are not in the mesh for topic, use fanout peers
 				gmap, ok = gs.fanout[topic]
 				if !ok || len(gmap) == 0 {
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
@MarcoPolo @sukunrt Please can I request you to take a look ?

gossipsub.go
```diff
@@ -1477,7 +1477,11 @@ func (gs *GossipSubRouter) Join(topic string) {
 			return !direct && !doBackOff && gs.score.Score(p) >= 0
 		})
 		gmap = peerListToMap(peers)
-		gs.mesh[topic] = gmap
+		// It is possible that we do not have any peers subscribed to this topic yet so continue fanout
+		// to ensure that messages we publish in the intermin are not lost
+		if len(gmap) > 0 {
+			gs.mesh[topic] = gmap
+		}
```
sukunrt, line 1484: Why do you need this? The len check when we build the list of peers to publish to is enough.

gossipsub.go
```diff
@@ -1364,7 +1364,7 @@ func (gs *GossipSubRouter) rpcs(msg *Message) iter.Seq2[peer.ID, *RPC] {
 
 			// gossipsub peers
 			gmap, ok := gs.mesh[topic]
-			if !ok {
+			if !ok || len(gmap) == 0 {
 				// we are not in the mesh for topic, use fanout peers
 				gmap, ok = gs.fanout[topic]
 				if !ok || len(gmap) == 0 {
```
sukunrt, line 1370: You can also consider adding: 
`Dlo - len(gmap)` peers from fanout.

sukunrt (approved):

PR #679, opened 2026-03-01
Elevate log level to Warn for message drops due to full buffers

### LINKED ISSUE #472: Message buffer filled log message should be an error message
I have encounter the case where I filled the buffer faster than I was reading it and messages were dropped. Took me a while to figure out because I didn't see any errors.

I noticed the `INFO` message after a while (see https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L928). It should be made more evident to people that messages are being dropped to avoid confusion by showing `ERROR` or at least a `WARN` tag in the logs.

twenty-tools:
## Summary

- Elevates log level from `Info`/`Debug` to `Warn` for all log messages related to message drops due to full buffers/queues
- Fixes #472

## Changes

When messages or RPCs are dropped because a buffer or queue is full, the log messages now use `Warn` instead of `Info` or `Debug`. This makes it immediately visible to operators when message loss occurs due to backpressure, without requiring them to scan `Info`-level logs.

Before this change, a user filling their subscription buffer faster than they were consuming it would only see an `Info`-level log line, which is easy to miss in production. This addresses the issue reported in #472.

### DIFF
--- a/floodsub.go
+++ b/floodsub.go
@@ -93,7 +93,7 @@ func (fs *FloodSubRouter) Publish(msg *Message) {
 
 		err := q.Push(out, false)
 		if err != nil {
-			fs.p.logger.Info("dropping message to peer: queue full", "peer", pid)
+			fs.p.logger.Warn("dropping message to peer: queue full", "peer", pid)
 			fs.tracer.DropRPC(out, pid)
 			// Drop it. The peer is too slow.
 			continue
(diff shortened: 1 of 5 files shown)


MarcoPolo:
Arguably this should have been done from the start. As is, I don't think we want this change as it could be unnecessarily noisy.

If you need this to be a warn or error, you should hook up a tracer and have that log appropriately.

PR #681, opened 2026-03-01
chore: bump go 1.25

laciferin2024:
Bumps the minimum Go version from 1.24 to 1.25 and applies idiomatic Go 1.22+ patterns across the codebase.

### DIFF
--- a/backoff_test.go
+++ b/backoff_test.go
@@ -34,7 +34,7 @@ func TestBackoff_Update(t *testing.T) {
 		t.Fatalf("invalid initialization: %v, \t, %s", d, err)
 	}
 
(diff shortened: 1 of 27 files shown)


MarcoPolo:
We may bump the minimum Go version in this library, but we don't need this right now.

laciferin2024:
@MarcoPolo The go-libp2p is already at 1.25 , won't it be best to bump it now. I have well tested this pr, used the `go fix` tool and even updated the code to use the idiomatic Go1.22+ patterns.

MarcoPolo:
This change should be done by core maintainers. We are not going to accept version updates from contributors. If there's something that would unblock or enable you by making this change, please open an issue first.

PR #685, opened 2026-03-05
partialmsgs: PeerRequestsPartial means PeerRequestsPartial

MarcoPolo:
Do not also check if we support partial messages on the topic because we may not even be subscribed to that topic. The application will simply not use this if they don't support partial messages on this topic.

### DIFF
--- a/extensions.go
+++ b/extensions.go
@@ -246,9 +246,13 @@ type partialMessageRouter struct {
 	gs *GossipSubRouter
 }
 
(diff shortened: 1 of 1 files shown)


sukunrt (approved):
>Do not also check if we support partial messages on the topic because we may not even be subscribed to that topic. The application will simply not use this if they don't support partial messages on this topic.

Can you add this to the method documentation?

PR #688, opened 2026-03-31
go.mod: upgrade to go-libp2p v0.48.0

dindinw:
Pion DTLS's usage of random nonce generation with AES GCM ciphers risks leaking the authentication key. upgrade to go-libp2p v0.48.0 to fix the issue. see CVE-2026-26014 && https://github.com/libp2p/go-libp2p/pull/3469


MarcoPolo:
You don't need this to update go-libp2p for your project. Adding this would force users to update go-libp2p even if they don't use WebRTC.

It's better to prefer the minimum version here so that users have options on what version they want to use.

PR #689, opened 2026-04-02
Don't process stale RPCs

MarcoPolo:
There's a subtle race condition where we end up handling an RPC from a disconnected peer if they disconnect after queuing an RPC to the `p.incoming` channel and we process the disconnect event before the queued RPC.

One example of this showing up was incorrectly flagging duplicate extension messages, which was observed in production. Consider the following sequence:
1. Peer sends a message, queued in `p.incoming`
2. Peer disconnects.
3. pubsub handles the disconnect.
4. clears `es.peerExtensions`
5. pubsub handles queued `p.incoming`
6. sets `es.peerExtensions` in response to receiving a "first" rpc.
7. peer reconnects and sends an extensions control message, pubsub penalizes the peer for sending duplicate extensions control.

Since this is so timing dependent, it's quite hard to have a test for this. In production it mostly shows up when the server is overloaded with other work (at 100% cpu).

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -899,8 +899,13 @@ func (gs *GossipSubRouter) Preprocess(from peer.ID, msgs []*Message) {
 
 func (gs *GossipSubRouter) HandleRPC(rpc *RPC) {
 	err := gs.extensions.HandleRPC(rpc)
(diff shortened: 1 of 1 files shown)


MarcoPolo:
replaced by #691

PR #690, opened 2026-04-07
partialmessages: init fanout if empty

sukunrt:


extensions.go
```diff
@@ -266,10 +267,20 @@ func (r partialMessageRouter) MeshPeers(topic string) iter.Seq[peer.ID] {
...
+			if len(peerSet) == 0 {
+				peers := r.gs.getPeers(topic, r.gs.params.D, func(p peer.ID) bool {
+					_, direct := r.gs.direct[p]
+					return !direct && r.gs.score.Score(p) >= r.gs.publishThreshold
+				})
+
+				if len(peers) > 0 {
+					peerSet = peerListToMap(peers)
+					r.gs.fanout[topic] = peerSet
+				}
```
MarcoPolo: Is this the same logic that inits the fanout state in a normal publish?

MarcoPolo: If so, should we dedupe that logic

MarcoPolo (approved):

PR #691, opened 2026-04-10
Fix leak around different lifecycles of inbound vs outbound streams.

MarcoPolo:
In certain cases peer state may be leaked if the state is created on incoming messages and only cleared when our outbound stream is closed.

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -88,7 +88,7 @@ type PubSub struct {
 	peerOutboundQueueSize int
 
 	// incoming messages from other peers
-	incoming chan *RPC
+	incoming chan incomingUnion
 
 	// addSub is a control channel for us to add and remove subscriptions
 	addSub chan *addSubReq
@@ -168,7 +168,7 @@ type PubSub struct {
 	peers map[peer.ID]*rpcQueue
 
 	inboundStreamsMx sync.Mutex
-	inboundStreams   map[peer.ID]network.Stream
+	inboundStreams   map[peer.ID]inboundHandler
 
 	seenMessages    timecache.TimeCache
 	seenMsgTTL      time.Duration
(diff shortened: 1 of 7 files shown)


pubsub.go
```diff
@@ -944,12 +979,7 @@ func (p *PubSub) processLoop(ctx context.Context) {
 			if ok {
 				q.Close()
 				delete(p.peers, pid)
-				for t, tmap := range p.topics {
-					if _, ok := tmap[pid]; ok {
-						delete(tmap, pid)
-						p.notifyLeave(t, pid)
```
sukunrt, line 950: maybe these four statements should be a method. 
```
q.Close()
delete(p.peers, pid)
p.clearPeerFromTopicsState(pid)
p.rt.RemovePeer(pid)
```

pubsub.go
```diff
@@ -918,9 +945,17 @@ func (p *PubSub) processLoop(ctx context.Context) {
...
+		case in := <-p.incoming:
+			switch in.kind {
+			case incomingKindRPC:
+				p.handleIncomingRPC(in.rpc)
+			case incomingKindNewStream:
+				p.rt.OnNewIncomingStream(
+					in.s.Conn().RemotePeer(), in.s.Protocol())
+			case incomingKindClosedStream:
+				p.onClosedIncomingStream(
+					in.s.Conn().RemotePeer(), in.s.Protocol())
```
sukunrt, line 957: Do we need the protocol in onClosedIncomingStream? `in.s.Protocol()`?

sukunrt (approved):

pubsub.go
MarcoPolo: It's nice symmetry, and lets you cleanup state associated with a specific gossipsub version. For example, if extensions is supported you need to clean up extensions state.

Of course you could just always clean up all state, but this seems a little clearer.

PR #699, opened 2026-05-08
gossipsub: clean peer gater stats on inbound close

MarcoPolo:


peer_gater.go
```diff
@@ -380,14 +380,29 @@ func (pg *peerGater) OnNewOutboundStream(p peer.ID, proto protocol.ID) {
 }
 
 func (pg *peerGater) OnClosedOutboundStream(p peer.ID) {
+	pg.removePeerStats(p, true)
+}
+
+func (pg *peerGater) OnClosedIncomingStream(p peer.ID, _ protocol.ID) {
+	pg.removePeerStats(p, false)
```
sukunrt, line 387: This feels like a NoOp to me as the underlying logic only decrements connected if outbound is true. 

Is this intentional?

sukunrt (approved):

peer_gater.go
MarcoPolo: The important part is this:

```
if st.connected == 0 {
delete(pg.peerStats, p)
}
```

Which will clear any state associate with the peerStats when the incoming stream is closed. This cleans up the case where we added some state due to a message on the incoming stream, but never had an outbound stream.

PR #701, opened 2026-05-16
fix: recover panics from topic validators

Sahil-4555:
## Issue

This PR fixes a crash path in the pubsub validation pipeline when an application-provided topic validator panics.

Incoming messages are passed through registered validators before they are delivered to local subscribers or forwarded through the network. Some applications use these validators to parse message payloads, check application-level rules, or reject malformed data. If that validator code hits an unexpected edge case and panics, the panic could previously escape the validation path.

This is especially risky in the async validation flow. Async validators run inside goroutines, and in Go an unrecovered panic in any goroutine terminates the whole process. That means a remote peer could send a message that triggers a panic bug in application validation code and cause the node process to exit instead of just dropping that message. The old code also released validation throttle tokens only after the validator returned normally, so panic paths could skip cleanup.

## Fix

This change catches panics at the validator boundary in `validateMsg`.

If a validator panics, pubsub now logs the panic with peer and topic context, then treats the result as `ValidationReject`. The message is dropped through the normal validation failure path, and the process continues running. This keeps validator bugs isolated to the message being validated instead of letting them crash the node.

The validation throttle cleanup is also moved into deferred cleanup blocks. This makes sure throttle tokens are released even when validation exits through a panic. The recovery is intentionally scoped to the application validator call, so internal pubsub programming errors are not broadly hidden.

## Tests

This PR adds regression tests for both async validation paths.

`TestValidatePanicRecovery` covers the normal async topic validator path. It sends a message that causes the validator to panic, verifies that the message is not delivered, then sends a valid message and checks that validation still works afterward.

`TestValidatePanicRecoveryMulti` covers the multi-validator path where validators run concurrently. It verifies the same behavior there: a panicking validator rejects the bad message, does not crash the process, and does not prevent a later valid message from being delivered.

### DIFF
--- a/validation.go
+++ b/validation.go
@@ -366,8 +366,10 @@ loop:
 		select {
 		case v.validateThrottle <- struct{}{}:
 			go func() {
+				defer func() {
+					<-v.validateThrottle
+				}()
 				v.doValidateTopic(async, src, msg, result, onValid)
-				<-v.validateThrottle
 			}()
 		default:
 			v.p.logger.Debug("message validation throttled; dropping message from peer", "peer", src)
@@ -440,8 +442,10 @@ func (v *validation) validateTopic(vals []*validatorImpl, src peer.ID, msg *Mess
 		select {
 		case val.validateThrottle <- struct{}{}:
 			go func(val *validatorImpl) {
+				defer func() {
+					<-val.validateThrottle
+				}()
 				rch <- val.validateMsg(ctx, src, msg)
-				<-val.validateThrottle
 			}(val)
 
 		default:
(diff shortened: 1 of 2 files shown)


marten-seemann (changes requested):
This does not belong in a library. You can implement this at the application layer.

Sahil-4555:
That makes sense. My main concern was about where this callback is running. The validator is called from the pubsub validation pipeline, and in the async path it runs inside a goroutine managed by pubsub. So if it panics on remote message data, the impact is not limited to that message; it can bring down the full process.

But I also agree that the actual panic is coming from application-provided validator code. Recovering it inside the library would also mean the library is deciding that every validator panic should become `ValidationReject`, and that may not be the right behavior for all users. Some applications may prefer to crash loudly during development, and others can wrap their validators if they want recovery behavior.

Please let me know if you have any other feedback on this?

marten-seemann (changes requested):
It is in general unsafe to recover from panics. This is why recover statements don't belong in libraries. They belong very close to the code that is actually being recovered from.

This PR should be rejected.

PR #703, opened 2026-05-17
fix: cap peer score delivery records

Sahil-4555:
## Issue

This PR adds a count limit to the message delivery records kept by peer scoring.

Peer scoring keeps a short-lived delivery record for each message ID it is tracking. These records are used while a message moves through validation, and they help account for duplicate deliveries, near-first deliveries, valid messages, invalid messages, ignored messages, and throttled validation. Today those records expire after `SeenMsgTTL`, but there is no count-based limit on how many can exist during that TTL window.

That means memory use can grow with the rate of unique messages that reach validation or delivery tracing. This is not a permanent leak, because the records are eventually removed by delivery-record GC. The issue is the active window: on a busy topic, or under adversarial traffic, peers can keep sending unique messages and cause the delivery-record map to grow until those records age out.

For production nodes with peer scoring enabled, this creates a resource-scaling problem. Memory use is bounded by time, but not by volume. A high enough message rate during `SeenMsgTTL` can still create a large number of retained records.

## Fix

This change adds a configurable cap for peer score delivery records.

`PeerScoreParams` now has a `DeliveryRecordCap` field. If it is left as zero, peer scoring uses `DefaultDeliveryRecordCap`. When the delivery-record table reaches the cap, `getRecord` first runs GC to clear any expired records. If the table is still full after that, it does not allocate a new tracking record for the new message ID.

Existing tracked records continue to work as before. The fix also keeps direct scoring behavior where possible: first-message delivery scoring still runs before delivery-record tracking, and validation-failed messages still penalize the sender even when the record table is full. What gets skipped at capacity is the extra per-message tracking used for duplicate and near-first delivery accounting.

This keeps memory bounded without adding new goroutines, changing message delivery behavior, or blocking the validation path. Operators that expect very high message throughput can raise the cap through score params.

## Tests

This PR adds coverage at both the score layer and the GossipSub layer.

`TestScoreDeliveryRecordCap` checks the delivery-record table directly. It verifies that the table does not grow beyond the configured cap, that invalid-message scoring still applies when the cap is full, and that new records can be created again after old records expire and GC runs.

`TestGossipsubScoreDeliveryRecordCap` drives the same behavior through real GossipSub publish and subscribe flow with peer scoring enabled. It publishes more unique messages than the configured cap, verifies those messages are still delivered, and checks that the peer score delivery-record table remains capped.

The peer score parameter validation tests were also updated so negative caps are rejected and valid configured caps are accepted.

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -2395,6 +2395,57 @@ func TestGossipsubPeerFeedback(t *testing.T) {
 	})
 }
 
+func TestGossipsubScoreDeliveryRecordCap(t *testing.T) {
+	synctestTest(t, func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		const topic = "delivery-record-cap"
+		const recordCap = 3
+
+		hosts := getDefaultHosts(t, 2)
+		sender := getGossipsub(ctx, hosts[0])
+		receiver := getGossipsub(ctx, hosts[1],
+			WithPeerScore(
+				&PeerScoreParams{
+					AppSpecificScore:  func(peer.ID) float64 { return 0 },
+					DecayInterval:     time.Second,
+					DecayToZero:       0.01,
+					DeliveryRecordCap: recordCap,
+				},
+				&PeerScoreThresholds{
+					GossipThreshold:   -1,
+					PublishThreshold:  -10,
+					GraylistThreshold: -100,
+				}))
+
+		connect(t, hosts[0], hosts[1])
+
+		sub, err := receiver.Subscribe(topic)
+		if err != nil {
+			t.Fatal(err)
+		}
+		time.Sleep(50 * time.Millisecond)
+
+		for i := range 10 {
+			msg := []byte(fmt.Sprintf("message %d", i))
+			if err := sender.Publish(topic, msg); err != nil {
+				t.Fatal(err)
+			}
+			assertReceive(t, sub, msg)
+		}
+
+		score := receiver.rt.(*GossipSubRouter).score
+		score.Lock()
+		recordCount := len(score.deliveries.records)
+		score.Unlock()
+
+		if recordCount != recordCap {
+			t.Fatalf("expected %d delivery records, got %d", recordCap, recordCount)
+		}
+	})
+}
+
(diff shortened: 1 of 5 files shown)


MarcoPolo:
I don't think this is the right solution.

1. We don't want to allow one peer to consume the whole allotment.
2. This is the wrong place to set limits on the number of unique messages received. Applications are responsible for validating messages, and protocol designers are responsible for ensuring there aren't infinite valid messages in a small time period (otherwise you end up with all sorts of resource exhaustion/amplification attacks).

This change should be first discussed as an issue.

Sahil-4555:
thanks, for going through this and sharing the feedback

**On one peer consuming the whole allotment** -  yeah that's a fair point actually. So the thing is, `deliveries.records` is a global `map[string]*deliveryRecord` keyed by message ID, not per peer. Inside each record there's a `peers map[peer.ID]struct{}` that tracks which all peers sent that particular message. So the growth here is really driven by unique messages coming in, not by how many peers are sending them. A per-peer cap would need quite a different data structure altogether - can't really just bolt a counter on top of what's there today. But if you feel per-peer is the right way to go, I'm happy to explore that direction.

**On wrong place / apps should handle this** - I understand the philosophy here, and generally I agree that applications should be handling validation. But the specific problem is - the delivery record gets created at `validation.go:332` through `tracer.ValidateMessage(msg)` -> `peerScore.ValidateMessage` -> `deliveries.getRecord()`. This runs before any application validator even gets to see the message (those kick in at `validation.go:335`). So even if the app rejects every single message, the record is already sitting in the map by that point. There's no way for application-level validation to prevent this allocation - it's the library's own internal bookkeeping for scoring.

And the library is already doing this kind of bounding in the same pipeline - `validateQ` is capped at 32, `validateThrottle` at 8192, `MaxIHaveLength` at 5000. `deliveries.records` is basically the only scoring structure that has no count limit at all - just a time-based GC running every 60s clearing things older than `SeenMsgTTL` (120s default). So for that whole 120s window, it just keeps growing with the message rate, no ceiling.

**On opening an issue first** - If you think this is something worth looking into, I can open an issue with the details and we can take the discussion there. Or we can just continue here itself, whichever you prefer. Let me know.

PR #704, opened 2026-05-25
pubsub: coalesce announce retries per peer and add backoff

Sahil-4555:
### Issue

When we subscribe to or unsubscribe from a topic, `announce` pushes a subscription RPC to every connected peer. If a peer's outbound queue is full the push fails with `ErrQueueFull` and we spawn a goroutine to retry that one `(peer, topic)` announcement: sleep 1-1000ms, hop onto `processLoop` through `p.eval`, try one push, and spawn another goroutine if it fails again. There's no backoff. The chain ends if the peer goes away, if pubsub shuts down, or if our local subscription state for that topic flips (the retry re-checks `mySubs`/`myRelays` and stops rescheduling when it no longer matches), but otherwise it keeps going for as long as the queue stays full.

A peer's queue can stay full for a long time without the peer being unhealthy. The 30s write deadline in `writeRpc` (`comm.go:232`) applies to each individual write, not to the congested period, so a peer that reads steadily but slower than we produce keeps every write well inside the deadline while its queue stays saturated. The retry chain then runs for as long as the imbalance lasts, at roughly two `p.eval` closures per second.

Because a chain is created per failed announce, the live chains work out to roughly congested peers times subscribed topics, and all of them serialise through `processLoop`, which also handles message delivery, validation results and peer add/remove. On a node with many peers under backpressure that's a lot of pointless traffic through the one place you least want it.

This isn't a goroutine leak, and it doesn't grow without bound over time: each link exits after scheduling the next, so one goroutine is alive per chain and nothing accumulates, and the staleness check above retires chains as subscriptions change. The problem is the steady-state rate.

### Fix

Announcements that fail to push go into a per-peer pending set, `pendingAnnounce map[peer.ID]map[string]bool`, owned by `processLoop`, with one retry loop per peer draining it.

- The set is keyed by topic, so a later announcement replaces an earlier one for the same topic. A subscribe followed by an unsubscribe during congestion delivers only the unsubscribe.
- One loop per peer flushes all of that peer's pending topics as a single RPC via `rpcWithSubs(subopts...)`, the same shape as the hello packet. Live retry work becomes one goroutine and one pending set per peer, independent of how many topics failed.
- The delay starts random over `(0, 1s]` as it does today, then doubles with up to 10% jitter, capped at 30s. A permanently congested peer costs one `processLoop` closure every 30s instead of two per second, and a briefly congested one still converges in about a second.

Retries aren't abandoned after a fixed number of attempts, because a lost announcement is a silent problem. Subscriptions travel only as deltas, and the full set goes out just once per new outbound stream in `getHelloPacket` (`comm.go:21`), so a peer that never receives an announcement keeps a stale `p.topics[topic]` for the life of that stream. That map is what `getPeers` draws mesh and fanout GRAFT candidates from (`gossipsub.go:2251`), what `emitGossip` picks IHAVE targets from (`gossipsub.go:2046`), and what decides who a published message is forwarded to in both routers (`gossipsub.go:1362`, `floodsub.go:87`). So a lost subscribe means the peer never grafts us, gossips to us or forwards to us for that topic, and a lost unsubscribe means it keeps sending us messages we discard. The loop exits instead when the announcements are delivered, when they go stale, when the peer leaves `p.peers`, or when the pubsub context is cancelled.

This also folds the duplicated `pb.RPC_SubOpts` construction in `announce` and the retry path into a `subOpts` helper, and adds `rpcQueue.Len` so tests don't reach past its mutex.

Behaviour is otherwise unchanged. Retries still run on `processLoop`, still re-check local topic state and skip announcements that no longer match it, still stop on context cancellation, still ignore departed peers. The outbound queue stays non-blocking and tracing still records dropped and sent RPCs. Mesh behaviour, scoring, validation, message propagation and shutdown are untouched.

### Tests

`announce_retry_test.go` adds six tests on a synctest harness that stands in for `processLoop` and counts the closures the retry path sends through it:

- five minutes of congestion, then space frees up, and the announcement still arrives
- five failed announces to one peer produce one pending set, one loop and one RPC carrying all five subopts (the queue has room for all five as separate RPCs, so uncoalesced retries would show up as five)
- subscribe then unsubscribe while congested delivers only the unsubscribe
- an announcement that no longer matches local state is discarded instead of sent
- a disconnected peer has its pending state and goroutine reclaimed, checked by watching for zero further `p.eval` traffic
- over a ten-minute window the flush count stays near window/maxBackoff and nowhere near the no-backoff rate

Each test runs in a synctest bubble, and since `synctest.Test` waits for every goroutine in the bubble to exit, they also cover the loop shutting down on context cancellation.

I checked the tests actually bite by breaking each guarantee on purpose: removing the backoff, dropping pending state when the push fails, sending one subopt instead of the coalesced set, not superseding, removing the staleness check, and not reclaiming a departed peer. Each of those fails at least one test.

`TestAnnounceRetry` is updated for the new entry point and still checks the announcement reaches the wire over a real stream.

Full package suite passes, including under `-race`.

One caveat worth stating: I don't have a production profile showing `processLoop` starving because of this. The case is about rate and scaling, argued from the code and from the flush counts in the backoff test, so if you'd rather see `p.eval` latency measured under synthetic congestion first, happy to go and get it.


Sahil-4555:
heya, @marten-seemann can you please review this PR when get chance?

PR #707, opened 2026-05-27
Limit control message size of RPCs

MarcoPolo:
This adds a new option to limit to total size of control fields (including subscriptions) when receiving a pb rpc. This fixes an issue where allowing a large message payload should not also allow a large amount of control messages at once. 

This depends on #706

### DIFF
--- a/pb/validator.go
+++ b/pb/validator.go
@@ -29,15 +29,53 @@ func ValidateRawRPCControlMessageSize(msg []byte, maxControlMessageSize int) err
 		}
 
 		switch field {
-		case rpcPublishFieldNumber, rpcPartialFieldNumber:
+		case rpcPartialFieldNumber:
+			// There can only be a single partial message per rpc. So the
+			// overhead size is small per RPC.
+		case rpcPublishFieldNumber:
+			payloadLen, err := dataFieldLength(fieldValue)
+			if err != nil {
+				return err
+			}
+			// Add the message overhead as part of the control size
+			controlSize += tagLen + fieldLen - payloadLen
+
 		default:
 			controlSize += tagLen + fieldLen
-			if controlSize > maxControlMessageSize {
-				return fmt.Errorf("rpc control size exceeds max control message size: %d > %d", controlSize, maxControlMessageSize)
-			}
 		}
-
 		msg = fieldValue[fieldLen:]
+
+		if controlSize > maxControlMessageSize {
+			return fmt.Errorf("rpc control size exceeds max control message size: %d > %d", controlSize, maxControlMessageSize)
+		}
 	}
 	return nil
+}
+
+func dataFieldLength(msg []byte) (int, error) {
+	const dataFieldNumber protowire.Number = 2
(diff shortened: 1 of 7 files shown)


pb/validator.go
```diff
@@ -0,0 +1,81 @@
...
+			payloadLen, err := dataFieldLength(fieldValue)
+			if err != nil {
+				return err
+			}
+			// Add the message overhead as part of the control size
+			controlSize += tagLen + fieldLen - payloadLen
+
+		default:
+			controlSize += tagLen + fieldLen
+		}
```
sukunrt, line 45: should this be more permissive? it restricts future field additions by counting everything non publish and partial as control. Maybe we should let unknown fields be handled as publish field

pubsub.go
```diff
@@ -427,8 +436,21 @@ func (rpc *RPC) split(limit int) iter.Seq[*RPC] {
...
+
+					nextRPC = &RPC{RPC: pb.RPC{Control: &pb.ControlMessage{
+						Extensions: extensions,
+					}}, from: rpc.from}
+				}
+			}
+
 			for _, graft := range ctl.GetGraft() {
-				if nextRPC.Control.Graft = append(nextRPC.Control.Graft, graft); proto.Size(&nextRPC.RPC) > limit {
+				if nextRPC.Control.Graft = append(nextRPC.Control.Graft, graft); nextRPC.exceedsSizeLimits(limit, controlLimit) {
```
sukunrt, line 453: These lines will do O(n) allocations now, I think. 
Maybe we can improve controlRPCLength to check subscriptions and control separately without introducing a new pb or binary search here to find the optimal message length?

pubsub.go
```diff
@@ -516,6 +564,20 @@ func (rpc *RPC) split(limit int) iter.Seq[*RPC] {
...
+
+func controlRPCSize(rpc *RPC) int {
+	if rpc == nil {
+		return 0
+	}
+	return proto.Size(&pb.RPC{
+		Subscriptions: rpc.Subscriptions,
+		Control:       rpc.Control,
+	})
+}
```
sukunrt, line 589: maybe set publish and partial to nil here to check the length. This will help avoid allocation.

sukunrt (approved):

pubsub.go
sukunrt: it'll also make it similar to the validation on the inbound side.

pb/validator.go
MarcoPolo: It's more permissive but may open a vulnerability that goes unnoticed. On the other hand, as is, it may incorrectly drop valid application messages. I'd hope that early application testing would catch that more readily than the more permissive option.

We can add documentation to the protobuf file and a README to have our future selves remember to update this method?

pubsub.go
MarcoPolo: fixed

pubsub.go
MarcoPolo: changed so this func doesn't allocate

Issue #705, opened 2026-05-25
Memory leak in clearPeerFromTopicsState retains empty topic maps indefinitely


johannamoran:
### Reported By
Reported by @tonghuaroot  and @lidel 

Verified by @seetadev 

### Summary
clearPeerFromTopicsState removes peer entries from topic maps but never deletes the now-empty topic map itself from p.topics. As a result, an attacker can cause unbounded growth of p.topics by repeatedly introducing unique topic IDs, even when peers disconnect cleanly.

This affects the current master and appears to have existed for the lifetime of the implementation.

The issue is independently exploitable even when downstream consumers apply per-RPC subscription caps (WithSubscriptionFilter), because a patient attacker can still accumulate topic keys over time using many small RPCs.

### Root cause
In clearPeerFromTopicsState:
func (p *PubSub) clearPeerFromTopicsState(pid peer.ID) {
    for _, tmap := range p.topics {
        delete(tmap, pid)
    }
}

the peer entry is deleted from each tmap, but empty topic maps are retained forever inside p.topics.

Similarly, in the unsubscribe path inside handleIncomingRPC, topic entries are removed from the per-topic map but the outer p.topics entry is never deleted when the map becomes empty.

There also appears to be no cleanup of p.topics during Close().

### Impact
An untrusted peer can:

1. Connect over /meshsub/1.x.0
2. Send subscriptions for many unique topic IDs
3. Disconnect

The peer entries are removed, but the topic-string keys remain retained for the process lifetime.

Repeated cycles permanently grow memory usage until restart.

Downstream consumers without WithSubscriptionFilter are more exposed because a single RPC can contain arbitrarily many subscriptions, but the underlying leak exists regardless of filtering.

### Related context
This is closely related to the recently disclosed advisory against js-libp2p:

- GHSA-4f8r-922h-2vgv

The same general pattern (state maps retaining attacker-controlled keys after disconnect/unsubscribe) appears here in the Go implementation.

### Suggested fix
Delete empty topic maps after peer removal.

Example outline:

func (p *PubSub) clearPeerFromTopicsState(pid peer.ID) {
    for topic, tmap := range p.topics {
        delete(tmap, pid)
        if len(tmap) == 0 {
            delete(p.topics, topic)
        }
    }
}

The unsubscribe branch in handleIncomingRPC should similarly remove empty topic maps.
It may also be worth clearing p.topics during Close().

### Notes on mitigations
[...]

PR #710, opened 2026-05-29
Fix memory leak due to retention of empty topic maps

bismuth01:
This PR aims to fix the issue #705 which was discovered and reported by @tonghuaroot and @lidel with verification by @seetadev

`clearPeerFromTopicsState` and `handleIncomingRPC` functions have been updated to check if the map of peers for a topic is empty after removal of a peer and then delete that entry.

In addition to the fix, testcase for the same has been added as `TestTopicDeleteEmptyTopicMap` to ensure future changes adhere to this requirment.
I have also created a [simple pubsub application](https://github.com/bismuth01/pubsub-mem-leak) with mDNS discovery to connect nodes a common topic to optionally application test if the changes have affected anything else.


MarcoPolo:
prefer #708

PR #711, opened 2026-05-30
PeerFeedback is now synchronous

MarcoPolo:
and returns an error after shutdown

closes #702

### DIFF
--- a/gossipsub_test.go
+++ b/gossipsub_test.go
@@ -22,6 +22,7 @@ import (
 	"sync/atomic"
 	"testing"
 	"testing/quick"
+	"testing/synctest"
 	"time"
 
 	"github.com/libp2p/go-libp2p-pubsub/internal/gologshim"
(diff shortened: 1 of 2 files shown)


Sahil-4555:
hey, i noticed that the mine issue was about a shutdown race where PeerFeedback may be called while PubSub is in the process of stopping. the test only verifies the behavior after shutdown has completed, but it doesn't seem to exercise the concurrent cancellation scenario that can cause the send on p.eval to block.

MarcoPolo:
I don't understand what you mean.

Sahil-4555:
what i mean was about the test - the name says `AfterShutdown` and the flow is `cancel` -> `synctest.Wait` -> call `PeerFeedback`. By the time `synctest.Wait()` returns, the process loop has already exited through `case <-ctx.Done(): return`, so there's no goroutine reading from `p.eval` anymore. At that point `syncEval`'s select will always pick `p.ctx.Done()` - it's basically just checking if the context is already dead.

But the actual bug that causes the hang in production is a bit different - it happens when `PeerFeedback` is called while the process loop is still alive but busy doing something else and can't read from `p.eval`. That's the during-shutdown race, not the after-shutdown case.

I think a test that keeps the process loop occupied (like sending a blocking thunk on `p.eval`) and then cancels the context while calling `PeerFeedback` would cover the real failure mode better. The current test does verify the post-shutdown path which is good, but the concurrent scenario is what actually hangs things in practice.

MarcoPolo:
If I understand you're point correctly, it's about getting coverage of ctx.Done in both select cases, once when inserting the eval thunk, and once when waiting for the result. That's reasonable. It should be a test of syncEval itself though. I'll make that change.

MarcoPolo:
Tests merged in #713

sukunrt (approved):

PR #712, opened 2026-05-30
feat: allow pubsub over limited (relayed) connections

mudler:
Hello from the LocalAI team :wave: 

I've been trying to trace an issue while running edgevpn (https://github.com/mudler/edgevpn, which is used by https://github.com/mudler/LocalAI ) between two VMs in qemu network configured user-mode (which uses slirp, and makes the networks completely isolated between each VM) where they can connect but quickly via DHT resolution and nat/holepunch, but loose connection right after. This scenario was working  in a very old version of libp2p, but it stopped to work at a certain version (which I can't pinpoint to), but it was really hard to track down so I eventually gave up. 

Lately I've been put Claude into bruteforcing this just for the sake of it, as it was quite of a timesink otherwise, and it managed to break down the issue (after a long session of driving it in the correct place), now, what I don't know if it is actually OK with the overall design, or if it should be tackled from another angle, so I'd be happy to work this out.

This seems part of the solution - I'm still struggling on other parts, but at least make nodes visible to each other.

Here is the result from Claude:

Three places hard-filtered peers whose only connection to the local host was Limited (network.Limited, i.e. relay/circuit-v2 path):

  * pubsub.go:1008 (handlePendingPeers):  main connection-event loop discarded any peer whose Connectedness wasn't strictly Connected.
  * peer_notify.go:28 (watchForNewPeers): initial peer enumeration discarded Limited peers.
  * comm.go:166 (handleNewPeer): opened the gossipsub stream without network.WithAllowLimitedConn, so even if a Limited peer made it past the prior filters the swarm layer would reject the stream with network.ErrLimitedConn.

These checks date back to when relay/circuit was considered too slow or unreliable for pubsub. With AutoRelay+DCUtR now a regular operating mode for NAT'd hosts, NAT-traversed peers routinely live on Limited connections, and they silently lose every pubsub message. The fix is to accept network.Limited everywhere alongside network.Connected and to pass network.WithAllowLimitedConn at the single stream-open call.

Verified with a standalone reproducer that puts two hosts behind a relay with a ConnectionGater that blocks all direct A↔B addresses, and confirms the A↔B connection is Limited before publishing. Prior to this change the message never arrives (60s timeout); with the change applied the message arrives in well under one heartbeat.

### DIFF
--- a/comm.go
+++ b/comm.go
@@ -163,7 +163,7 @@ func (p *PubSub) notifyPeerDead(pid peer.ID) {
 }
 
 func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing *rpcQueue) {
-	s, err := p.host.NewStream(ctx, pid, p.rt.Protocols()...)
+	s, err := p.host.NewStream(network.WithAllowLimitedConn(ctx, "pubsub"), pid, p.rt.Protocols()...)
 	if err != nil {
 		p.logger.Debug("error opening new stream to peer", "err", err, "peer", pid)
 
--- a/peer_notify.go
+++ b/peer_notify.go
@@ -25,7 +25,7 @@ func (ps *PubSub) watchForNewPeers(ctx context.Context) {
 	ps.newPeersPrioLk.RLock()
 	ps.newPeersMx.Lock()
 	for _, pid := range ps.host.Network().Peers() {
-		if ps.host.Network().Connectedness(pid) != network.Connected {
+		if c := ps.host.Network().Connectedness(pid); c != network.Connected && c != network.Limited {
 			continue
 		}
 		ps.newPeersPend[pid] = struct{}{}
--- a/pubsub.go
+++ b/pubsub.go
@@ -1005,7 +1005,7 @@ func (p *PubSub) handlePendingPeers() {
 	for pid := range newPeers {
 		// Make sure we have a non-limited connection. We do this late because we may have
 		// disconnected in the meantime.
-		if p.host.Network().Connectedness(pid) != network.Connected {
+		if c := p.host.Network().Connectedness(pid); c != network.Connected && c != network.Limited {
 			continue
 		}
 


comm.go
```diff
@@ -163,7 +163,7 @@ func (p *PubSub) notifyPeerDead(pid peer.ID) {
 }
 
 func (p *PubSub) handleNewPeer(ctx context.Context, pid peer.ID, outgoing *rpcQueue) {
-	s, err := p.host.NewStream(ctx, pid, p.rt.Protocols()...)
+	s, err := p.host.NewStream(network.WithAllowLimitedConn(ctx, "pubsub"), pid, p.rt.Protocols()...)
```
mudler, line 166: I'm guessing there should be some constants that could be used here instead, the literal feels wrong

mudler:
cc @MarcoPolo :pray:

mudler:
confirmed to be working - ready for review

cortze:
I'm not @MarcoPolo but it looks like this solution is not directly solving the issue in the first place (correct me if I'm wrong).

`circuit-relay-v2` offers restricted connections between two nodes through a relay. They are limited to ensure that the relay is not overwhelmed and it is generally used to coordinate the hole-punch over the DCUtR protocol, which will allow later the direct connection between both nodes after the hole-punch succeeds. The fact that the connection is established in the first place means that the relay connection worked, but it might get shut down as long as the limit in bandwidth is reached.

That said, allowing pubsub to use limited connections seems to be just a trick that won't fix the issue as long as the connection isn't upgraded successfully. It will only survive until the bandwidth limit is reached, and then you will miss messages anyways. You might want to consider going back to `circuit-relay-v1` if that is the behavior that your expect.

Can you check if there DCUtR was enabled? And if it was, what was the status of the connection upgrade attempt between the nodes?
- linking here the reference to the libp2p.Host [option](https://github.com/libp2p/go-libp2p/blob/6a6b1ffd8a9ae70dd927862f7d3d558d24be8070/options.go#L550)

CC: @dennis-tra

mudler:
Thanks for digging into this @cortze your read is fair and honestly the part that bugged me about this patch from the start is exactly what you're pointing at: it doesn't solve anything if the connection never upgrades. Let me share where I've ended up after living with this for a while, and I'd love your push-back on it.

DCUtR is on (edgevpn turns on EnableHolePunching by default), and in the setup that originally surfaced this I'm pretty sure it's trying but not winning, both peers are sitting behind QEMU user-mode slirp, which from the peer's side looks roughly symmetric-NAT and is just, very hostile to holepunch in general. 

https://github.com/mudler/edgevpn/blob/e378f0baace95ee1e41619d209c95c7c10f93f1b/pkg/config/config.go#L536

I want to stop arguing about this on vibes though, so I'm spinning up the repro again right now to pull the actual dcutr / holepunch events out of the two daemons and I'll drop them in a follow-up comment here. Better to talk about it with real logs. 
  
Even with that, I take your point about relay-v2 being a "coordinate the upgrade, then get out of the way" path rather than a substrate. Where I keep landing is that for our use case, and I think for anyone using pubsub more as a small-record gossip ledger than as a fat data plane, gossipsub's control traffic is in the low-kilobyte-per-minute range. 
IHAVE/GRAFT/PRUNE/IWANT, plus a handful of tiny app-level RPCs. That fits inside a default circuit-v2 reservation pretty comfortably between refreshes. The actual failure mode I kept hitting wasn't "we ran out of relay budget", it was "the peer never even joins the mesh because Connectedness != Connected rejects it on sight," at handlePendingPeers, watchForNewPeers, and handleNewPeer. 

From an operator standpoint the patched behaviour gives me "mesh forms;  DCUtR keeps trying in the background; messages flow either way" instead of "no mesh until DCUtR wins, and on a hostile NAT it just doesn't." That's a strict improvement for me, but I get that "strict improvement for me" is not the same as "good default for everyone."

So, if you'd rather not flip this on unconditionally, would you be open to it as a PubSubOption (something like WithAllowLimitedConnections())? Applications that know they're running tiny ledgers over pubsub, or that just want the mesh to form while DCUtR keeps trying, can opt in; everyone else gets today's semantics. Happy to rework the PR that way if you think it's a more acceptable shape.
  
I'll be back with the DCUtR logs from the repro.

mudler:
OK, ran the repro again with `libp2p` set to `debug`. Putting the data here so you've got the same view I do:

The cluster peer is reachable to us only via `/p2p-circuit`. Pulled from the DHT-provider records on one node, here is the other cluster node's set of advertised addresses:

```
12D3KooWEnEsEWqC8psws1FhAvMhbQbGbNgvwPqfdxoZqYKxfc5x:
  /ip4/23.94.186.31/tcp/4001/p2p/12D3KooWFvy2xvb…/p2p-circuit
  /ip4/51.81.209.112/tcp/4001/p2p/12D3KooWCvcygNQ…/p2p-circuit
  /ip4/51.81.209.112/udp/4001/quic-v1/p2p/12D3KooWCvcygNQ…/p2p-circuit
  /ip4/10.0.2.15/…           ← peer's own slirp NAT addr (same prefix as our own → unroutable)
  /ip4/10.42.0.1/…           ← k3s pod CIDR (overlay, only routable inside the cluster)
  /ip6/fec0::5054:…/…        ← site-local IPv6 (same prefix as our own → unroutable)
  /ip4/127.0.0.1/…  /ip6/::1/…    ← loopback
```

So all routable paths to that peer are `/p2p-circuit` over two different public relay operators. There is no direct path. Despite that, the mesh is fully functional: `/api/nodes` of edgevpn returns both peers `Online` from each side, k3s forms across the link, and role coordination completes. (I'm using this to form a kubernetes cluster, so edgevpn sits as network layer).

DCUtR is enabled but is silent. `EnableHolePunching` is in edgevpn's option list, but at `libp2p-log-level=debug` the only namespaces emitting are `dht` and `dht/RtRefreshManager`, nothing from `p2p-holepunch`, `dcutr`, `autonat`, or `autorelay`. So no upgrade attempt is being kicked off. Not sure why.

On `circuit-relay-v1`: I went to try the v1 path you suggested as the alternative, but v1 isn't in go-libp2p anymore. As of `v0.48.0` (the latest, which edgevpn pins), only `circuit-v2` remains; the v1 transport package was removed. So unless we're prepared to fork `go-libp2p` back to a pre-removal tag, v1 isn't an option for us.

cc @MarcoPolo @dennis-tra @cortze

sukunrt:
>would you be open to it as a PubSubOption (something like WithAllowLimitedConnections())

Given that you know that your network load is small enough that you can get away with limited relays, I think it's okay to add this option. Also maybe log on startup with INFO that this option is on and may lead to spurious disconnections to peer.

sukunrt:
On ^ :
the relays will only transmit a fixed amount of data and not a fixed limited bandwidth. They'll also only relay for 2 minutes. Does that work for you?

mudler:
> On ^ : the relays will only transmit a fixed amount of data and not a fixed limited bandwidth. They'll also only relay for 2 minutes. Does that work for you?

that would work, but I'm curious about the 2m relay. is that hardcoded/done on purpose? can I control this if e.g. operating my own relays?

sukunrt:
>can I control this if e.g. operating my own relays?

If you're operating your own relays you can configure this, and you can even make them unlimited in which case this problem goes away. 

The 2m is just go-libp2p default and it is what I think the IPFS(kubo) nodes run with.

mudler:
> > can I control this if e.g. operating my own relays?
> 
> If you're operating your own relays you can configure this, and you can even make them unlimited in which case this problem goes away.

that'd be perfect then and would work well for us :+1:

PR #715, opened 2026-06-15
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replace `interface{}` with `any` across 5 Go files.

- gossipsub.go
- peer_notify.go
- pubsub.go
- score.go
- validation.go

This is a mechanical refactor using Go 1.18+ `any` alias.


gammazero:
Handled as part of #716

PR #719, opened 2026-07-02
rerun validation if node publishes a previously seen message

sukunrt:
This allows the node to publish messages that were previously ignored/rejected for whatever reason. Forwarding only happens once. 

This is a problem in ethereum when a node sees a block very early and attests immediately. If the node pushed the attestation as fanout, it'd mean that the fanout peers if they've not seen the block with park the message for later processing(ValidationIgnore). But now when they later do see the block and process the attestation, they won't forward the attestation to the mesh again. See test `TestGossipsubForwardAfterIgnore`

This will always rerun validation on locally published messages which don't pass validation. I think this is okay as it's strictly for locally published messages and not network received messages. 

@Wondertan does this change break anything for you?

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -922,16 +922,16 @@ func (gs *GossipSubRouter) HandleRPC(rpc *RPC) {
 	}
 
 	iwant := gs.handleIHave(rpc.from, ctl)
-	ihave := gs.handleIWant(rpc.from, ctl)
+	msgs := gs.handleIWant(rpc.from, ctl)
 	prune := gs.handleGraft(rpc.from, ctl)
 	gs.handlePrune(rpc.from, ctl)
 	gs.handleIDontWant(rpc.from, ctl)
 
-	if len(iwant) == 0 && len(ihave) == 0 && len(prune) == 0 {
+	if len(iwant) == 0 && len(msgs) == 0 && len(prune) == 0 {
 		return
 	}
 
-	out := rpcWithControl(ihave, nil, iwant, nil, prune, nil)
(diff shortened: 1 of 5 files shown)


Wondertan:
Presumably it shouldn't break Celestia's DA network, but still better be tested. However, I am not part of the team anymore, so cc @vgonkivs

vgonkivs:
Thanks @Wondertan. Tbh I don't see how this PR breaks our nodes.

pubsub.go
```diff
@@ -1587,20 +1602,26 @@ func (p *PubSub) checkSigningPolicy(msg *Message) error {
...
+		return
+	}
 	p.notifySubs(msg)
 	if !msg.Local {
 		p.rt.Publish(msg)
 	}
 }
 
 func (p *PubSub) publishMessageBatch(batchAndOpts messageBatchAndPublishOptions) {
+	msgs := make([]*Message, 0, len(batchAndOpts.messages))
```
MarcoPolo: Why do you copy the list?

gossipsub.go
```diff
@@ -922,16 +922,16 @@ func (gs *GossipSubRouter) HandleRPC(rpc *RPC) {
 	}
 
 	iwant := gs.handleIHave(rpc.from, ctl)
-	ihave := gs.handleIWant(rpc.from, ctl)
+	msgs := gs.handleIWant(rpc.from, ctl)
```
MarcoPolo: as long as we're changing this, `msgs` is very generic. Maybe `iWantResponses`

MarcoPolo (approved):

PR #730, opened 2026-08-26
Track outstanding IWANTs correctly

cskiraly:
This PR adds IWANT tracking on top of the previous change, improving
IWANT efficiency and reducing duplicates.
It builds on a rebased version of #625, including the work of @MarcoPolo 

## Details
Replace the per-message-ID request counter with a ledger of which peers
have an outstanding IWANT for each message and when it was sent. The
limit now caps how many distinct peers we ask for a given message at
once, and a peer with a still-live request is not asked again, so one
peer can no longer soak up the whole budget. An ask older than the
follow-up time expires and frees its slot.

Requests are recorded only after the per-peer truncation that decides
what we actually send, so an ID dropped by truncation no longer leaves a
phantom ask that would block the message from being requested elsewhere.

Clear the ledger when a message is received and validated rather than in
the pre-validation Preprocess step, so an unvalidated or rejected message
can no longer cancel a pending request.

### DIFF
--- a/gossipsub.go
+++ b/gossipsub.go
@@ -82,7 +82,7 @@ var (
 	GossipSubIDontWantMessageThreshold        = 1024 // 1KB
 	GossipSubIDontWantMessageTTL              = 3    // 3 heartbeats
 
-	GossipSubMaxIWantsPerMessageIDPerHeartbeat = 10
+	GossipSubMaxIWantsPerMessageID = 10
 )
 
 type checksum struct {
@@ -249,10 +249,13 @@ type GossipSubParams struct {
 	// IDONTWANT is cleared when it's older than the TTL.
 	IDontWantMessageTTL int
 
-	// MaxIWantsPerMessageIDPerHeartbeat is the maximum number of pending IWANT
-	// requests allowed per message ID per heartbeat. This helps limit the
-	// number of duplicates we'll receive from peers.
-	MaxIWantsPerMessageIDPerHeartbeat int
+	// MaxIWantsPerMessageID is the maximum number of IWANT requests allowed
+	// per message ID while requests are outstanding. This helps limit the
+	// number of duplicates we'll receive from peers. The budget for a message
+	// ID refills IWantFollowupTime after its last request, so a request that
+	// went unanswered can be retried without waiting for a heartbeat boundary
+	// and a request in flight is not re-issued at one.
+	MaxIWantsPerMessageID int
 }
 
 func (params *GossipSubParams) validate() error {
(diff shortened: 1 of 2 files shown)


gossipsub.go
```diff
@@ -1913,6 +1958,62 @@ func (gs *GossipSubRouter) heartbeat() {
...
+	if ledger == nil {
+		ledger = make(iwantLedger)
+		gs.outstandingIWants[mid] = ledger
+	}
+	ledger[p] = now
+}
+
+// fulfillIWant clears any outstanding IWANTs for a message once it has been
+// received and validated.
+func (gs *GossipSubRouter) fulfillIWant(msg *Message) {
```
sukunrt, line 2003:
```suggestion
func (gs *GossipSubRouter) clearIWants(mid string) {
```

gossipsub.go
```diff
@@ -1913,6 +1958,62 @@ func (gs *GossipSubRouter) heartbeat() {
 	gs.extensions.Heartbeat()
 }
 
+// iwantLedger records, per peer, when we sent that peer an IWANT for a message
+// ID. An entry older than IWantFollowupTime has expired: it no longer counts
+// against MaxIWantsPerMessageID and the peer may be asked again.
+type iwantLedger map[peer.ID]time.Time
```
sukunrt, line 1964: NIT: I think the raw type map[peer.ID]time.Time reads better. We also have `mesh` as the same type in GossipSubRouter.

gossipsub.go
sukunrt, line 242: We should update the documentation for this to suggest that this is the wait time for followup iwants

gossipsub.go
```diff
@@ -1004,6 +1038,12 @@ func (gs *GossipSubRouter) handleIHave(p peer.ID, ctl *pb.ControlMessage) []*pb.
 	iwantlst = iwantlst[:iask]
 	gs.iasked[p] += iask
 
+	// record the asks only for the messages we are actually sending, so a
+	// request dropped by truncation leaves no phantom entry in the ledger
+	for _, mid := range iwantlst {
+		gs.recordIWant(mid, p, now)
+	}
+
 	gs.gossipTracer.AddPromise(p, iwantlst)
```
sukunrt, line 1047: This struct `gossipTracer` does nearly the same logic as what you've implemented. I think we should merge the two. It's a sizeable refactor so we should do this in a different PR. This one is fine as it is for now.

gossipsub.go
```diff
@@ -1344,6 +1385,7 @@ func (gs *GossipSubRouter) PublishBatch(messages []*Message, opts *BatchPublishO
 }
 
 func (gs *GossipSubRouter) Publish(msg *Message) {
+	gs.fulfillIWant(msg)
```
sukunrt, line 1388: we can expire these in `PreProcess`, no?

sukunrt (approved):

gossipsub.go
sukunrt: >Clear the ledger when a message is received and validated rather than in
the pre-validation Preprocess step, so an unvalidated or rejected message
can no longer cancel a pending request.

This seems fine. I was thinking of hash based messageIDs.

sukunrt: But doing it in Preprocess is better in case of `ValidationIgnore` verdicts from the application. This happens in case we get an attestation but don't have a block. So that particular attestation will be parked as `ValidationIgnore` and we'll keep sending IWants to get that. Arguably, this situation will resolve before we send a new iwant.

sukunrt:
I think we need to store IHave's as well now so that we can send a new IWant when one expires. Right now we have to wait for a new IHave signal from someone before fetching. We should do this in a different PR.

